PR #4, opened 2022-07-20
Add minimal Android example and an attempt to build an example with Compose

jmartinesp:
Created a minimal editor in `example-android` which renders the output from the `ComposerModel` and can format selections to bold.

The Compose example is not working due to the bug in the `CoreTextField` implementation mentioned [here](https://issuetracker.google.com/issues/172239032). I'd still want to keep it in case it gets solved some time soon, as migrating from legacy Android views could be worth it in the long run.


bindings/wysiwyg-ffi/Cargo.toml
```diff
@@ -6,7 +6,7 @@ repository = "https://gitlab.com/andybalaam/wysiwyg-rust"
 description = "Swift and Kotlin bindings for wysiwyg-rust"
 keywords = ["matrix", "chat", "messaging", "composer", "wysiwyg"]
 license = "Apache-2.0"
-name = "wysiwyg-ffi"
+name = "uniffi-wysiwyg-composer"
```
jmartinesp, line 9: I also had to change the name of the lib to not to have to edit the generated bindings for Android every single time new bindings were generated. This might affect iOS @aringenbach.

PR #6, opened 2022-07-21
Add replace and delete in range functions, dump_state function

jmartinesp:


crates/wysiwyg/src/composer_model.rs
```diff
@@ -88,6 +103,17 @@ where
         self.replace_text(&[])
     }
 
+    /**
+     * Deletes text in an arbitrary start..end range.
+     */
+    pub fn delete_in(&mut self, start: usize, end: usize) -> ComposerUpdate<C> {
```
andybalaam, line 109: Can we have a test for this one?

examples/example-android/app/src/main/java/uniffi/wysiwyg_composer/wysiwyg_composer.kt
```diff
@@ -42,15 +42,15 @@ open class RustBuffer : Structure() {
 
     companion object {
         internal fun alloc(size: Int = 0) = rustCall() { status ->
-            _UniFFILib.INSTANCE.ffi_wysiwyg_composer_e48b_rustbuffer_alloc(size, status).also {
```
andybalaam, line 45: Is this generated code? Maybe it should not be checked in to git?

bindings/wysiwyg-ffi/src/wysiwyg_composer.udl
```diff
@@ -32,6 +35,12 @@ interface ComposerUpdate {
     sequence<ComposerAction> actions();
 };
 
+dictionary ComposerState {
+    sequence<u16> html;
+    u64 start;
```
andybalaam, line 40: Why u64? In other places I was using u32 - should I have used u64?

andybalaam (approved):
A few comments, but feel free to merge when ready.

examples/example-android/app/src/main/java/uniffi/wysiwyg_composer/wysiwyg_composer.kt
jmartinesp: I was adding it as part of the project because I had to change the name of the lib inside this file to make it work before. We can move it to the build folder as we used to do and generate it on each build.

bindings/wysiwyg-ffi/src/wysiwyg_composer.udl
jmartinesp: `u32` might be enough, I'm used to using Long (64bit signed int) for indexes or sizes, but since this is `u64` it will be twice the size. I used `u64` because the bindings cannot directly translate `usize` as far as I could see and casting `usize` to `u32` could hypothetically truncate the number. But since we'll be using this for text messages I doubt anyone will write a message that long 😅 .

andybalaam: Yeah, I think `u64` is probably a better choice everywhere I used `u32` since it has a lossless conversion from `usize`, although it does mean we have a lossy conversion on the way in, so maybe it's no better? (And as you say, in practice we are not expecting numbers to get anywhere near the limits.)

PR #13, opened 2022-08-04
Model ownership and Combine

aringenbach:
This change moves the view model ownership to either the parent view in SwiftUI implementation or in the hosting view in UIKit implementation.
This is required in order to allow model altering interactions from the outside (such as clearing all text).

It also replaces the SwiftUI PreferenceKey usage by a Combine implementation, directly set up on the view model.


jmartinesp (approved):
LGTM!

PR #21, opened 2022-08-16
Split dom in different modules.

jmartinesp:
Please let me know if you would have split them differently. Also, I'm not 100% sure how or if I should split the tests too.


andybalaam (approved):
Looks good - I think we should split the tests, but we can do it later. Good to merge this quickly to avoid conflicts, I reckon.

PR #26, opened 2022-08-17
Add support for links

jmartinesp:
* Also added minimal support on Android.

Limitations:

* Links can only be added, not removed or replaced.
* They can only be added inside a single text node.
* There is not url escape at the moment since we have no way to do it. It would probably have to be done at the client side.


crates/wysiwyg/src/dom/range.rs
```diff
@@ -38,3 +39,19 @@ pub struct SameNodeRange {
...
+                    "SameNode: {} {}",
+                    range.start_offset, range.end_offset
+                )
+                .as_str(),
+            ),
+            Range::TooDifficultForMe => f.write_str("TooDifficultForMe"),
+            Range::NoNode => f.write_str("NoNode"),
+        }
+    }
+}
```
andybalaam, line 57: Since Range already implements Debug, you can just use `{:?}` in the formatting code inside `replace_text_in`, and then you won't need to impl `Display` here.

andybalaam (approved):
Looks good, 1 comment.

PR #27, opened 2022-08-18
Merge Formatting and Hyperlink nodes into ContainerNode.

jmartinesp:
Creates a `ContainerNodeKind` enum with `Generic`, `Formatting` and `Link` cases. As Formatting and Link need to clone the `Vec<C>` values passed I also had to add `where C: Clone` constraints to several types and impls.

Formatting and Link nodes can now be created using `ContainerNode::new_formatting` and `ContainerNode::new_link`. There are also helpers in `DomNode`.


andybalaam (approved):
Nice, simplifies things quite a bit. I think a next step to consider would be to remove the `Element` trait and provide its functionality directly in `ContainerNode`, if that works?

PR #30, opened 2022-08-18
Basic HTML parsing into a Dom

andybalaam:


crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -0,0 +1,200 @@
...
+
+    /// Copy all panode's information into node (now we know it's a container).
+    fn convert_container(
+        padom: &PaDom,
+        child: &PaNodeContainer,
+        node: &mut ContainerNode<u16>,
+    ) {
+        let tag = child.name.local.as_ref();
+        match tag {
+            "b" | "i" | "strong" | "em" | "del" | "u" => {
```
jmartinesp, line 78: We should probably add "code" to the list of formatting tags.

jmartinesp (changes requested):

crates/wysiwyg/src/dom/parser/parse.rs
andybalaam: Good catch, fixed.

jmartinesp (approved):
LGTM!

PR #32, opened 2022-08-19
Move cm and tx functions into a separate testutils module

andybalaam:


jmartinesp (approved):
LGTM! I was actually wondering how we should approach something like this.

PR #36, opened 2022-08-19
Move Dom struct into separate file from dom module

andybalaam:


jmartinesp (approved):
LGTM. I'm not too sure about the `_struct` suffix but I guess it's difficult to find a better name.

PR #38, opened 2022-08-19
Replacing text in multi-node selections

andybalaam:


crates/wysiwyg/src/composer_model.rs
```diff
@@ -298,6 +310,95 @@ where
...
+                        }
+                    }
+                }
+            }
+        }
+
+        // Delete in reverse order to avoid invalidating handles
+        to_delete.reverse();
+
+        while !to_delete.is_empty() {
```
jmartinesp, line 375: The logic of `replace_multiple_nodes` and especially this loop is quite clever, but it took me a while to understand how this worked. Maybe we could add some comments explaining how the function works?

crates/wysiwyg/src/composer_model.rs
```diff
@@ -101,29 +104,38 @@ where
...
+            self.state.dom.find_range(start, end),
+        )
+    }
+
+    fn do_replace_text_in_range(
+        &mut self,
+        new_text: &[C],
+        mut start: usize,
+        range: Range,
+    ) -> ComposerUpdate<C> {
```
jmartinesp, line 119: It's not really an issue, but couldn't this still be used as a single function passing start, end and calculating the range inside?

andybalaam: Thank you - that was left around from me using fake ranges. Fixed in d1faf9f

crates/wysiwyg/src/composer_model.rs
andybalaam: Yes, definitely. I've just refactored it heavily in my next PR, so if I promise to it do after that, can we merge this?

jmartinesp: Sure, that's fine 👍

jmartinesp (approved):
LGTM!

PR #40, opened 2022-08-22
Select multiples nodes

jmartinesp:


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -125,6 +125,14 @@ where
         // anyone pushing onto this, because the handles will be unset
         &mut self.children
     }
+
+    pub fn len(&self) -> usize {
```
andybalaam, line 129: What do you think of the name `len`? - my first impression would be that it tells you how many children there are. I might go for `num_codeunits` or `len_codeunits` or even `codeunits` - what do you think?

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -125,6 +125,14 @@ where
...
+        let mut result: usize = 0;
+        for child in self.children() {
```
andybalaam, line 131: This is possibly better as `self.children().iter().map(|child| child.len()).sum()`

crates/wysiwyg/src/lib.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+extern crate core;
```
andybalaam, line 15: Shouldn't need this in modern Rust, unless I am confused.

andybalaam (commented):
Looks good so far. Haven't delved deep since it's still draft.

crates/wysiwyg/src/dom/nodes/container_node.rs
jmartinesp: I actually changed it to `text_len`. Maybe `len_codeunits` is a better name 🤔 .

crates/wysiwyg/src/lib.rs
jmartinesp: I think the IDE automatically added that on some buggy autocompletion 😅 .

crates/wysiwyg/src/dom/find_result.rs
```diff
@@ -12,15 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::dom::dom_handle::DomHandle;
+use crate::dom::range::DomLocation;
 
 #[derive(Clone, Debug, PartialEq)]
 pub enum FindResult {
```
jmartinesp, line 18: `FindResult` might not even be needed anymore, we could just return the list of locations. If it's empty, it's the same as `NotFound`. What do you think?

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -108,166 +80,265 @@ enum RangeLocation {
...
+        let container_node_len = node.text_len();
+        if !node.handle().is_root()
+            && start <= cur_offset
+            && cur_offset + container_node_len <= end
+        {
+            results.push(DomLocation {
+                node_handle: node.handle(),
+                start_offset: 0,
+                end_offset: container_node_len,
+                location_type: RangeLocationType::Middle,
```
jmartinesp, line 157: I'm not too sure about how to handle `Start`, `End`. Maybe manually mutate the first and last returned locations in `find_range`?

crates/wysiwyg/src/dom/nodes/container_node.rs
andybalaam: Happy with either

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -342,4 +418,62 @@ mod test {
...
         }
     }
+
+    #[test]
+    fn finding_a_range_across_several_nodes_works() {
+        let d = cm("test<b>ing a| </b>new feature").state.dom;
+        let range = d.find_range(2, 12);
+        if let Range::MultipleNodes(r) = range {
+            // 3 text nodes + bold node
+            assert_eq!(4, r.nodes.len());
```
andybalaam, line 428: We could `.to_html()` the nodes here to check we found what we think we found?

crates/wysiwyg/src/tests/testutils_composer_model.rs
```diff
@@ -208,13 +205,36 @@ pub fn tx(model: &ComposerModel<u16>) -> String {
             }
         }
         Range::NoNode => panic!("No node!"),
-        Range::TooDifficultForMe => {
-            dbg!((model.state.start, model.state.end));
-            todo!("Range too difficult!")
+        Range::MultipleNodes(locations) => {
```
andybalaam, line 208: This is awesome! Can we have some tests for it in `testutils_composer_model.rs` please?

andybalaam (approved):
Minor comments, but looks great. I suggest we merge Arnaud's one first though.

crates/wysiwyg/src/dom/find_range.rs
jmartinesp: Added `to_html()` check for each node + a helper function, is that what you meant?

crates/wysiwyg/src/tests/testutils_composer_model.rs
jmartinesp: Added some new test cases to `cm_and_tx_roundtrip` with multiple nodes, fixed `tx` since it didn't apply the selection properly.

PR #41, opened 2022-08-22
Handle ordered/unordered lists

aringenbach:


crates/wysiwyg/src/composer_model.rs
```diff
@@ -238,6 +242,47 @@ where
...
+                    let text = t.data();
+                    let list_node = DomNode::new_list(
+                        list_tag.to_html(),
+                        vec![DomNode::Container(ContainerNode::new_list_item(
+                            "li".to_html(),
+                            vec![DomNode::Text(TextNode::from(text.to_vec()))],
+                        ))],
+                    );
+                    self.state.dom.replace(range.node_handle, vec![list_node]);
+                }
```
andybalaam, line 265: We should panic in an else clause here - if it's not a text node, something is wrong.

crates/wysiwyg/src/composer_model.rs
```diff
@@ -354,6 +436,61 @@ where
...
     }
+
+    fn add_list_item(&mut self, handle: DomHandle, location: usize) {
+        let list_node = self.state.dom.lookup_node_mut(handle);
+        if let DomNode::Container(list) = list_node {
+            list.append(DomNode::new_list_item(
+                "li".to_html(),
+                vec![DomNode::Text(TextNode::from("\u{200B}".to_html()))],
+            ));
+            self.state.start = Location::from(location + 1);
```
andybalaam, line 447: I think it would be good to select the hidden space, so should `self.state.start` be set to `location` instead of `location + 1`?

crates/wysiwyg/src/composer_model.rs
```diff
@@ -354,6 +436,61 @@ where
...
+                    parent.remove(handle.index_in_parent());
+                    if parent.children().len() == 0 {
+                        parent.append(DomNode::Text(TextNode::from(
+                            "".to_html(),
+                        )));
+                    }
+                    self.state.start = Location::from(location);
+                    self.state.end = Location::from(location);
+                } else {
+                    panic!("List has no parent container")
```
andybalaam, line 475: Longer term, we should handle `li` not inside `ol` or `ul` - this could happen in HTML, even though we won't generate it. Probably fine for it to be a TODO for now, but we should think about it.

crates/wysiwyg/src/composer_model.rs
```diff
@@ -354,6 +436,61 @@ where
...
+                }
+            } else {
+                list.remove(list_item_index);
+                let parent_handle = handle.parent_handle();
+                let parent_node = self.state.dom.lookup_node_mut(parent_handle);
+                if let DomNode::Container(parent) = parent_node {
+                    // TODO: should probably append a paragraph instead
```
andybalaam, line 482: Why a paragraph instead of a zero width space?

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -98,6 +98,35 @@ where
...
 
+    pub fn find_parent_list_item(
+        &mut self,
+        child_handle: DomHandle,
+    ) -> Option<DomHandle> {
+        fn find_list_item<C: Clone>(
+            dom: &Dom<C>,
+            node: &DomNode<C>,
+        ) -> Option<DomHandle> {
+            if node.handle().has_parent() == false {
```
andybalaam, line 109: `            if !node.handle().has_parent() {`

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -149,6 +178,22 @@ impl ContainerNode<u16> {
             handle: DomHandle::new_unset(),
         }
     }
+
+    pub fn is_empty_list_item(&self) -> bool {
+        match self.kind {
+            ContainerNodeKind::ListItem() => {
+                let html = self.to_html();
```
andybalaam, line 185: I'm a bit uncomfortable with converting to html here - I think `to_html` might have side effects in the future.

But this can be a TODO to fix later if you like.

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -149,6 +178,22 @@ impl ContainerNode<u16> {
...
+                if html == "<li></li>".to_html()
+                    || html == "<li>\u{200b}</li>".to_html()
```
andybalaam, line 187: This `if` could be `return html == "<li></li>".to_html() || html == "<li>\u{200b}</li>".to_html()`

crates/wysiwyg/src/tests/test_lists.rs
```diff
@@ -0,0 +1,58 @@
...
+
+#[test]
+fn test_ordered_list() {
+    let mut model = cm("|");
+    model.create_ordered_list();
+    assert_eq!(tx(&model), "<ol><li>|</li></ol>");
+    replace_text(&mut model, "abcd");
+    assert_eq!(tx(&model), "<ol><li>abcd|</li></ol>");
+    model.enter();
+    assert_eq!(tx(&model), "<ol><li>abcd</li><li>\u{200b}|</li></ol>");
```
andybalaam, line 29: If we selected the space here...

crates/wysiwyg/src/tests/test_lists.rs
```diff
@@ -0,0 +1,58 @@
...
+    replace_text(&mut model, "efgh");
+    assert_eq!(tx(&model), "<ol><li>abcd</li><li>\u{200b}efgh|</li></ol>");
```
andybalaam, line 31: ...then it would be gone here, which would be nice

crates/wysiwyg/src/tests/test_lists.rs
```diff
@@ -0,0 +1,58 @@
...
+}
+
+#[test]
+fn test_unordered_list() {
```
andybalaam, line 35: I suspect we need tests for where some other text already exists.

I think the cursor will be in the wrong place in that case.

I am happy for this code to be merged, and then we fix that case. There are going to be LOADS of corner cases!

andybalaam (approved):
Happy to merge now, but we will need tons more test cases for this, and (obviously) we are going to need to handle cross-node selections as soon as that work is merged.

crates/wysiwyg/src/composer_model.rs
aringenbach: I got a slightly weird selection issue with this on iOS, while this could probably be fixed with a platform-specific workaround, there is also the issue where we move the cursor somewhere else then move back to this TextNode...
Rather than reselecting the hidden space, it feels like it could be better to have some mechanic to automatically remove the hidden space when inserting near it ? (or even keeping it should be fine)

crates/wysiwyg/src/composer_model.rs
aringenbach: I mean a paragraph with the zero width space inside it, perhaps it's not relevant.

crates/wysiwyg/src/dom/nodes/container_node.rs
aringenbach: Yeah actually what we need here is a helper that actually just writes out the plain text from all children, this shouldn't be too hard to achieve I suppose. 
Because if I have something like "<li><strong>\u{200b}</strong></li>" I would expect to clear the `<li>` as well.

crates/wysiwyg/src/composer_model.rs
andybalaam: Yeah. I was also wondering about just deleting all zero-width spaces whenever we convert to HTML.

Fine to leave it for now. When we add testcases all this is going to work itself out.

crates/wysiwyg/src/composer_model.rs
andybalaam: I'm not sure whether we want that or not ... fine to leave a TODO to think about it.

crates/wysiwyg/src/dom/nodes/container_node.rs
andybalaam: Good point. That is another test case we can create as soon as we have multi-node ranges.

andybalaam (approved):
Great!

PR #44, opened 2022-08-25
[iOS] Expand tests for attributed/HTML ranges handling

aringenbach:


platforms/ios/example/Wysiwyg/ContentView.swift
```diff
@@ -37,11 +38,25 @@ struct ContentView: View {
...
+                    Text(sentMessage.html)
+                        .accessibilityIdentifier(.htmlContentText)
+                }
+            }
         }
         Spacer()
     }
 }
+
+private extension View {
```
andybalaam, line 60: Is this supposed to be here even though it's empty? (Forgive my ignorance)

andybalaam (approved):
Looks good, with my very limited Swift :-)

platforms/ios/example/Wysiwyg/ContentView.swift
aringenbach: Nope, I missed it 🙈

PR #45, opened 2022-08-25
Add trait for raw text representation

aringenbach:
Add a new trait for raw text representation and implement it in node/document. 
It also updates the helper testing for empty list items.

### DIFF
--- a/crates/wysiwyg/src/tests/test_formatting.rs
+++ b/crates/wysiwyg/src/tests/test_formatting.rs
@@ -16,7 +16,7 @@
 
 use crate::tests::testutils_composer_model::{cm, tx};
 
-use crate::{InlineFormatType, Location, ToRawText};
+use crate::{InlineFormatType, Location};
 
 #[test]
 fn selecting_and_bolding_multiple_times() {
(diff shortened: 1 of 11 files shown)


crates/wysiwyg/src/tests/test_formatting.rs
```diff
@@ -29,6 +29,7 @@ fn selecting_and_bolding_multiple_times() {
         &model.state.dom.to_string(),
         "<strong>aa</strong>bb<strong>cc</strong>"
     );
+    assert_eq!(&model.state.dom.to_raw_text(), "aabbcc");
```
andybalaam, line 32: This would be better as a separate test. Would also be worth checking e.g. an empty Dom, some nested tags, some empty tags etc.

crates/wysiwyg/src/tests/test_lists.rs
```diff
@@ -32,6 +32,7 @@ fn test_ordered_list() {
     assert_eq!(tx(&model), "<ol><li>abcd</li><li>\u{200b}|</li></ol>");
     replace_text(&mut model, "efgh");
     assert_eq!(tx(&model), "<ol><li>abcd</li><li>\u{200b}efgh|</li></ol>");
+    assert_eq!(&model.state.dom.to_raw_text(), "abcd\u{200b}efgh");
```
andybalaam, line 35: Again, separate tests would be better, maybe even in a different test file.

andybalaam (changes requested):
Looks great - just requesting some more tests.

andybalaam (approved):
Nice thanks!

PR #46, opened 2022-08-25
Join nodes when the selection spans across them

andybalaam:


crates/wysiwyg/src/node_joiner.rs
```diff
@@ -0,0 +1,207 @@
...
+            let current_container = current_text.parent_handle();
+
+            // Find the parent that contains both the <b> tags that we are
+            // going to join.
+            let parent = current_container.parent_handle();
+
+            // Look for our sibling container - e.g. the other <b>
+            // TODO: check we have a sibling instead of crashing here
+            let sibling =
+                parent.child_handle(current_container.index_in_parent() + 1);
```
jmartinesp, line 97: While writing the code for formatting several nodes, I added `fn next_sibling() -> DomHandle` and `prev_sibling` for `DomHandle`. Do you think it makes sense to do it there?

crates/wysiwyg/src/node_joiner.rs
```diff
@@ -0,0 +1,207 @@
...
+///
+/// For example, if your selection starts in a bold tag, leaves that tag,
+/// and then ends up in another bold tag, the final result should be a
+/// single bold tag containing all your text.
+pub(crate) struct NodeJoiner<S>
+where
+    S: UnicodeString,
+{
+    first_type: Option<String>,
+    last_type: Option<String>,
```
jmartinesp, line 31: For this I would just use `ContainerNodeKind` so we can merge `<b>` and `<strong>`, but it depends on whether we want to do it or not. This approach would also need extra checks for `Formatting` kind, i.e.

jmartinesp (approved):
Just two comments about possible improvements, but the code LGTM as it is.

crates/wysiwyg/src/node_joiner.rs
andybalaam: Yes, I'll do this next week

crates/wysiwyg/src/node_joiner.rs
andybalaam: Yeah, that sounds good - I'll look next week

andybalaam:
Thanks, I'll merge now to avoid conflicts, and do those improvements on Tuesday.

PR #49, opened 2022-08-26
Allow formatting of several nodes

jmartinesp:
What this PR does:

* Separates formatting nodes into a separate file.
* Adds `ComposerModel.check_format_selection_type` fn that returns `Extend` to create new formatting nodes and `Remove` to remove them.
* `find_range` returns all intersections with other nodes, including parents of the text nodes. There is a workaround for SameRangeNode where we check that there's only 1 'leaf node' (that is, one text node right now).
* Added several extra properties to `DomLocation` such as _position, length, is_leaf_. Also added some extra functions to check if it's the start or the end of the selection and if the selection completely covers it, as well as next and prev sibling generator.
* Adding formatting nodes when `check_format_selection_type` returns `Extend` is supported, removing them when it returns `Remove` is left as a to-do.

There might also be some conflicting parts: I created a `merge_children` function on `ContainerNode` to merge sibling children of the same type, then created a helper `ComposerModel.merge_formatting_nodes` function that uses it. Recently we added a `NodeJoiner`, I need to check if these 2 functions do the same and if they do, which one we should keep (or both).

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -18,6 +18,7 @@ use crate::dom::nodes::dom_node::DomNode;
 use crate::dom::to_html::ToHtml;
 use crate::dom::to_raw_text::ToRawText;
 use crate::dom::{HtmlChar, UnicodeString};
+use crate::InlineFormatType;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct ContainerNode<S>
@@ -37,7 +38,7 @@ where
     S: UnicodeString,
 {
     Generic, // E.g. the root node (the containing div)
-    Formatting,
+    Formatting(InlineFormatType),
     Link(S),
     List,
     ListItem,
@@ -66,10 +67,28 @@ where
         }
     }
 
-    pub fn new_formatting(format: S, children: Vec<DomNode<S>>) -> Self {
+    pub fn new_formatting_from_tag(
+        format: S,
+        children: Vec<DomNode<S>>,
+    ) -> Self {
         Self {
             name: format.clone(),
-            kind: ContainerNodeKind::Formatting,
+            kind: ContainerNodeKind::Formatting(
+                InlineFormatType::try_from(format).unwrap(),
+            ),
+            attrs: None,
+            children,
+            handle: DomHandle::new_unset(),
+        }
+    }
+
+    pub fn new_formatting(
+        format: InlineFormatType,
+        children: Vec<DomNode<S>>,
+    ) -> Self {
+        Self {
+            name: UnicodeString::from_str(format.tag().clone()),
+            kind: ContainerNodeKind::Formatting(format),
             attrs: None,
             children,
             handle: DomHandle::new_unset(),
@@ -101,20 +120,22 @@ where
 
         let child_index = self.children.len();
         let child_handle = self.handle.child_handle(child_index);
-        child.set_handle(child_handle);
+        child.set_handle(child_handle.clone());
         self.children.push(child);
     }
 
-    pub fn remove_child(&mut self, index: usize) {
+    pub fn remove_child(&mut self, index: usize) -> DomNode<S> {
         assert!(self.handle.is_set());
         assert!(index < self.children().len());
 
-        self.children.remove(index);
+        let ret = self.children.remove(index);
 
         for child_index in index..self.children.len() {
             let new_handle = self.handle.child_handle(child_index);
             self.children[child_index].set_handle(new_handle);
         }
+
+        ret
     }
 
     pub fn replace_child(&mut self, index: usize, nodes: Vec<DomNode<S>>) {
@@ -142,6 +163,18 @@ where
 
     pub fn last_child_mut(&mut self) -> Option<&mut DomNode<S>> {
         self.children.last_mut()
+    }
+
+    pub fn insert_child(&mut self, index: usize, node: DomNode<S>) {
+        assert!(self.handle.is_set());
+        assert!(index <= self.children().len());
+
+        self.children.insert(index, node);
+
+        for i in index..self.children.len() {
+            let new_handle = self.handle.child_handle(i);
+            self.children[i].set_handle(new_handle);
+        }
     }
 
     pub fn handle(&self) -> DomHandle {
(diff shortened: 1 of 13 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -198,6 +231,64 @@ where
             _ => false,
         }
     }
+
+    pub fn kind(&self) -> ContainerNodeKind<S> {
+        self.kind.clone()
+    }
+
+    pub fn is_same_kind(&self, other: &ContainerNode<S>) -> bool {
```
aringenbach, line 239: If I understand this correctly and its usage within `merge_children`, shouldn't we rather have some kind of `is_mergeable_with` func, as I don't expect we want to merge nodes like Links, ListItems, etc

jmartinesp: You might be right, right now on `NodeJoiner` we check that tags are the same and I thought we could replace that with this `is_same_kind` function, but a more specialised version could be better for both cases.

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -198,6 +231,49 @@ where
...
+                self.remove_child(from_index);
+            }
+            (DomNode::Text(into_node), DomNode::Text(from_node)) => {
+                let mut new_data = into_node.data().clone();
+                let from_data = from_node.data();
+                new_data.push_string(from_data);
+                into_node.set_data(new_data.clone());
+                self.remove_child(from_index);
+            }
+            _ => {}
```
andybalaam, line 274: A comment here about not merging these might be helpful

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -65,9 +66,27 @@ where
...
 
-    pub fn new_formatting(format: S, children: Vec<DomNode<S>>) -> Self {
+    pub fn new_formatting_from_tag(
+        format: S,
+        children: Vec<DomNode<S>>,
+    ) -> Self {
         Self {
             name: format.clone(),
+            kind: ContainerNodeKind::Formatting(
+                InlineFormatType::try_from(format).unwrap(),
```
andybalaam, line 76: I suggest instead of unwrapping here we should return an Option, so the caller can decide whether unwrap is the right way to handle it, because the caller knows whether the tag is known.

crates/wysiwyg/src/format_nodes.rs
```diff
@@ -0,0 +1,405 @@
...
+};
+use crate::{ComposerModel, ComposerUpdate, InlineFormatType, UnicodeString};
+
+#[derive(Eq, PartialEq, Debug)]
+enum FormatSelectionType {
+    Extend,
+    Remove,
+}
+
+impl<S> ComposerModel<S>
```
andybalaam, line 14: I'm a bit uncomfortable about providing more `impl` for `ComposerModel` in a different file, just because it might fool me later when I go looking for e.g. `format()` and don't find it in `composer_model.rs`. What do you think? I'd prefer free functions (maybe taking a `ComposerModel` as an argument, and annoyingly all having to have the generic arg specified), or a new struct like `NodeJoiner`, but am open to persuasion.

crates/wysiwyg/src/format_nodes.rs
```diff
@@ -0,0 +1,405 @@
...
+
+    fn format_several_nodes(
+        &mut self,
+        range: &MultipleNodesRange,
+        format: InlineFormatType,
+    ) {
+        let selection_type =
+            self.check_format_selection_type(&range.locations, &format);
+
+        // Start from the end so modifications to the dom doesn't conflict with next steps
```
andybalaam, line 119: Is this comment now in the wrong place?

crates/wysiwyg/src/format_nodes.rs
```diff
@@ -0,0 +1,405 @@
...
+        let selection_type = model.check_format_selection_type(
+            &range.locations,
+            &InlineFormatType::Bold,
+        );
+        assert_eq!(selection_type, FormatSelectionType::Extend);
+    }
+
+    #[test]
+    fn selection_type_remove() {
+        let model = cm("<b>hel{lo </b><b>wor}|ld</b>");
```
andybalaam, line 351: This example is weird because there are 2 bold nodes next to each other

andybalaam: (also a few more similar ones below)

crates/wysiwyg/src/format_nodes.rs
```diff
@@ -0,0 +1,405 @@
...
+            model.state.dom.to_string(),
+            "<strong>hello </strong><i><strong>wor</strong>ld</i>"
+        );
+    }
+
+    #[test]
+    fn formatting_several_nodes_works_with_same_format() {
+        let mut model = cm("{hello <b>wor}|ld</b>");
+        model.format(InlineFormatType::Bold);
+        assert_eq!(model.state.dom.to_string(), "<strong>hello world</strong>");
```
andybalaam, line 396: I am really happy it works like this (transforming b to strong in this case, but not when not needed). Awesome.

andybalaam (approved):
Looks great. Some comments but all optional.

crates/wysiwyg/src/format_nodes.rs
aringenbach: Definitely wanted to talk about this in the sync. A struct with associated functions might be fine.

Otherwise what I would expect at least is having a composer_model dedicated crate/folder with just the files that are directly tied to it. Either the multiple `impl` or the structs with associated funcs. 
I'm not sure what would be the convention in Rust but if we go with multiple `impl` files I would expect us to change naming to be something like `composer_model+format.rs`, `composer_model+lists.rs`, etc then it would be easier to browse.

crates/wysiwyg/src/dom/nodes/container_node.rs
jmartinesp: Do you mean returning `Option<Self>` here depending on whether the tag is valid or not?

crates/wysiwyg/src/format_nodes.rs
jmartinesp: I don't have a strong preference for one over the other, I just did it this way because it seemed simpler. I agree with @aringenbach though, if we do provide several impl we should probably group them in a crate so they're easy to discover.

crates/wysiwyg/src/format_nodes.rs
jmartinesp: True. I'll wrap it into another format type so we can get a `MultipleNodeRange` to verify `check_format_selection_type`.

jmartinesp:
This branch has duplicated merging logic, but it's probably better to merge now and create a new PR later to try to unify them.

PR #50, opened 2022-08-29
Fix node_joiner `remove_child` call

aringenbach:


jmartinesp (approved):
LGTM

crates/wysiwyg/src/node_joiner.rs
```diff
@@ -149,7 +149,7 @@ where
         // Delete the sibling's container
         let parent_node = dom.lookup_node_mut(parent);
         if let DomNode::Container(parent_node) = parent_node {
-            parent_node.remove(sibling.index_in_parent())
+            parent_node.remove_child(sibling.index_in_parent())
```
andybalaam, line 152: Argh, sorry!

PR #51, opened 2022-08-29
Improve list location handling & separate list actions from main model file

aringenbach:
* Move composer_model lists handling to a separate file
* Handle enter() when the selection length is > 0
* Handle slicing when enter() in a text node contained in a list item

### DIFF
--- a/crates/wysiwyg/src/composer_list_handler.rs
+++ b/crates/wysiwyg/src/composer_list_handler.rs
@@ -36,6 +36,9 @@ where
         location: usize,
         range: SameNodeRange,
     ) -> ComposerUpdate<S> {
+        // do_enter_in_list should only be called on a single location
+        // as selection can be deleted beforehand.
+        assert_eq!(range.start_offset, range.end_offset);
         // Store current Dom
         self.push_state_to_history();
         let parent_node = self.state.dom.lookup_node(parent_handle.clone());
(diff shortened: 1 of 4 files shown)


crates/wysiwyg/src/composer_list_handler.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+
+    fn slice_list_item(
+        &mut self,
+        handle: DomHandle,
+        location: usize,
+        range: SameNodeRange,
+    ) {
+        assert_eq!(range.start_offset, range.end_offset);
```
andybalaam, line 109: Why should these be equal? Maybe a comment explaining this function would help here.

andybalaam: (a doc comment I mean)

crates/wysiwyg/src/composer_list_handler.rs
```diff
@@ -0,0 +1,190 @@
...
+                        // TODO: should probably append a paragraph instead
+                        parent.append_child(DomNode::Text(TextNode::from(
+                            S::from_str("\u{200B}"),
+                        )));
+                        let new_location =
+                            Location::from(location - li_len + 1);
+                        self.state.start = new_location;
+                        self.state.end = new_location;
+                    } else {
+                        panic!("List has no parent container")
```
andybalaam, line 180: I think that should say "Parent node is not a container!"

andybalaam (approved):
Looks great! See small comments, and discussion in the room about splitting `impl` across multiple files.

crates/wysiwyg/src/composer_list_handler.rs
aringenbach: Actually it's in the wrong location, it should be in `do_enter_in_list`. I moved it there with a comment. (it's equal because we're expecting `enter` to delete current selection first then `do_enter_in_list` is called)

PR #52, opened 2022-08-30
Implement merging algorithm for structure and formatting nodes

jmartinesp:
Implemented the algorithm that @andybalaam proposed for structure nodes + a simpler one for formatting nodes:

<details>
  <summary>See structure nodes merging algorithm</summary>
There are 4 types of tag:

* structural e.g.  `<ul>, <li>, <div>`
* block formatting e.g. blockquote, code blocks
* inline formatting e.g. `<b>, <em>`
* items e.g. `<a>`

This logic should go into NodeJoiner (which is in PR #46).

1. This only applies if our selection spans multiple nodes. If we're in one node, don't do anything.
2. Remember the start text node of the selection.  Call it S.
3. Go ahead and do the replace_text, so some nodes will get deleted, but our start text node still exists.
4. Find the next text node after our start text node. Call it N. (It will probably have at least 1 character in it, so we can probably just look for the text node which corresponds to the position 1 code unit further forward.)
5. Find the lowest-level structural tag ancestor of S and N - call them struct_parent(S) and struct_parent(N). Join struct_parent(N) into struct_parent(S) - remove all struct_parent(N)'s children and add them to struct_parent(S).
6. For both S and N, find their entire ancestor list (including themself), and sort it so that the root is first.
7. Step through S and N's ancestor lists in lock-step (root first):
a. if this ancestor is also in N's ancestor list, move forward in both lists
b. if they are both containers, and their tags are the same, join them and continue stepping through the lists. To join them, append all the children of the N ancestor to the S ancestor, and delete the N ancestor. Adjust the N's ancestor list so it refers to the nodes even though they moved.
c. if they are both containers but with different tags, stop. We are done.
d. if S's ancestor is a container but N's is text (i.e. it is actually N), step forward in S's list, but not N's, so next step we will compare S's next ancestor with N again. Continue stepping.
e. if S's ancestor is text (i.e. it's S) but N's is a container, stop. We are done.
f. if both are text (i.e. we're looking at S and N) then join them: append N's text to S, and delete N. We are done.
Note: it's probably easier actually to sort the ancestors so root is first, and pop them off the list as we process them.
</summary>
</details>

For merging formatting nodes, we search for the next node again and from the root we go down the tree comparing the next node at different levels and its previous sibling, if they match, we merge.

### DIFF
--- a/crates/wysiwyg/src/node_joiner.rs
+++ b/crates/wysiwyg/src/node_joiner.rs
@@ -16,7 +16,7 @@ use std::marker::PhantomData;
 
 use crate::dom::nodes::{ContainerNodeKind, DomNode, TextNode};
 use crate::dom::{Dom, DomHandle, DomLocation, MultipleNodesRange, Range};
-use crate::{ToHtml, UnicodeString};
+use crate::UnicodeString;
 
 /// Handles joining together nodes after an edit event.
 ///
@@ -92,7 +92,7 @@ where
     }
 
     fn join_format_nodes(&self, dom: &mut Dom<S>, new_pos: usize) {
-        if let Some(mut next_node_handle) = Self::find_next_node(dom, new_pos) {
+        if let Some(next_node_handle) = Self::find_next_node(dom, new_pos) {
             self.join_format_nodes_at_level(dom, next_node_handle, 1);
         }
     }
@@ -222,7 +222,6 @@ where
                     if let DomNode::Container(parent) =
                         dom.lookup_node_mut(start_handle.parent_handle())
                     {
-                        let mut parent = parent;
                         parent.replace_child(
                             start_handle.index_in_parent(),
                             vec![text_node],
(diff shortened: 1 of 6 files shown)


crates/wysiwyg/src/node_joiner.rs
```diff
@@ -48,34 +46,285 @@ where
...
+                    if start_i.is_structure_node()
+                        && next_i.is_structure_node()
+                        && start_i.name() == next_i.name()
+                    {
+                        // Both containers with the same tag.
+                        // Move children from next to start node, remove next node.
+                        let new_index_in_parent =
+                            Self::move_children(dom, next_handle, start_handle);
+                        // This alters ancestors, so we need to re-calculate them and start again.
+                        // TODO: maybe move this to another fn?
```
andybalaam, line 188: Would be good to shorten this function.

andybalaam (approved):
Can't believe it works!

crates/wysiwyg/src/node_joiner.rs
jmartinesp: I tried extracting the re-calculation of ancestors to another function. I'm thinking about extracting text node merging too, I'll see if it's easy to do.

PR #57, opened 2022-08-30
Use Option<> to represent unset handles

andybalaam:


crates/wysiwyg/src/dom/dom_handle.rs
```diff
@@ -14,61 +14,82 @@
...
-        DomHandle::from_raw(new_path)
+    /// Returns true if this handle has been set to a value
+    pub fn is_set(&self) -> bool {
+        !self.path.is_none()
+    }
+
+    /// Returns true if this handle refers to a root node
+    /// Panics if this handle is unset.
+    pub fn is_root(&self) -> bool {
+        let path = self.path.as_ref().expect("Handle is unset!");
```
jmartinesp, line 44: Since this line is repeated in most functions, could we maybe create a small private function to replace it? So instead of:

```
let path = self.path.as_ref().expect("Handle is unset!");
path.is_empty()
```

We could write something like:

```
self.safe_path().is_empty()
```

andybalaam: Ha, I did this and then realised the function already existed! (`raw`)

PR #59, opened 2022-08-31
Remove redundant node merging strategies

jmartinesp:
Modified `NodeJoiner` so it can also join nodes after formatting is applied to several nodes.

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -115,13 +115,14 @@ where
         }
     }
 
-    pub fn append_child(&mut self, mut child: DomNode<S>) {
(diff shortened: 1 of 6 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -115,13 +115,14 @@ where
         }
     }
 
-    pub fn append_child(&mut self, mut child: DomNode<S>) {
+    pub fn append_child(&mut self, mut child: DomNode<S>) -> DomHandle {
```
jmartinesp, line 118: I also changed the return signatures of append, insert, etc. so they return the new `DomHandle`s and we don't have to re-calculate them or assume what they should be.

andybalaam (approved):

PR #60, opened 2022-08-31
Add debug tree representation

aringenbach:
Adds a Rust-based solution that allows to display the current tree


andybalaam (approved):
Looks great!

PR #61, opened 2022-09-01
Handle backspace in lists

aringenbach:
Add backspace handling code for lists and expand Rust test suite for list cases


jmartinesp (approved):
LGTM!

PR #66, opened 2022-09-02
[iOS] Move UI components ownership to hosting application

aringenbach:
Moves all buttons (bold, italic, ...) ownership from the composer lib to the hosting application


andybalaam (approved):
LTGM!

PR #68, opened 2022-09-06
Testcase generation

andybalaam:
rec-2022-09-06_10 43 14


jmartinesp:
I think this is missing the simplest test case, if I type `cool!`, in test case I get:

```
let mut model = cm("|cool!");
assert_eq!(tx(&model), "cool!|");
```

The selection indexes don't match.

Also, on webkit based browsers I can't select anything from the test case text block, the selection gets overridden and I never get to either right click and copy or even use a keyboard shortcut on time.

https://user-images.githubusercontent.com/480955/188815471-bc7aad76-f9d4-46db-a624-2799df0fbade.mov

andybalaam:
OK, the second part is https://element-io.atlassian.net/browse/PSU-753 .

But darnit, I will try to fix the first one, thanks :-)

andybalaam:
OK, my latest commit fixes the problem @jmartinesp - thank you!

PR #69, opened 2022-09-07
Android: add lists support

jmartinesp:
* Adds lists for Android (BulletSpan and OrderedListSpan).
* Creates `HtmlToSpansParser` based on the default `Html.fromHtml` implementation.
* Adds indexes mapping between the composer and the actual UI element on Android, as they diverged when we included line breaks and zero-width characters.


andybalaam (approved):
Looks good!

PR #70, opened 2022-09-07
Make it easier to build Android

andybalaam:


bindings/wysiwyg-ffi/README.md
```diff
@@ -67,13 +76,19 @@ cargo install uniffi_bindgen
 Make sure you install same version of uniffi_bindgen that is mentioned in
 `bindings/wysiwyg-ffi/Cargo.toml`, otherwise you will get weird errors.
 
-* Build shared object:
+* Build the library using Gradle:
+
+```bash
+ANDROID_SDK_ROOT=/home/andy/AndroidSdk make android
```
jmartinesp, line 82: I would normally set this variable as an env var in `.bashrc`, but this is probably fine.

bindings/wysiwyg-ffi/README.md
```diff
@@ -12,9 +12,18 @@ wysiwyg-rust.
   android-ndk-r22b-linux-x86_64.zip and unzip it e.g.:
 
 ```bash
-unzip android-ndk-r22b-linux-x86_64.zip
+cd $ANDROID_SDK_HOME
```
jmartinesp, line 15: It seems like there are several variables that can be used here:

- ANDROID_SDK_ROOT
- ANDROID_SDK_HOME
- ANDROID_HOME

It seems like the proper one is `ANDROID_HOME`, the other 2 are deprecated. There is also `ANDROID_NDK_ROOT` that we could have tried before. It's probably worth to mention that this variable should be set somewhere before we try to use it here.

bindings/wysiwyg-ffi/README.md
jmartinesp: In the pre-requisites we're already using `ANDROID_HOME` or a similar variable, so we assume it already exists. There's probably no need to manually set it here too.

jmartinesp (approved):
Couple of changes, other than that, LGTM.

bindings/wysiwyg-ffi/README.md
andybalaam: Thanks, I updated to use ANDROID_HOME

platforms/android/library/build.gradle
```diff
@@ -54,6 +54,8 @@ android {
     viewBinding {
         enabled = true
     }
+
+    ndkVersion = "22.1.7171670"
```
andybalaam, line 58: @jmartinesp this part was particularly unsure for me. I needed it for the build to work, but is it right?

jmartinesp: In my case gradle seems to just ignore it, I don't have that version installed and it's building just fine. I guess it's using what's on `bindings/wysiwyg-ffi/.cargo/config.toml` instead. It's weird that you have to add it to be able to build the lib.

PR #71, opened 2022-09-07
Handle list toggling & button states

aringenbach:
* Add the possibility to toggle between list types
* Handle inserting in previous list (for now, with simple nodes)
* Handle button state for lists
* Update list binding functions signature to reflect that it doesn't always create a list

Simulator Screen Recording - iPhone 13 - 2022-09-07 at 17 10 45


crates/wysiwyg/src/composer_model/menu_state.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub(crate) fn compute_menu_state(&mut self) -> MenuState {
+        let (s, e) = self.safe_selection();
+        let range = self.state.dom.find_range(s, e);
+        match range {
+            Range::SameNode(range) => {
+                return self.compute_active_buttons(range.node_handle);
+            }
+            Range::MultipleNodes(range) => {
+                return self.compute_active_buttons(
+                    // TODO: lookup all locations and output union instead
```
andybalaam, line 36: I'm not actually sure what the right behaviour is here - we should compare with other editors.

aringenbach, line 36: *intersection is meant here obviously

crates/wysiwyg/src/tests/test_lists.rs
```diff
@@ -112,11 +112,34 @@ fn entering_mid_text_node_with_selection() {
...
+    assert_eq!(tx(&model), "<ol><li>ab</li></ol>cd|");
+}
+
+#[test]
+fn appending_new_list_to_previous() {
+    let mut model = cm("<ol><li>ab</li></ol>cd|");
+    model.ordered_list();
+    assert_eq!(tx(&model), "<ol><li>ab</li><li>cd|</li></ol>");
+}
+
```
andybalaam, line 142: This all looks great, but there is one testcase that I think it currently does wrong. Here is my attempt at expressing it in Rust:

```rust
let mut model = cm("<ul><li>foo|</li></ul>bar");
model.ordered_list();
assert_eq!(tx(&model), "<ul><li>foo</li></ul><ol><li>​bar|</li></ol>");
```

Currently it ends up like:

```rust
assert_eq!(tx(&model), "<ol><li>foo</li><li>​bar|</li></ol>");
```

I.e. it changes the preceding list into `ol`. I think it should not do that, but instead make a new `ul`.

andybalaam (changes requested):
One behaviour I think is not right, but looks great!

crates/wysiwyg/src/tests/test_lists.rs
andybalaam: (I just confirmed in Google Docs it works the way I think we should work)

crates/wysiwyg/src/composer_model/menu_state.rs
aringenbach: I've self reviewed up here ^ I meant intersection, from looking at some others editors, same rule applies to list than the one for formatting. 
Button is active if and only if the entire selection is inside a list of expected type
For example, no active button with this:
image

andybalaam: I think the button state should exactly match what will happen then I click it.
So if the button is not highlighted, clicking it will add (or replace) a list, and if it is highlighted, clicking it will remove a list.

aringenbach: Will have to check what happens when this is implemented for us.
In my screenshot here if I hit one of the list buttons, it will transform the "unmatching" list to the same type and merge which looks like a correct behaviour to me.

crates/wysiwyg/src/tests/test_lists.rs
aringenbach: I added this test case (slightly different as we don't insert ZWSP right now - better do it on a separated piece of work since it might be tricky and mess up selection)

andybalaam (approved):
Awesome!

PR #75, opened 2022-09-08
Support inserting br tags when you press enter.

andybalaam:


aringenbach (commented):
Looks pretty good.
Actually I'm currently wondering how much of the join_nodes utility could/should be defined on dom rather than the compoer_model 🤔

andybalaam:
I actually think this might break Android and iOS. I'll chat in the channel, and mark this as draft in the meantime

PR #76, opened 2022-09-09
Skip the selection behavior when the selection is not in the composer

florianduros:
In the Web example app, when we select some text outside the composer, the selection is updated to (-1, -1), which is wrong.

We also lose our selection because we grab focus by replacing the editor contents.

When the selection is outside the composer, we should do nothing.


jmartinesp (approved):
LGTM! If only all PRs were this simple 😄

PR #77, opened 2022-09-09
Disable removal of empty text nodes for formatting, as it broke some functionality

jmartinesp:
Also, fixed issue in `tx(&model)`.


andybalaam (approved):
Thanks!

PR #78, opened 2022-09-09
Handle active button state for formatting & cross-node selections

aringenbach:
* Handles cross node selection to set expected buttons as active
* Handles active/inactive states for formatting buttons

Simulator Screen Recording - iPhone 13 - 2022-09-09 at 14 03 40


andybalaam (approved):
Nice!

PR #79, opened 2022-09-09
Handle disabled state for undo/redo buttons

aringenbach:
Disabled state for Undo/Redo buttons

Simulator Screen Recording - iPhone 13 - 2022-09-09 at 15 29 22

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -18,6 +18,13 @@ impl ComposerModel {
         }
     }
 
+    pub fn set_html(self: &Arc<Self>, html: String) -> Arc<ComposerUpdate> {
+        let html = Utf16String::from_str(&html);
+        Arc::new(ComposerUpdate::from(
+            self.inner.lock().unwrap().replace_all_html(&html),
+        ))
+    }
+
     pub fn select(
         self: &Arc<Self>,
         start_utf16_codeunit: u32,
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -18,7 +18,7 @@ impl ComposerModel {
         }
     }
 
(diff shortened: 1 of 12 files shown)


bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -18,6 +18,13 @@ impl ComposerModel {
         }
     }
 
+    pub fn set_html(self: &Arc<Self>, html: String) -> Arc<ComposerUpdate> {
```
andybalaam, line 21: I went for replace_all_html to make it clear the whole thing is replaced. I think we should try to be consistent - if you think set_html is better, maybe we should change everything to that?

andybalaam (changes requested):
One question/comment, looks very good otherwise.

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
aringenbach: Yeah `replace_all_html` seems fine to me, I wonder if this should also reset undo/redo states (perhaps with an extra parameter), as I suppose there might be cases where we either want to push a state to history or we want to completely remove old states (right now it does nothing - which means that if we play with the editor, then replace_all_html, undo we will get to state at N-2).

andybalaam: Hmm, good point. What do we want right now? We should implement the behaviour we want now (either delete history or add this to history properly), and add an option later.

aringenbach: I suppose right now we want to simply clear history, as I don't expect to do something other than setting up the initial HTML at start.
Still not sure there is a real case for preserving undo. (perhaps something around switches between plain/WYSIWYG, if we want to save or keep the model alive in these kind of cases)

andybalaam: Yeah, let's leave that use case until we need it, and clear the undo/redo history for now.

andybalaam (approved):
Great, thanks!

PR #80, opened 2022-09-12
Add test case for creating a new list immediately after an old one join

aringenbach:
Add a test case that proves joining list to previous works


andybalaam (approved):
Great, thanks!

PR #81, opened 2022-09-12
Convert ZWSP in tree representation

aringenbach:
Simulator Screen Shot - iPhone 13 - 2022-09-12 at 10 47 35


andybalaam (approved):
Great, sorry I missed this.

PR #82, opened 2022-09-13
Compute manually the focus offset from the current node

florianduros:
I add a small recursive function to compute manually the focus offset.

The original offset has an unexpected value in case of a `ctrl+a` in Firefox. 

The focus has  a different behaviour between Chrome and Firefox.
- Chrome => focus  on the text node
- Firefox => focus on the parent div `contentEditable=true`

It clearly needs unitary tests which will arrive _soon_ in the `react` version :)


andybalaam (changes requested):
Did you know we do have tests?

They are very manual, but they are there, in tests.js

Would be great to add some tests for this.

florianduros:
I added some tests.

PR #85, opened 2022-09-13
Use reference in lookup_node_mut

florianduros:
I would like Dom.lookup_node and Dom.lookup_node_mut to take an argument of type &DomHandle instead of DomHandle.

Then everywhere they are called, we don't need to call handle.clone() before we call them. Instead just &handle.


andybalaam (approved):
Nice, thanks!

PR #86, opened 2022-09-13
Insert br on enter

andybalaam:
Probably helpful to look commit by commit, but the main one is quite big, sorry :-(

This makes newlines work on Web in quite a few places, but there are definite gaps, hopefully mostly covered by the ignored tests I have included.


crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -201,11 +211,91 @@ where
...
+                    Where::After
+                } else {
+                    Where::During
+                }
+            }
+        };
+
+        match wh {
+            Where::Before => {
+                self.parent(handle).insert_child(0, new_node);
```
aringenbach, line 263: Shouldn't the index here be index_in_parent rather than 0 ?

crates/wysiwyg/src/tests/test_to_tree.rs
```diff
@@ -18,35 +18,39 @@ use crate::tests::testutils_composer_model::cm;
 use crate::ToTree;
 
 #[test]
-fn computing_tree() {
+fn single_nested_tag_produces_tree() {
     let model = cm("<b>abc<i>def</i></b>|");
     assert_eq!(
         model.state.dom.to_tree(),
-        "
+        r#"
```
aringenbach, line 25: Pretty cool

aringenbach (approved):
A small comment inline. 
Otherwise looks pretty good. 

I wonder if we will get consistent **paste** results across platforms, we might have to expect this to pass:

```rust
let mut model = cm("|")
replace_text(&mut model, "abc\ndef");
assert_eq!(tx(&model), "abc<br />def|");
```

crates/wysiwyg/src/dom/dom_struct.rs
andybalaam: Good spot, fixed - thanks.

andybalaam:
> A small comment inline. Otherwise looks pretty good.
> 
> I wonder if we will get consistent **paste** results across platforms, we might have to expect this to pass:
> 
> ```rust
> let mut model = cm("|")
> replace_text(&mut model, "abc\ndef");
> assert_eq!(tx(&model), "abc<br />def|");
> ```

Yeah, we should check the behaviour. I'll make a task

PR #87, opened 2022-09-13
Handle un-formatting nodes

aringenbach:
Allow to un-format selected nodes

Simulator Screen Recording - iPhone 13 - 2022-09-13 at 17 46 18

### DIFF
--- a/crates/wysiwyg/src/composer_model/format.rs
+++ b/crates/wysiwyg/src/composer_model/format.rs
@@ -155,7 +155,7 @@ where
             }
             self.state.dom.replace(&range.node_handle, new_nodes);
         } else {
-            panic!("Trying to bold a non-text node")
+            panic!("Trying to format a non-text node")
         }
     }
 
@@ -279,7 +279,7 @@ where
             if let DomNode::Container(node) = node {
                 match node.kind() {
                     ContainerNodeKind::Formatting(f) => {
-                        if f.clone() == format.clone() {
+                        if *f == format {
                             let index_in_parent =
                                 location.node_handle.index_in_parent();
                             let parent_node = self.state.dom.lookup_node_mut(
@@ -290,6 +290,8 @@ where
                                 parent_node.remove_child_and_own_hierarchy(
                                     index_in_parent,
                                 );
+                            } else {
+                                panic!("Parent node is not a container")
                             }
 
                             // Re-apply formatting to slices before and after the selection if needed
@@ -340,7 +342,7 @@ where
         if let DomNode::Container(container) = node {
             match container.kind() {
                 ContainerNodeKind::Formatting(f) => {
-                    if f.clone() == format {
+                    if *f == format {
                         handle
                     } else {
                         self.find_parent_formatting_node(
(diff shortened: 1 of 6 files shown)


crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -190,6 +267,98 @@ where
...
+        end: usize,
+        range: &MultipleNodesRange,
+        format: InlineFormatType,
+    ) {
+        for location in range.locations.iter() {
+            let node = self.state.dom.lookup_node(&location.node_handle);
+            if let DomNode::Container(node) = node {
+                match node.kind() {
+                    ContainerNodeKind::Formatting(f) => {
+                        if f.clone() == format.clone() {
```
andybalaam, line 282: Should be able to do if `f == format` without cloning, or something similar.

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -140,6 +140,20 @@ where
         ret
     }
 
+    pub fn remove_child_and_own_hierarchy(
```
andybalaam, line 143: Can we find a better name e.g. `remove_child_and_take_grandchildren` ? (Or something better??)

andybalaam (commented):
Looks good so far.

crates/wysiwyg/src/composer_model/format.rs
jmartinesp, line 282: This seems to work fine:

```suggestion
                        if *f == format {
```

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -190,6 +267,98 @@ where
...
+    fn find_parent_formatting_node(
+        &self,
+        handle: DomHandle,
+        format: InlineFormatType,
+    ) -> DomHandle {
+        let node = self.state.dom.lookup_node(&handle);
+        if let DomNode::Container(container) = node {
+            match container.kind() {
+                ContainerNodeKind::Formatting(f) => {
+                    if f.clone() == format {
```
jmartinesp, line 343: Same here, `*f` is probably enough to use the reference's contents.

```suggestion
                    if *f == format {
```

jmartinesp:
I'm actually surprised at how little code it took to write `unformat_several_nodes` by just removing the formatting node altogether and then re-creating it. I spent several days in this banging my head before giving up and this solution is *so simple*. Congrats!

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -89,6 +89,25 @@ where
...
+    /// moves them to its parent container children.
+    pub fn remove_and_keep_children(&mut self, node_handle: &DomHandle) {
+        let parent_node = self.lookup_node_mut(&node_handle.parent_handle());
+        match parent_node {
+            DomNode::Container(parent) => {
+                let index = node_handle.index_in_parent();
+                let node = parent.remove_child(index);
+                if let DomNode::Container(node) = node {
+                    for child in node.children().iter().rev() {
+                        parent.insert_child(index, child.clone());
```
aringenbach, line 102: I really wish we could avoid this clone, otherwise it is not really better than cloning the entire children array and using existing `replace` fn

jmartinesp (approved):
LGTM! It also seems to work great on Android.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -449,4 +596,56 @@ mod test {
         model.format(InlineFormatType::Bold);
         assert_eq!(model.state.dom.to_string(), "<strong>hello world</strong>");
     }
+
+    #[test]
```
andybalaam, line 600: I'd like to see a lot more tests in `test_formatting.rs` - then this looks good to go!

andybalaam (changes requested):
Just more tests please!

crates/wysiwyg/src/tests/test_formatting.rs
```diff
@@ -34,21 +34,85 @@ fn selecting_and_bolding_multiple_times() {
...
+}
+
+#[test]
+fn unformat_toplevel_node_moves_nested_nodes() {
+    let mut model = cm("aa<em>{b<strong>bc}|</strong></em>c");
+    model.italic();
+    assert_eq!(tx(&model), "aa{b<strong>bc}|</strong>c");
+}
+
+#[ignore] // TODO: partial_unformat_toplevel_node_reconstructs_expected_model
```
aringenbach, line 87: pretty sure this is fixed by removing SameNode

andybalaam (approved):
Looks great!

PR #88, opened 2022-09-14
Fix list focus when the editor is empty

florianduros:
Clicking "list" when the composer is empty loses focus. Focus should stay in the composer.

### DIFF
--- a/platforms/web/example/test.js
+++ b/platforms/web/example/test.js
@@ -146,7 +146,14 @@ run_tests([
         assert_eq(offset, 0);
     }},
 
-    { name: "node_and_offset finds inside an list", test: () => {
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks good, and if it works, it works, but I am wondering whether this is something more general we should deal with, and what happens when we have two adjacent empty list items? Does the focus always snap to the first one?

florianduros:
> Looks good, and if it works, it works, but I am wondering whether this is something more general we should deal with, and what happens when we have two adjacent empty list items? Does the focus always snap to the first one?

Good question about the two adjacents lists, "normally" the algorithm will return the first list. 
Let me test this one before I merged this PR :)

florianduros:
I added a test with the two empty list case and it works :)

PR #89, opened 2022-09-14
Web: add button for inline code

florianduros:
Add a button in the toolbar for inline code

The code [code html element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code) doesn't have a input type (https://rawgit.com/w3c/input-events/v1/index.html#interface-InputEvent-Attributes) like bold, italic...

So instead of using an input event (no inputType is fitting), I'm using a custom event.

The main drawback is that the code tag is not inserted by the browser but by the return of the composer model.

Nice sandbox to play with input event https://d-toybox.com/studio/lib/input_event_viewer.html

### DIFF
--- a/platforms/web/example/index.html
+++ b/platforms/web/example/index.html
@@ -134,22 +134,23 @@ h2 {
 <div>
 <div class="editor_container">
 <div class="editor_toolbar">
-    <button id="button_undo" href="#"><img src="images/undo.svg"/></button>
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Looks like a nice way to do it. I'd prefer an SVG of the button image, but no need to do that right now.

Thanks!

PR #92, opened 2022-09-14
Lists indentation

jmartinesp:
Add support to indent and un-indent list items.

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -147,6 +147,14 @@ impl ComposerModel {
         ))
     }
 
+    pub fn indent(self: &Arc<Self>) -> Arc<ComposerUpdate> {
+        Arc::new(ComposerUpdate::from(self.inner.lock().unwrap().indent()))
+    }
+
+    pub fn un_indent(self: &Arc<Self>) -> Arc<ComposerUpdate> {
+        Arc::new(ComposerUpdate::from(self.inner.lock().unwrap().un_indent()))
+    }
+
     pub fn to_tree(self: &Arc<Self>) -> String {
         self.inner.lock().unwrap().to_tree().to_string()
     }
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -152,7 +152,7 @@ impl ComposerModel {
     }
 
     pub fn un_indent(self: &Arc<Self>) -> Arc<ComposerUpdate> {
-        Arc::new(ComposerUpdate::from(self.inner.lock().unwrap().un_indent()))
+        Arc::new(ComposerUpdate::from(self.inner.lock().unwrap().unindent()))
     }
 
     pub fn to_tree(self: &Arc<Self>) -> String {
--- a/crates/wysiwyg/src/composer_model/lists.rs
+++ b/crates/wysiwyg/src/composer_model/lists.rs
@@ -385,7 +385,7 @@ where
         false
     }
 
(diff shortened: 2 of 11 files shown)


crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -357,4 +359,464 @@ where
...
+            self.find_closest_list_ancestor(&handle)
+        {
+            self.find_closest_list_ancestor(&closest_list_handle)
+                .is_some()
+        } else {
+            false
+        }
+    }
+
+    fn find_closest_list_ancestor(
```
andybalaam, line 409: This and `find_parent_list_item` in `dom_struct.rs` should be near each other. I'm not actually sure where...maybe even in a separate module if we think there will be enough of them?

crates/wysiwyg/src/dom/dom_handle.rs
```diff
@@ -112,3 +114,33 @@ impl DomHandle {
...
 }
+
+impl PartialOrd<Self> for DomHandle {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for DomHandle {
+    fn cmp(&self, other: &Self) -> Ordering {
```
andybalaam, line 125: Is this equivalent to `self.raw.cmp(other.raw)`? I assume this exists and is a lexical compare?

andybalaam (changes requested):
Overall looks great. I really want that refactor so ComposerModel gets simpler, but that is not part of this change.

crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -357,4 +359,464 @@ where
...
+    fn un_indent_handles(&mut self, handles: Vec<DomHandle>) {
+        let by_list_sorted = Self::group_sorted_handles_by_list_parent(handles);
+
+        for (list_handle, handles) in by_list_sorted.iter().rev() {
+            let mut sorted_handles = handles.clone();
+            sorted_handles.sort();
+
+            let at_index = list_handle.parent_handle().index_in_parent() + 1;
+            if let Some(into_handle) =
+                self.find_closest_list_ancestor(list_handle)
```
andybalaam, line 617: Somewhere near here the compile failed because I changed some interfaces to take `&DomHandle`, sorry!

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -147,6 +147,14 @@ impl ComposerModel {
         ))
     }
 
+    pub fn indent(self: &Arc<Self>) -> Arc<ComposerUpdate> {
+        Arc::new(ComposerUpdate::from(self.inner.lock().unwrap().indent()))
+    }
+
+    pub fn un_indent(self: &Arc<Self>) -> Arc<ComposerUpdate> {
```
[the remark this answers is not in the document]
jmartinesp: I actually used `unindent` everywhere at the beginning, but my IDE was complaining every single time about it, so I changed it to `un_indent` 😅 . I can revert that, no worries.

crates/wysiwyg/src/dom/dom_handle.rs
jmartinesp: I'd have to double check. I implemented this myself because I didn't trust the `Vec::cmp` default implementation to work as we expected, but if it does it would simplify everything.

crates/wysiwyg/src/composer_model/lists.rs
jmartinesp: Will rebase and fix it 👍

andybalaam (approved):
Awesome!

PR #94, opened 2022-09-15
[iOS] Display inline code formatting

aringenbach:
Enable basic inline code display on iOS

### DIFF
--- a/platforms/ios/example/Shared/WysiwygAction+Utils.swift
+++ b/platforms/ios/example/Shared/WysiwygAction+Utils.swift
@@ -56,7 +56,7 @@ extension WysiwygAction: CaseIterable, Identifiable {
         case .underline:
             return .underlineButton
         case .inlineCode:
-            return .inlineCode
+            return .inlineCodeButton
         case .undo:
             return .undoButton
         case .redo:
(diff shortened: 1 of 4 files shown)


platforms/ios/example/Shared/WysiwygAction+Utils.swift
```diff
@@ -55,6 +55,8 @@ extension WysiwygAction: CaseIterable, Identifiable {
             return .strikeThroughButton
         case .underline:
             return .underlineButton
+        case .inlineCode:
+            return .inlineCode
```
jmartinesp, line 59: I think this needs to be:

```suggestion
            return .inlineCodeButton
```

It wouldn't let me build the project any other way.

jmartinesp (changes requested):
Had some trouble building it as is. I added a comment with the change I had to make to build the project.

platforms/ios/example/Shared/WysiwygAction+Utils.swift
aringenbach: My bad, didn't commit this as I was doing things with links in the same branch 🙈

jmartinesp (approved):
After the changes it seems to build and work fine. LGTM!

PR #95, opened 2022-09-16
[Android]: Add menu state listeners.

jmartinesp:
Also, refactor code:

* Extract RichTextEditor to the example module.
* Add extra tests for EditorEditText.


aringenbach (approved):
LGTM

PR #96, opened 2022-09-16
Small fixes to comments and tests

andybalaam:
I suggest reviewing each commit separately.


jmartinesp:
If `find_parent_list_item` will now return the passed `child_handle` as a possible result, we might want to find a name that reflects this change.

PR #97, opened 2022-09-16
Allow writing examples with leading cursor br tag

andybalaam:


crates/wysiwyg/src/composer_model/example_format.rs
```diff
@@ -368,41 +368,58 @@ impl SelectionWritingState {
...
+        let mut do_first = !self.done_first && self.first < self.current_pos;
 
         // If we just passed last or we're at the end, write out }
         let do_last = !self.done_last
             && (self.last <= self.current_pos
                 || self.current_pos == self.length);
 
+        // In some weird circumstances with empty text nodes, we might
+        // do_last when we haven't done_first, so make sure we do_first too.
+        if do_last && !self.done_first {
```
jmartinesp, line 380: Should we add a TODO to delete this when we properly handle empty text nodes?

jmartinesp (approved):
Added a comment, but it's just a suggestion. LGTM!

crates/wysiwyg/src/composer_model/example_format.rs
andybalaam: I'm not 100% sure we will want to delete it, and this is pretty much test code, so I'll avoid adding yet more TODOs in this case, I think.

PR #99, opened 2022-09-19
Add a select API to text update

aringenbach:
This adds a new type of text update, that only changes the current selection.


jmartinesp (approved):
LGTM!

PR #101, opened 2022-09-19
Allow creation of empty format nodes

jmartinesp:
Empty text nodes could only be created at the index 0 of the composer before. With this workaround they can be created anywhere.

### DIFF
--- a/crates/wysiwyg/src/composer_model/format.rs
+++ b/crates/wysiwyg/src/composer_model/format.rs
@@ -140,11 +140,17 @@ where
             let text = t.data();
             // TODO: can we be globally smart about not leaving empty text nodes ?
             let before = slice_to(text, ..range.start_offset);
-            let during = slice(text, range.start_offset..range.end_offset);
+            let mut during = slice(text, range.start_offset..range.end_offset);
             let after = slice_from(text, range.end_offset..);
             let mut new_nodes = Vec::new();
             if before.len() > 0 {
                 new_nodes.push(DomNode::new_text(before));
+            }
+            if during.len() == 0 {
+                // TODO: Find a more generic way to do this.
+                // TextNodes should probably never be created empty
+                during = UnicodeString::from_str("\u{200B}");
+                self.state.end += 1;
             }
             new_nodes.push(DomNode::new_formatting(
                 format,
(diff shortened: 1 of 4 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -141,7 +141,7 @@ class EditorEditText : TextInputEditText {
         if (result != null) {
             editableText.clear()
             editableText.replace(0, editableText.length, result.text)
-            setSelectionFromComposerUpdate(result.selection.last)
+            setSelectionFromComposerUpdate(result.selection.first, result.selection.last)
```
jmartinesp, line 144: This is needed to get it properly working on Android.

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -149,7 +149,10 @@ where
     let container_end = *offset;
     let container_node_len = container_end - container_start;
     // We never want to return the root node
-    if !node.handle().is_root() {
+    if container_end >= start
+        && container_start <= end
+        && !node.handle().is_root()
+    {
```
jmartinesp, line 155: This is not really needed for the PR, but I just realised we were always adding *any* container node as part of the selection. It didn't matter much in general because we always checked the `is_leaf` nodes first, but this is more correct.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -648,4 +666,13 @@ mod test {
         model.strike_through();
         assert_eq!(model.state.dom.to_string(), "<del>a</del>bcd<del>ef</del>");
     }
+
+    #[test]
+    fn format_and_unformat_empty_selection() {
+        let mut model = cm("AAA |");
+        model.bold();
+        assert_eq!(tx(&model), "AAA <strong>{~}|</strong>");
```
aringenbach, line 674: What if I change selection from this state and move back to it afterwards ? 
I suppose we need to make Rust "cheat" a bit so it selects the `~` when needed ? 
Probably worth an additional test.

aringenbach (approved):
LGTM 

One comment inline, it can be treated in another PR depending on how complicated it is.

crates/wysiwyg/src/composer_model/format.rs
jmartinesp: I tried to get `replace_same_node` to remove zero-width whitespaces at the beginning of a TextNode, but that broke list, so I had to revert it. It's probably worth its own PR.

PR #102, opened 2022-09-20
Remove SameNodeRange implementation of to_example_format

andybalaam:
First small steps towards Removing samenoderange


aringenbach (approved):
LGTM!

PR #109, opened 2022-09-20
[iOS] Extend and normalize log messages

aringenbach:
* Slightly extend iOS log messages
* Add normalised log messages for consistency


jmartinesp (approved):
LGTM!

PR #110, opened 2022-09-20
[iOS] Remove workaround for linebreaks

aringenbach:
Removes a now unnecessary workaround that replaced "\n" with `br` tags


jmartinesp (approved):
LGTM

PR #111, opened 2022-09-20
[iOS] Fix test target build

aringenbach:
Fixes the build fro the test target with TextUpdate's `select` case.


jmartinesp (approved):
LGTM

PR #112, opened 2022-09-20
[iOS] Minor refactors

aringenbach:
* Rename files and move some code
* Added some missing comments
* Added a helper for creating initial ranges


platforms/ios/example/Wysiwyg/Extensions/UIAlertController.swift
```diff
@@ -0,0 +1,32 @@
...
+// limitations under the License.
+//
+
+import UIKit
+
+extension UIAlertController {
+    convenience init(alert: AlertConfig) {
+        self.init(title: alert.title, message: nil, preferredStyle: .alert)
+        addTextField()
+        //addTextField { $0.placeholder = alert.placeholder }
```
jmartinesp, line 23: Should we restore it or remove it?

jmartinesp (approved):
Small comment, but nothing important. LGTM!

platforms/ios/example/Wysiwyg/Extensions/UIAlertController.swift
aringenbach: Yeah we definitely don't use this, thanks, I removed it.

PR #115, opened 2022-09-21
[iOS] Fix keyboard autocomplete

aringenbach:
Fixes most issues on iOS keyboard autocompletion.
There is still a pending issue on standard text input with a specific location that is blocked for selection in some cases.


andybalaam (approved):
Looks good!

aringenbach:
Replaced by #116, which is unfortunately not really testable but seems to work way better.

PR #116, opened 2022-09-22
[iOS] Fix keyboard autocomplete

aringenbach:
Fixes iOS issues with keyboard autocomplete by accepting input changes and reconciliate the UITextView with the Rust model when needed.


andybalaam (approved):
Looks good!

PR #124, opened 2022-09-22
Support multiple whitespaces

jmartinesp:
* Added `&nbsp;` support to our the HTML output.
* Re-wrote the way that selection is written to the output in `to_example_format`.
* Added some extra helper method for tests `restore_whitespace` and `restore_whitespace_u16` so those `&nbsp;` are translated back to simple whitespaces and the strings containing can be used in assertions with no issues.

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/text_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/text_node.rs
@@ -61,6 +61,50 @@ where
     pub fn set_handle(&mut self, handle: DomHandle) {
         self.handle = handle;
     }
+
+    fn handle_several_whitespaces(
+        formatter: &mut HtmlFormatter<S>,
+        pos: usize,
+        is_last_node_in_parent: bool,
+    ) {
+        let mut ranges_to_replace = Vec::new();
+        let mut current_range = usize::MAX..usize::MAX;
+        let mut whitespaces: usize = 0;
+        let mut needs_to_replace = false;
+        let w = S::c_from_char(' ');
+        let nbsp = S::c_from_char('\u{A0}');
+        for (i, c) in formatter.chars_from(pos..).iter().enumerate() {
+            if *c == w || *c == nbsp {
+                if *c == w {
+                    needs_to_replace = true;
+                }
+                if current_range.start == usize::MAX {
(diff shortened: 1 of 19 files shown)


andybalaam (approved):
Looks good, sorry for the merge hell!

crates/wysiwyg/src/dom/nodes/text_node.rs
```diff
@@ -60,17 +61,71 @@ where
...
+        formatter: &mut HtmlFormatter<S>,
+        pos: usize,
+        is_last_node_in_parent: bool,
+    ) {
+        let mut ranges_to_replace = Vec::new();
+        let mut current_range = usize::MAX..usize::MAX;
+        let mut whitespaces: usize = 0;
+        let mut needs_to_replace = false;
+        let w = S::c_from_char(' ');
+        let nbsp = S::c_from_char('\u{A0}');
```
jplatte, line 75: Hey @jmartinesp, I'm currently working on a refactoring of the `UnicodeString` trait and during that, noticed that this is a problematic use of `c_from_char`. It will panic if `S` is `String` because with `CodeUnit = u8`, the nbsp character doesn't fit inside a single code unit.

jplatte: I think the right fix would be to change the inner buffer of `HtmlFormatter` to be `S` instead of `Vec<S::CodeUnit>`, and update `chars_from` to iterate over `char`s instead of code units here. Let me know whether you think that is a good idea. I'm also reachable on Matrix (@jplatte:flipdot.org).

jmartinesp: Actually, the `let mut whitespaces: usize = 0;` used here kind of works around either this issue or a similar one since I remember facing panics for utf8 strings, but it's probably better to re-write HtmlFormatter. 

@andybalaam since you wrote most of it and the abstractions around strings, do you think it's worth it?

andybalaam: I don't have full context, but I think using `S` as the buffer of `HtmlFormatter` sounds right.

PR #126, opened 2022-09-23
[iOS] Enable Swiftlint & SwiftFormat from the example app

aringenbach:
Enables both tools to be run on any application build. (linting and formatting code for both the library and the app)
Added additional setup information in example app README.


andybalaam (approved):
Good to do, thanks!

PR #127, opened 2022-09-23
Add script to push ios library to the package repo

langleyd:
- Adds a script that opens a PR against the package repo with the latest iOS lib
- Creating a PR means we can test the changes against the PR branch before merging
- Can add versioning/tags/release int he future but not sure we need that yet

### DIFF
--- a/release_ios.sh
+++ b/release_ios.sh
@@ -9,12 +9,12 @@ else
   exit 1
 fi
 
-# if [ -z "$(git status --porcelain)" ]; then 
-#   echo "Working directory is clean."
-# else 
-#   echo "Working directory is not clean. Exiting..."
-#   exit 1
-# fi
+if [ -z "$(git status --porcelain)" ]; then 
+  echo "Working directory is clean."
+else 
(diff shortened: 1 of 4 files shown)


release_ios.sh
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+if [ -z "$(git status --porcelain)" ]; then 
+  echo "Working directory is clean."
+else 
+  echo "Working directory is not clean. Exiting..."
+  exit 1
+fi
+
+# Complete any prerequisates as defined in /bindings/wysiwyg-ffi/README.md#ios
```
aringenbach, line 19:
```suggestion
# Complete any prerequisites as defined in /bindings/wysiwyg-ffi/README.md#ios
```

aringenbach (approved):
One small typo, otherwise LGTM. 

Might be useful to add some information in the main README.md on how to release the Swift Package (it's pretty straight forward at the moment, but it'll be a good starting point for documentation when adding versions etc)

PR #129, opened 2022-09-26
Refined action type

florianduros:
Remove remaining any type and add typeguard for select tuple


andybalaam (approved):
Wow, sorry for writing such a nasty type! Feel free to change it to be less weird.

florianduros:
> Wow, sorry for writing such a nasty type! Feel free to change it to be less weird.

I was the one to write this TS type :)

I fine with tuple for typing the action

andybalaam:
I mean I wrote the code that caused us to need such a complex type, .... right? Anyway, thanks!

PR #131, opened 2022-09-26
Fix enter with no text node

aringenbach:
Fixes an issue with pressing enter when no text node are detected at current selection (completely empty model)
Add a non-passing test for similar crash with doubled empty text node after backspacing on model containing only a line break.


andybalaam (approved):
Looks good, thanks!

PR #132, opened 2022-09-26
[iOS] Run iOS tests on CI

aringenbach:
Sets up iOS tests to run on CI


andybalaam (approved):
Looks good except the sleeps :vomiting_face: 

But failing right now, so maybe more work to do on it?

PR #133, opened 2022-09-26
Forbid empty text nodes

jmartinesp:
Any TextNode created should either be non-empty or contain a ZWSP.

I also tried to minimise the amount of empty text nodes added while replacing text, and tried replacing `\u{200B}` where possible with `ZWSP` constant.


jmartinesp:
Closing this as we'd reduced quite a bit the scope of what should be done for this task.

PR #136, opened 2022-09-27
More UnicodeString refactoring

jplatte:
Text node escaping doesn't do exactly the same thing as before, but should create no worse results (see code comments).


andybalaam (approved):
Looks great, but I think it needs rebasing on #133.

andybalaam (approved):

PR #137, opened 2022-09-27
Remove unnecessary empty text nodes.

jmartinesp:
Try to join sibling text nodes after replacing text.


andybalaam (approved):
Looks good! Will be better when I get my refactor in...

PR #139, opened 2022-09-27
Add menu states to all buttons in the example UI

jmartinesp:
I forgot to add the tint attribute to other buttons besides bold, undo and redo.


aringenbach (approved):
LGTM

PR #140, opened 2022-09-28
test: Move `#![cfg(test)]` on test module, rather than submodule

Hywan:
Instead of having `#![cfg(test)]` on all test files, we can put it directly in the `src/tests.rs` module.


aringenbach (approved):
Nice 👍

PR #141, opened 2022-09-28
Fix merge formatting nodes crash by keeping track of moved handles

jmartinesp:


crates/wysiwyg/src/composer_model/join_nodes.rs
```diff
@@ -271,7 +299,8 @@ where
         &mut self,
         from_handle: &DomHandle,
         to_handle: &DomHandle,
-    ) -> usize {
+    ) -> (usize, HashMap<DomHandle, DomHandle>) {
```
jmartinesp, line 302: I don't really like the idea of returning the moved handles here, but I couldn't find a better way to do it. Suggestions are welcome.

crates/wysiwyg/src/dom/range.rs
```diff
@@ -72,6 +72,17 @@ impl DomLocation {
         }
     }
 
+    pub fn with_new_handle(&self, handle: DomHandle) -> Self {
```
jmartinesp, line 75: It's probably not a great name, but I couldn't come up with something better.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -225,9 +226,16 @@ where
...
     ) {
+        let mut moved_handles = HashMap::<DomHandle, DomHandle>::new();
         // Go through the locations in reverse order to prevent Dom modification issues
         for loc in locations.iter().rev() {
-            if Self::needs_format(&self.state.dom, loc, format) {
+            let mut loc = loc.clone();
+            let moved_parent_handle = moved_handles.get(&loc.node_handle);
+            if let Some(new_handle) = moved_parent_handle {
+                // Careful here, the location's position is no longer valid
+                loc = loc.with_new_handle(new_handle.clone());
```
jmartinesp, line 236: I tried to avoid modifying the location here and using DomHandle instead, but `split_text_node_by_offsets` relies on several DomLocation values and functions.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -384,17 +396,26 @@ where
         }
     }
 
-    fn merge_formatting_node_with_siblings(&mut self, handle: &DomHandle) {
+    fn merge_formatting_node_with_siblings(
+        &mut self,
+        handle: &DomHandle,
+    ) -> HashMap<DomHandle, DomHandle> {
```
andybalaam, line 402: A doc comment here explaining the return value would be nice.

crates/wysiwyg/src/composer_model/join_nodes.rs
andybalaam: We could pass in a mutable reference similar to how we are in join_format_nodes_at_level?

crates/wysiwyg/src/dom/range.rs
andybalaam: Seems fine to me.

crates/wysiwyg/src/tests/test_formatting.rs
```diff
@@ -115,3 +115,14 @@ fn formatting_twice_adds_no_formatting() {
     }
     assert_eq!(tx(&model), input);
 }
+
+#[test]
+fn formatting_nested_format_nodes_and_line_breaks() {
```
andybalaam, line 120: A few more tests would make me feel better - how about spanning multiple nodes, some of which contain BRs, and some of those BRs are not just at the top level?

andybalaam (approved):
Looks good - main comment is some ideas for more tests.

This feels like a step towards the right way, but not fully there yet...

crates/wysiwyg/src/composer_model/join_nodes.rs
jmartinesp: To me it makes more sense to just return the value and it's up to the user to actually use it or not rather than forcing us to pass an empty HashMap just in case we want to use them later. I'd use an `Option<&mut HashMap>` here, but we've already seen that passing mutable references inside Options is no easy feat 😅 .

PR #142, opened 2022-09-28
Allow for dynamic height text view and add support for dynamic font size

langleyd:
- Allow the Wysiwyg text view to grow with a dynamic height. 
- Add support for dynamic font sizes
- I moved the wrapping code(WywiwygView, WysiwygHostingView, etc) out of the library as I think there will be sufficient customisation there in the parent projects that it's more trouble than it's work to try generalise it. A bit simpler if the library just provides the textview with no insets etc and the parent projects can do as they wish.

Simulator Screen Recording - iPhone 13 - 2022-09-28 at 11 14 53


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -266,13 +267,20 @@ private extension WysiwygComposerViewModel {
     ///   - textView: The composer's text view.
     func updateIdealHeightIfNeeded(_ textView: UITextView) {
         // TODO: remove magic numbers
```
aringenbach, line 269:
```suggestion
```

aringenbach (approved):
One small comment inline, otherwise LGTM

Note: I took the liberty to add an iOS label to the PR, this automatically launches the iOS build & test which is skipped otherwise.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -211,8 +211,9 @@ private extension WysiwygComposerViewModel {
     ///   - end: End location for the selection.
     func applyReplaceAll(codeUnits: [UInt16], start: UInt32, end: UInt32) {
         do {
-            let html = String(utf16CodeUnits: codeUnits,
-                              count: codeUnits.count)
+            let htmlFragment = String(utf16CodeUnits: codeUnits,
+                                      count: codeUnits.count)
+            let html = generateHtmlBodyWithStyle(htmlFragment: htmlFragment)
```
aringenbach, line 216: I think this should only be applied at the UI level, as I expect this property to store the HTML we intend to send in the actual event formatted_body at some point.

langleyd: 😂 Yea that wasn't particularly thought through. Thanks

PR #143, opened 2022-09-28
feat(wysiwyg): Add Markdown support

Hywan:
This PR adds a new trait: `ToMardown` behind a new `to-markdown` Cargo feature. It allows to transform a DOM tree into Markdown.

In the test suite, by default, `assert_to_md` will roundtrip from HTML to Markdown, to HTML again. Otherwise, I've used `assert_to_md_no_roundtrip`. It happens when the Markdown to HTML compiler isn't able to parse our generated Markdown (it seems to be a bug, because it works with other parsers…) or because there is some formatting added (like newlines). This “roundtrip guarantee” isn't necessary for now. I was anticipating a next feature :-).

## Progress

- [x] line break node
- [x] text node
- [x] container node
  - [x] generic node
  - [x] formatting node
    - [x] bold
    - [x] italic
    - [x] strike through
    - [x] underline
    - [x] inline code 
  - [x] link node
  - [x] list node
  - [x] list item node

### DIFF
--- a/crates/wysiwyg/Cargo.toml
+++ b/crates/wysiwyg/Cargo.toml
@@ -19,7 +19,3 @@ widestring = "1.0.2"
 [dev-dependencies]
 speculoos = "0.9"
 pulldown-cmark = { version = "0.9.2", default-features = false }
-
-[features]
-default = ["to-markdown"]
-to-markdown = []
--- a/crates/wysiwyg/Cargo.toml
+++ b/crates/wysiwyg/Cargo.toml
@@ -10,8 +10,6 @@ name = "wysiwyg"
 version = "0.1.0"
 rust-version = "1.60"
 
-[features]
-
 [dependencies]
 html-escape = "0.2.11"
 html5ever = "0.25.2"
@@ -20,3 +18,7 @@ widestring = "1.0.2"
 
 [dev-dependencies]
 speculoos = "0.9"
+
+[features]
+default = ["to-markdown"]
+to-markdown = []
--- a/crates/wysiwyg/Cargo.toml
+++ b/crates/wysiwyg/Cargo.toml
@@ -19,6 +19,7 @@ widestring = "1.0.2"
 
 [dev-dependencies]
 speculoos = "0.9"
+pulldown-cmark = { version = "0.9.2", default-features = false }
 
 [features]
 default = ["to-markdown"]
--- a/crates/wysiwyg/Cargo.toml
+++ b/crates/wysiwyg/Cargo.toml
@@ -11,7 +11,6 @@ version = "0.1.0"
 rust-version = "1.60"
 
 [dependencies]
-bitflags = { version = "1.3.2", optional = true }
 html-escape = "0.2.11"
 html5ever = "0.25.2"
 once_cell = "1.13.0"
@@ -23,4 +22,4 @@ pulldown-cmark = { version = "0.9.2", default-features = false }
 
 [features]
 default = ["to-markdown"]
-to-markdown = ["dep:bitflags"]
+to-markdown = []
--- a/crates/wysiwyg/Cargo.toml
+++ b/crates/wysiwyg/Cargo.toml
@@ -11,6 +11,7 @@ version = "0.1.0"
 rust-version = "1.60"
 
 [dependencies]
+bitflags = { version = "1.3.2", optional = true }
 html-escape = "0.2.11"
 html5ever = "0.25.2"
 once_cell = "1.13.0"
@@ -21,4 +22,4 @@ speculoos = "0.9"
 
 [features]
 default = ["to-markdown"]
-to-markdown = []
+to-markdown = ["dep:bitflags"]
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -17,7 +17,7 @@ use crate::dom::dom_handle::DomHandle;
 use crate::dom::nodes::dom_node::DomNode;
 use crate::dom::to_html::ToHtml;
 #[cfg(feature = "to-markdown")]
-use crate::dom::to_markdown::{Error as MarkdownError, ToMarkdown};
+use crate::dom::to_markdown::{MarkdownError, MarkdownOptions, ToMarkdown};
 use crate::dom::to_raw_text::ToRawText;
 use crate::dom::to_tree::ToTree;
 use crate::dom::unicode_string::{UnicodeStrExt, UnicodeStringExt};
@@ -381,14 +381,18 @@ impl<S> ToMarkdown<S> for ContainerNode<S>
 where
     S: UnicodeString,
 {
-    fn fmt_markdown(&self, buf: &mut S) -> Result<(), MarkdownError<S>> {
+    fn fmt_markdown(
+        &self,
+        buffer: &mut S,
+        options: &MarkdownOptions,
+    ) -> Result<(), MarkdownError<S>> {
         use ContainerNodeKind::*;
         use InlineFormatType::*;
 
         match self.kind() {
             Generic => {
                 for child in self.children.iter() {
-                    child.fmt_markdown(buf)?;
+                    child.fmt_markdown(buffer, options)?;
                 }
             }
 
@@ -400,13 +404,13 @@ where
                 // `foo_bar_baz`. We reckon it's good to follow this
                 // trend to avoid unexpected behaviours for our users.
 
-                buf.push("*");
+                buffer.push("*");
 
                 for child in self.children.iter() {
-                    child.fmt_markdown(buf)?;
+                    child.fmt_markdown(buffer, options)?;
                 }
 
-                buf.push("*");
+                buffer.push("*");
             }
 
             // Strong emphasis.
@@ -421,13 +425,13 @@ where
                 // interpreted by various Markdown compilers out
                 // there. Instead, it will produce `*__…__*`.
 
-                buf.push("__");
+                buffer.push("__");
 
                 for child in self.children.iter() {
-                    child.fmt_markdown(buf)?;
+                    child.fmt_markdown(buffer, options)?;
                 }
 
-                buf.push("__");
+                buffer.push("__");
             }
 
             Formatting(StrikeThrough) => {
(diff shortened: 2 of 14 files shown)


crates/wysiwyg/src/dom/to_markdown.rs
```diff
@@ -0,0 +1,64 @@
...
+use super::UnicodeString;
+use bitflags::bitflags;
+use std::{error::Error, fmt};
+
+#[derive(Debug)]
+pub enum MarkdownError<S>
+where
+    S: UnicodeString,
+{
+    InvalidListItem(Option<<S::Str as ToOwned>::Owned>),
```
jplatte, line 24:
```suggestion
    InvalidListItem(Option<S>),
```

Hywan: True!

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -374,6 +381,265 @@ where
...
+
+                child.fmt_markdown(buffer, options)?;
+            }
+
+            Ok(())
+        }
+
+        let mut options = *options;
+
+        match self.kind() {
```
andybalaam, line 421: This got long. What do you think about breaking the match arms into separate functions?

crates/wysiwyg/Cargo.toml
```diff
@@ -10,13 +10,17 @@ name = "wysiwyg"
...
 once_cell = "1.13.0"
 widestring = "1.0.2"
 
 [dev-dependencies]
 speculoos = "0.9"
+pulldown-cmark = { version = "0.9.2", default-features = false }
+
+[features]
+default = ["to-markdown"]
+to-markdown = ["dep:bitflags"]
```
andybalaam, line 26: I'm not sure I see the benefit of breaking out a to-markdown feature - we are always going to need it to send Matrix messages.

crates/wysiwyg/src/dom/to_markdown.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    fn to_markdown(&self) -> Result<S, MarkdownError<S>> {
+        let mut buffer = S::default();
+        self.fmt_markdown(&mut buffer, &MarkdownOptions::empty())?;
+
+        Ok(buffer)
+    }
+}
+
+bitflags! {
```
andybalaam, line 60: Is it worth this dependency just to hold some (currently 1!) boolean flag?

crates/wysiwyg/src/tests/test_to_markdown.rs
```diff
@@ -0,0 +1,228 @@
...
+    assert_to_md(
+        "abc<br /><br />def",
+        r#"abc\
+\
+def"#,
+    );
+}
+
+#[test]
+fn text_with_italic() {
```
andybalaam, line 46: Great tests!

crates/wysiwyg/src/tests/test_to_markdown.rs
```diff
@@ -0,0 +1,228 @@
...
+line2\
+\
+line3* def"#,
+    );
+}
+
+#[test]
+fn text_with_bold() {
+    assert_to_md("<strong>abc</strong>", "__abc__");
+    assert_to_md_no_roundtrip("abc<strong>def</strong>ghi", "abc__def__ghi");
```
andybalaam, line 64: Assuming this markdown is valid, we could consider fixing this as a bug in our markdown library (or choosing another that works better?)

crates/wysiwyg/src/tests/test_to_markdown.rs
```diff
@@ -0,0 +1,228 @@
...
+    assert_to_md("abc <code>def</code> ghi", "abc `` def `` ghi");
+    assert_to_md("abc<code> def </code>ghi", "abc``  def  ``ghi");
+    // It's impossible to get a line break inside an inline code with Markdown.
+    assert_to_md_no_roundtrip(
+        "abc <code>line1<br />line2<br /><br />line3</code> def",
+        "abc `` line1 line2  line3 `` def",
+    );
+    assert_to_md_no_roundtrip(
+        "abc <code>def <strong>ghi</strong> jkl</code> mno",
+        "abc `` def __ghi__ jkl `` mno",
```
andybalaam, line 126: Is this valid markdown? Maybe we should strip formatting inside code blocks?

crates/wysiwyg/src/tests/test_to_markdown.rs
```diff
@@ -0,0 +1,228 @@
...
+}
+
+#[test]
+fn list_ordered() {
+    assert_to_md(
+        r#"<ol><li>item1</li><li>item2</li></ol>"#,
+        r#"1. item1
+2. item2"#,
+    );
+}
```
andybalaam, line 168: I'd like to see some tests with multiple nested ordered and unordered lists, with formatting and code blocks inside them.

crates/wysiwyg/src/tests/test_to_markdown.rs
```diff
@@ -0,0 +1,228 @@
...
+    use md_parser::{html, Options, Parser};
+
+    let mut options = Options::empty();
+    options.insert(Options::ENABLE_STRIKETHROUGH);
+
+    let markdown = markdown.as_ustr().to_string_lossy();
+
+    let parser = Parser::new_ext(&markdown, options);
+
+    let mut html = String::new();
```
andybalaam, line 201: This confused me having a variable called `html` as well as `md_parser::html` so close together. Can we think of a different name for the variable?

andybalaam (changes requested):
This is great, thank you. Some questions, minor comments and one new test.

crates/wysiwyg/src/dom/nodes/container_node.rs
Hywan: Functions will be super small, and the `match` will stay. What's long is the comments, not the code. If you want me to change, that's fine, but I'm not sure it gonna make things any simpler: more code (due to function declaration), and more jump in the code.

crates/wysiwyg/Cargo.toml
Hywan: I thought a feature was a good idea for the moment since it's still experimental. But I'm fine removing it. Tell me what you would prefer :-).

crates/wysiwyg/src/dom/to_markdown.rs
Hywan: Haha, probably not! Previously I got 3 options, but we can remove `bitflags` for sure ;-).

crates/wysiwyg/src/tests/test_to_markdown.rs
Hywan: To clarify, the problem doesn't come from “our” code, but from `pulldown` I guess. The problem is that Markdown is not very clear about what to do here. I'll see with `pulldown` why they did this as such.

crates/wysiwyg/src/tests/test_to_markdown.rs
Hywan: So… yeah, having 2 backticks + 1 space as a surroundding pair is valid! See [the specification](https://github.github.com/gfm/#example-339). Using 2 backticks allows to have one backtick inside the code. It's the most weird-syntax-proof syntax I can come with. Example, to write  `` `foo _bar_ baz` `` I need to write:

```md
`` `foo _bar_ baz` ``
```

But why can't ensure a roundtrip here?

The HTML code is valid. The resulting Markdown is valid. But the specification is ambiguous here.

The following HTML code `<code>foo <em>bar</em> baz</code>` must map to `` `foo _bar_ baz` `` in Markdown. That's correct. However, `` `foo _bar_ baz` `` doesn't necessarily map to `<code>foo <em>bar</em> baz</code>` back to HTML. Actually, let's see what the Github compiler does: “`foo _bar_ baz`”. It renders this `foo \_bar\_ baz` (without the backslash). The `_` isn't compiled to `<em>`. Annoying right? That's because HTML allows inline element inside a `<code>` element, but Markdown doesn't.

crates/wysiwyg/src/tests/test_to_markdown.rs
Hywan: OK!

crates/wysiwyg/src/tests/test_to_markdown.rs
Hywan: Sure.

crates/wysiwyg/src/dom/nodes/container_node.rs
andybalaam: Yeah, I do want it broken up please. Long functions, even if it's comments making them long, make it hard to follow the flow.

crates/wysiwyg/Cargo.toml
andybalaam: Yeah, remove it please.

crates/wysiwyg/src/dom/to_markdown.rs
andybalaam: Great, please do.

crates/wysiwyg/src/tests/test_to_markdown.rs
andybalaam: Yeah, just musing here that we could consider reporting/fixing a bug in pulldown.

crates/wysiwyg/src/tests/test_to_markdown.rs
andybalaam: Yeah, which is fine: we know that markdown does not support all of HTML. So arguably the right behaviour is to render

```
abc <code>def <strong>ghi</strong> jkl</code> mno
```

as

```
abc `` def ghi jkl `` mno
```

but it's a corner case and not something we need to do right now.

crates/wysiwyg/src/dom/to_markdown.rs
Hywan: Done.

crates/wysiwyg/src/tests/test_to_markdown.rs
Hywan: Done.

crates/wysiwyg/src/tests/test_to_markdown.rs
Hywan: Done.

crates/wysiwyg/src/dom/nodes/container_node.rs
Hywan: Done, https://github.com/matrix-org/matrix-wysiwyg/pull/143/commits/568250f784fa180fbc0343b3e40e821349fce322.

Hywan:
All feedbacks have been addressed :-).

andybalaam (approved):

andybalaam:
Awesome, thank you!

PR #144, opened 2022-09-28
Toggle format with no selection

aringenbach:
* Add the possibility to toggle formatting with no selection without updating the current HTML, but saving a state instead.
* The state is cleared on selection changes.
* The state is restored if needed when using `undo`.
* Status for highlighted buttons in client is created regarding both this new state and the tags at the current position.

Simulator Screen Recording - iPhone 13 - 2022-09-28 at 15 01 37

### DIFF
--- a/crates/wysiwyg/src/composer_model/format.rs
+++ b/crates/wysiwyg/src/composer_model/format.rs
@@ -74,13 +74,17 @@ where
     }
 
     pub(crate) fn apply_pending_formats(&mut self, start: usize, end: usize) {
-        self.state.toggled_format_types.clone().iter().for_each(|format| {
-            if self.reversed_actions.contains(&format.action()) {
-                self.format_range(start, end, format.clone());
-            } else {
-                self.unformat_range(start, end, format.clone());
-            }
-        });
+        self.state
+            .toggled_format_types
+            .clone()
+            .iter()
+            .for_each(|format| {
+                if self.reversed_actions.contains(&format.action()) {
+                    self.format_range(start, end, format);
+                } else {
+                    self.unformat_range(start, end, format);
+                }
+            });
         self.state.toggled_format_types.clear();
     }
 
@@ -90,10 +94,10 @@ where
         let (s, e) = self.safe_selection();
 
         if s == e {
-            self.toggle_zero_length_format(format);
+            self.toggle_zero_length_format(&format);
             ComposerUpdate::update_menu_state(self.compute_menu_state())
         } else {
-            self.format_range(s, e, format);
+            self.format_range(s, e, &format);
             self.create_update_replace_all()
         }
     }
@@ -102,7 +106,7 @@ where
         &mut self,
         start: usize,
         end: usize,
-        format: InlineFormatType,
+        format: &InlineFormatType,
     ) {
         assert!(start != end);
         let range = self.state.dom.find_range(start, end);
@@ -113,10 +117,10 @@ where
         let (s, e) = self.safe_selection();
 
         if s == e {
-            self.toggle_zero_length_format(format);
+            self.toggle_zero_length_format(&format);
             ComposerUpdate::update_menu_state(self.compute_menu_state())
         } else {
-            self.unformat_range(s, e, format);
+            self.unformat_range(s, e, &format);
             self.create_update_replace_all()
         }
     }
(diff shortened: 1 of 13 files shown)


crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -73,31 +73,77 @@ where
         }
     }
 
-    pub fn format(&mut self, format: InlineFormatType) -> ComposerUpdate<S> {
+    pub(crate) fn apply_pending_formats(&mut self, start: usize, end: usize) {
+        self.state
+            .toggled_format_types
+            .clone()
```
andybalaam, line 79: You should be able to avoid the clone here, if format_range and unformat_range took references to a format

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -73,31 +73,77 @@ where
...
+        let range = self.state.dom.find_range(start, end);
+        self.unformat_several_nodes(start, end, &range, format);
+    }
+
+    fn toggle_zero_length_format(&mut self, format: &InlineFormatType) {
+        if self.state.toggled_format_types.contains(&format) {
+            self.state
+                .toggled_format_types
+                .iter()
+                .position(|f| f == format);
```
andybalaam, line 143: @Hywan I hadn't seen this before - is this a normal way to search and modify an element in a Vec?

crates/wysiwyg/src/tests/test_formatting.rs
```diff
@@ -115,3 +117,55 @@ fn formatting_twice_adds_no_formatting() {
...
+#[test]
+fn selecting_removes_toggled_format_types() {
+    let mut model = cm("aaa|");
+    model.bold();
+    assert_eq!(
+        model.state.toggled_format_types,
+        Vec::from([InlineFormatType::Bold]),
+    );
+    model.select(Location::from(2), Location::from(2));
+    assert_eq!(model.state.toggled_format_types, Vec::new(),);
```
andybalaam, line 170: Worth asserting here that when we type the formatting is not applied.

andybalaam (changes requested):
One tiny change and a question I asked of Ivan, but looks great!

crates/wysiwyg/src/composer_model/format.rs
aringenbach: Actually this is missing some code 🙈  

Should be something like that 

```rust
            let index = self.state
                .toggled_format_types
                .iter()
                .position(|f| f == format);
            if let Some(index) = index {
                self.state.toggled_format_types.remove(index);
            }
```

But perhaps there is a better alternative

aringenbach: (which means we can remove the `if contains` in that case) 

/me is making test cases for multiple toggling cause it definitely doesn't work 😅

crates/wysiwyg/src/composer_model/format.rs
aringenbach: Yeah actually we want to clear anyway so we can just remove first until it's empty, no need to iterate

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -74,18 +74,14 @@ where
...
-            .iter()
-            .for_each(|format| {
-                if self.reversed_actions.contains(&format.action()) {
-                    self.format_range(start, end, format);
-                } else {
-                    self.unformat_range(start, end, format);
-                }
-            });
-        self.state.toggled_format_types.clear();
+        while self.state.toggled_format_types.len() > 0 {
```
andybalaam, line 78: you should be able to replace the` while` and the line immediately after with:

```rust
for format in self.state.toggled_format_types.into_iter() {
```

andybalaam (changes requested):
One tiny change, but everything else looks great.

crates/wysiwyg/src/composer_model/format.rs
aringenbach: This adds some borrowing issue for me, and I don't see how it will remove from the array at the same time ? 
I changed a bit, with a reverse + pop instead of removing 0 potentially multiple times.

andybalaam (approved):
Great!

PR #146, opened 2022-09-28
Fix reversed_actions state when selecting a line break or right after it

aringenbach:
Fixes issues with the menu state when selecting line breaks or having the cursor next to a line break. 
Line breaks now behave like text nodes and lookup formatting nodes in their ancestors.


jmartinesp (approved):
LGTM, although the logic is a bit harder to follow now.

aringenbach:
@jmartinesp Added a couple comments and a minor fix.

jmartinesp (approved):
LGTM.

PR #147, opened 2022-09-29
Create DomActionList struct to represent future actions to be performed to the Dom

jmartinesp:


crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -0,0 +1,316 @@
...
+            .iter()
+            .filter_map(|a| match a {
+                DomAction::Move(a) => Some(a),
+                _ => None,
+            })
+            .collect()
+    }
+
+    /// To avoid unnecessary copies, this function returns the [actions] and marks the DomActionList
+    /// as moved.
```
andybalaam, line 120: "and marks the DomActionList as moved" => "and consumes the DomActionList"

crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+        (to_add, to_remove, to_move)
+    }
+
+    pub fn push(&mut self, action: DomAction<S>) {
+        self.actions.push(action);
+    }
+
+    #[allow(dead_code)]
+    pub fn replace_actions(&mut self, new_actions: Vec<DomAction<S>>) {
```
andybalaam, line 147: Why do we need this? Can't we just create a new DomActionList?

crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -0,0 +1,316 @@
...
+        self.actions = new_actions;
+    }
+
+    pub fn find_moved_parent_or_self(
+        &self,
+        handle: &DomHandle,
+    ) -> Option<(DomHandle, DomHandle)> {
+        for action in &self.actions {
```
andybalaam, line 155: I think self.actions.iter().find_map might be nicer here?

crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -0,0 +1,316 @@
...
+        assert_eq!(list.actions().get(2).unwrap(), actions.get(2).unwrap());
+    }
+
+    #[test]
+    fn additions_returns_only_add_actions() {
+        let actions = default_actions();
+        let list = DAL::new(actions.clone());
+        assert_eq!(1, list.additions().len());
+        assert_eq!(
+            DomAction::Add(list.additions().get(0).unwrap().clone().clone()),
```
andybalaam, line 238: Double clone?

andybalaam: Also in a couple of other places.

crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn new_can_create_empty_list_too() {
+        let list = DAL::new(Vec::new());
+        assert!(list.actions.is_empty());
+    }
+
+    #[test]
+    fn actions_fn_returns_all_actions() {
+        let actions = default_actions();
+        let list = DAL::new(actions.clone());
+        assert_eq!(list.actions().get(0).unwrap(), actions.get(0).unwrap());
```
andybalaam, line 227: All these `.get(0).unwrap()` can be replaced by `[0]` I think.

crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -0,0 +1,316 @@
...
+    #[test]
+    fn find_moved_parent_or_self_looks_for_it_in_move_actions() {
+        let actions = vec![DomAction::move_node(
+            DomHandle::from_raw(vec![1, 0]),
+            DomHandle::from_raw(vec![0, 0]),
+        )];
+        let list = DAL::new(actions);
+        let found_parent =
+            list.find_moved_parent_or_self(&DomHandle::from_raw(vec![1, 0, 0]));
+        assert!(found_parent.is_some());
```
andybalaam, line 293: Maybe worth adding one assertion that it actually returns what we expect?

andybalaam (changes requested):
Minor changes, looks great.

crates/wysiwyg/src/dom/action_list.rs
jmartinesp: If you're passing a reference to the current `DomActionList` you might need to replace its contents while keeping the same reference.

crates/wysiwyg/src/dom/action_list.rs
jmartinesp: I think it's a double reference. Should I use `deref()` instead? I'm still quite inexperienced in this part of memory management in Rust 😅 .

jplatte: Should actually sort itself out if you replace `get(0).unwrap()` with `[0]`.

andybalaam (approved):
Looks great, thanks!

PR #148, opened 2022-09-29
fix(wysiwyg): Remove useless a `to_string` call + code simplification with `matches!`

Hywan:
This PR is twofold:

1. It removes a useless call to `to_string` (hence removing one memory allocation),
2. It simplify code with `matches!`.

Explanations for `matches!`:

```rust
match self {
    DomNode::Container(n) => n.is_structure_node(),
    _ => false,
}
```

is equivalent to:

```rust
match self {
    DomNode::Container(n) if n.is_structure_node() => true,
    _ => false,
}
```

which then can be used with `matches!` as:

```rust
matches!(DomNode::Container(n) if n.is_structure_node())
```

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/dom_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/dom_node.rs
@@ -116,24 +116,15 @@ where
     }
 
     pub fn is_structure_node(&self) -> bool {
-        match self {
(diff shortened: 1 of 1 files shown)


aringenbach (approved):
LGTM

jplatte (approved):

PR #149, opened 2022-09-29
Safari: fix formatting buttons and shortcuts

florianduros:
Safari "loses" the `inputType` when a `input` event is manually dispatched.

- I used the custom event `formatBlock` instead
- I move the `assert` file which is only used by the events in the `useListeners` folder


platforms/web/lib/useFormattingActions.ts
```diff
@@ -17,20 +17,15 @@ limitations under the License.
...
-        // For this case, we send a custom event and we do not use the browser input event handling
-        const sendFormatBlockEvent = (e: ReactMouseEvent<HTMLElement, MouseEvent>, blockType: BlockType) => {
-            e.preventDefault();
-            e.stopPropagation();
-            editorRef.current?.dispatchEvent(new CustomEvent('formatBlock', { detail: { blockType } }));
-        };
+        // Safari does not keep the inputType in an input event when the input event is fired manually
+        // So we send a custom event and we do not use the browser input event handling
+        const sendEvent = (e: ReactMouseEvent<HTMLElement, MouseEvent> | KeyboardEvent,
+            blockType: BlockType) => editorRef.current && sendFormatBlockEvent(e, editorRef.current, blockType);
```
andybalaam, line 28: sendFormatBlockEvent should be renamed now, right?

andybalaam (approved):
One question - looks great otherwise.

platforms/web/lib/useFormattingActions.ts
florianduros: It sends the `formatBlock` event, for me the name is correct.

Rename the `formatBlock` event is more relevant in my opinion. I don't have a more suitable name in my mind ^^

andybalaam: Doesn't it send all types of event now, not just formatBlock?

If formatBlock now means all events, that should be renamed to `ourInput` or similar I guess.

PR #150, opened 2022-09-29
Fix unformatting multiple nodes at the same time

aringenbach:
Fixes issues (panics) when trying to unformat multiple consecutive node of the same type with nested node or line breaks inside

### DIFF
--- a/crates/wysiwyg/src/composer_model/format.rs
+++ b/crates/wysiwyg/src/composer_model/format.rs
@@ -206,45 +206,58 @@ where
         range: &Range,
         format: &InlineFormatType,
     ) {
-        for location in range.locations.iter() {
-            let node = self.state.dom.lookup_node(&location.node_handle);
-            if let DomNode::Container(n) = node {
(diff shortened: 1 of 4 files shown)


crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -206,46 +206,59 @@ where
...
+
+        // Find slices of text before and after the selection that will require re-format.
+        let mut reformat_to: Option<usize> = None;
+        let mut reformat_from: Option<usize> = None;
+        if let Some(location) = formatting_locations.first() {
+            // Actual last node, find text to reformat after.
+            let (after_start, after_end) = self.safe_locations_from(
+                Location::from(end),
+                Location::from(end + location.length - location.end_offset),
+            );
```
aringenbach, line 228: Wonder if we really need the `safe_locations_from` here, as if the model and the find_range work properly we'll never have invalid locations (perhaps it's better to crash than having weird behaviour if it ever happens)

andybalaam (approved):
Looks good, thanks!

PR #151, opened 2022-09-30
Display link href inside debug tree representation

aringenbach:
Allows displaying the href from link in the tree debug representation


jmartinesp (approved):
LGTM!

PR #152, opened 2022-09-30
[iOS] Add a log when the ViewModel tries to apply some action on the current selection

aringenbach:
Adds an iOS log when the ViewModel tries to apply an action.


jmartinesp (approved):
LGTM

PR #154, opened 2022-10-03
[Android] Setup maven for local distribution, fix integration issues

jmartinesp:


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -19,11 +19,11 @@ import kotlin.math.min
...
 class EditorEditText : TextInputEditText {
 
-    lateinit var inputConnection: InterceptInputConnection
+    private var inputConnection: InterceptInputConnection? = null
     private val inputProcessor = InputProcessor(
         context,
-        composer = newComposerModel(),
-        menuStateCallback = { menuStateChangedListener?.menuStateChanged(it) }
+        menuStateCallback = { menuStateChangedListener?.menuStateChanged(it) },
+        composer = if (isInEditMode) null else newComposerModel()
```
jmartinesp, line 26: This allows us to use the component in XMLs without layout preview crashing.

andybalaam: Worth a comment in the code?

andybalaam (approved):
Looks good, one optional comment.

PR #159, opened 2022-10-04
Handle replacing text with newline characters

aringenbach:
Automatically replaces newline characters in replace_text with `<br />` or `<li>` depending on context. 

Simulator Screen Recording - iPhone 13 - 2022-10-04 at 14 47 54

Simulator Screen Recording - iPhone 13 - 2022-10-04 at 15 03 25

(second screen record contains an iOS selection issue around lists that is definitely not related to this change)


andybalaam (approved):
Great! Encouraged how relatively simple this was...

PR #163, opened 2022-10-05
Add br tag at hook init

florianduros:


andybalaam (approved):
Looks good. I don't think this guarantees that we always add a BR after the content given to us by ComposerModel, but we should do that. Is that elsewhere, or is it missing at the moment?

PR #165, opened 2022-10-05
Escape html in the DOM view

andybalaam:


platforms/web/lib/dom.ts
```diff
@@ -68,8 +68,8 @@ export function refreshComposerView(
         while ((child = children.next())) {
             const nodeType: string = child.node_type(composerModel);
             if (nodeType === 'container') {
-                // TODO I'm a bit septic about this id :p
-                let id = idCounter;
+                // TODO: I'm a bit sceptical about this id :p (Florian)
```
florianduros, line 71: I think you can remove this comment, without the `id++`. I understand the purpose of the `id` :)

florianduros (approved):

platforms/web/lib/dom.ts
andybalaam: No idea what I was doing there - must be the cruft of something else.

PR #166, opened 2022-10-05
[Android] Handle cut & paste events

aringenbach:
Some code that handles cut & paste on Android. Works for text with or without newline characters


jmartinesp:
The code seems to be working fine. @aringenbach is there anything else you wanted to add or can we review and merge this now?

PR #171, opened 2022-10-05
[Android] Setup CI to run UI tests

jmartinesp:


platforms/android/library/src/androidTest/res/layout/activity_test.xml
```diff
@@ -18,7 +18,8 @@
         <io.element.android.wysiwyg.EditorEditText
             android:id="@id/rich_text_edit_text"
             android:layout_width="match_parent"
-            android:layout_height="wrap_content" />
+            android:layout_height="wrap_content"
+            android:inputType="text|textNoSuggestions" />
```
jmartinesp, line 22: This fixes the `EditorEditTextInputTests.testHardwareKeyboardTyping` test: it seems like the issue was the IME's autocorrect feature getting in the way.

.github/workflows/android.yml
```diff
@@ -0,0 +1,88 @@
...
+          api-level: ${{ matrix.api-level }}
+#          ndk: 25.1.8937393
+          arch: x86
+          profile: Nexus 5X
+#          target: playstore
+          force-avd-creation: false
+          emulator-options: -no-snapshot-save -no-window -gpu swiftshader_indirect -noaudio -no-boot-anim -camera-back none
+          disable-animations: true
+          script: |
+            ./gradlew :library:testDebugUnitTest :library:connectedDebugAndroidTest ${{ env.CI_GRADLE_ARG_PROPERTIES }}
```
jmartinesp, line 79: I split the build+run step into two separate steps in case we want to separate the actual tests from the build in the future. This way, we can only retry the actual tests without needing to rebuild (i.e. for flaky tests). 

Merging them back might save some build time though, it might be worth exploring.

andybalaam (approved):
Looks good, thanks.

PR #173, opened 2022-10-06
Customisable Text Color To Support Dark Theme

Velin92:
Added a `textColor` variable inside the WysiwygComposerView that can be used to set programmatically at each update the color of the text.

By default the `textColor` variable is equal to `label` colour which is already dark theme responsive (black for light theme, white for dark theme).

The `textColor ` can either be customised in the init or using a view modifier in SwiftUI style.

Also fixed a concurrency bug that slowed down the loading of the view.

https://element-io.atlassian.net/jira/software/c/projects/PSU/boards/29?modal=detail&selectedIssue=PSU-848

EDIT: Removed the value from the init, now the `textColor` can only be changed through SwiftUI-like modifier function

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
@@ -104,7 +104,7 @@ public struct WysiwygComposerView: UIViewRepresentable {
 }
 
 public extension WysiwygComposerView {
-    /// Sets the textColor of the WYSIWYG textView, if not used the default label UIColor is used.
+    /// Sets the textColor of the WYSIWYG textView, if not used the default value is Color.primary.
     func textColor(_ textColor: Color) -> Self {
         var newSelf = Self(
             content: content,
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
@@ -25,15 +25,18 @@ public struct WysiwygComposerView: UIViewRepresentable {
     public var replaceText: (NSAttributedString, NSRange, String) -> Bool
     public var select: (NSAttributedString, NSRange) -> Void
     public var didUpdateText: (UITextView) -> Void
+    public var textColor: UIColor
 
     public init(content: WysiwygComposerContent,
                 replaceText: @escaping (NSAttributedString, NSRange, String) -> Bool,
                 select: @escaping (NSAttributedString, NSRange) -> Void,
-                didUpdateText: @escaping (UITextView) -> Void) {
+                didUpdateText: @escaping (UITextView) -> Void,
+                textColor: UIColor = .label) {
         self.content = content
         self.replaceText = replaceText
         self.select = select
         self.didUpdateText = didUpdateText
+        self.textColor = textColor
     }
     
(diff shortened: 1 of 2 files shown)


Velin92:
@langleyd Hey David I think I don't have yet permission to add labels to the PR, could you add it for me?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
```diff
@@ -25,15 +25,18 @@ public struct WysiwygComposerView: UIViewRepresentable {
...
     public var select: (NSAttributedString, NSRange) -> Void
     public var didUpdateText: (UITextView) -> Void
+    public var textColor: UIColor
 
     public init(content: WysiwygComposerContent,
                 replaceText: @escaping (NSAttributedString, NSRange, String) -> Bool,
                 select: @escaping (NSAttributedString, NSRange) -> Void,
-                didUpdateText: @escaping (UITextView) -> Void) {
+                didUpdateText: @escaping (UITextView) -> Void,
+                textColor: UIColor = .label) {
```
langleyd, line 34: Probably makes sense to order this param before the closures so that we can use the trailing closure syntax?

langleyd, line 34: Should we change this to a SwiftUI `Color` instead of a `UIColor`so that we are not exposing UIKit? Can then convert to a UIKit color below in the `makeUIView`?

langleyd (approved):
Just  a couple of comments, otherwise LGTM

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
Velin92: Absolutely, actually I was wondering if we even want the parameter in init, we could also just leave it as a public var and use the modifier I created for the view, to set it. What do you think?

langleyd: Works for me

Velin92: done

Velin92: done!

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
```diff
@@ -25,6 +25,7 @@ public struct WysiwygComposerView: UIViewRepresentable {
     public var replaceText: (NSAttributedString, NSRange, String) -> Bool
     public var select: (NSAttributedString, NSRange) -> Void
     public var didUpdateText: (UITextView) -> Void
+    private var textColor: UIColor = .label
```
langleyd, line 28:
```suggestion
    private var textColor = Color(.label)
```

Should probably also make the property a `Color` I think properties using in swiftui should be value types or you can get weirdness.

Velin92: Yeah I agree, I can even use Color.primary since is the SwiftUI equivalent of UIColor.label

PR #174, opened 2022-10-06
[iOS] Fix text view height computation

aringenbach:
Fixes the text view sometimes outputting wrong required height (most notably around trailing line breaks). This was due to the computation not taking into account the changes from reconciliating the content with the Rust model.


langleyd (approved):
Sorry didn't see your message int he room about this earlier. Looks great!

PR #176, opened 2022-10-06
[iOS] textColor modifier code improvement

Velin92:
Just realised that by using the init inside the modifier function, that would reset the state of the view on the default values, invalidating any previously used modifier on the view (even custom ones we might build and use in the future). 
The best approach is just to copy self, as it is since is a struct, and then change the value of the textColor.


aringenbach (approved):
LGTM

PR #177, opened 2022-10-07
Fix `replace_all_html` not moving the selection

aringenbach:
* Fixes the `replace_all_html` so it moves the selection to the end of the content.
* Enable and use the `clear` variant in iOS lib.
* Add a public API in the iOS lib to allow the hosting application to set the HTML content.


andybalaam (approved):
Looks great!

PR #178, opened 2022-10-07
Hack the generated code to avoid the use of fetch

andybalaam:
Using fetch() to load WASM from a data: URL can be blocked by a Content Security Policy, so we replace the use of fetch with directly extracting the data from the URL and returning a Response object containing it.


platforms/web/scripts/hack_myfetch.js
```diff
@@ -0,0 +1,108 @@
...
+    return myfetch($1)
+}())`;
+
+// Note: we do very simple search+replace here, so if the code contains other
+// fetch calls, they will be messed up.
+
+replace({
+  regex: /fetch\((\w+)\)/,
+  replacement: code.replace(/\n\s*/g, ''),
+  paths: ['./dist/matrix-wysiwyg.umd.cjs'],
```
florianduros, line 105: Two comments here!

- Should we also patch the `dist/wysisyg.js` ?
- In the `react-sdk`, if I'm not wrong, we are using the `dist/wysisyg.js`

florianduros (changes requested):

platforms/web/scripts/hack_myfetch.js
andybalaam: I found that the one I needed to change was matrix-wysiwyg.umd.cjs but it may actually be different for different setups.

I added processing the other one.

florianduros (approved):

PR #192, opened 2022-10-11
Cut and paste

andybalaam:


platforms/web/lib/useListeners/useListeners.ts
```diff
@@ -50,6 +50,26 @@ export function useListeners(
             );
         editorNode.addEventListener('input', onInput);
 
+        const onPaste = (e: Event) => {
+            console.log(e);
```
florianduros, line 54: Remaining console.log

platforms/web/package.json
```diff
@@ -49,6 +51,7 @@
         "@vitejs/plugin-react": "^2.1.0",
         "@vitest/coverage-c8": "^0.23.4",
         "@vitest/ui": "^0.23.4",
+        "cypress": "^10.9.0",
```
florianduros, line 54: Great cypress test!

florianduros (changes requested):

platforms/web/lib/useListeners/useListeners.ts
andybalaam: Thanks, fixed in 7d878cf

florianduros (approved):

PR #200, opened 2022-10-13
Web: Add formatting state

florianduros:
Multiple things here:

- I removed the `onChange` handler and instead I return `content` in the `useWysiwyg` hook. It fits more the react _way_
- The hook returns `formattingStates` which is an object which contains the `formattingState` of all buttons
- I change a bit the typescript generation of the library to make work the complex type when the library is used. But it's still very febrile
- Remove remaining console.log
- Fix a call of `setInterval` instead of `setTimeout`

### DIFF
--- a/platforms/web/lib/composer.ts
+++ b/platforms/web/lib/composer.ts
@@ -25,7 +25,6 @@ export function processInput(
 ) {
     if (e instanceof ClipboardEvent) {
         const data = e.clipboardData?.getData('text/plain') ?? '';
-        console.log(data);
         return action(composerModel.replace_text(data), 'paste');
     }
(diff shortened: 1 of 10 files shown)


platforms/web/lib/composer.ts
```diff
@@ -25,7 +25,6 @@ export function processInput(
 ) {
     if (e instanceof ClipboardEvent) {
         const data = e.clipboardData?.getData('text/plain') ?? '';
-        console.log(data);
```
andybalaam, line 28: Sorry!

andybalaam (approved):
Looks great, thanks!

PR #202, opened 2022-10-13
Allow formatting in empty composer

jmartinesp:
Also prevent selecting the same cursor position from resetting menu state. 
Fixes for Android included too.


andybalaam (approved):
Nice one

PR #209, opened 2022-10-14
Fix replacing first line with line break

jmartinesp:
When calling `ComposerModel.enter()` in this situation, what happens internally is:
```
// Clear selection then enter.
self.do_replace_text_in("".into(), s, e);
self.do_enter() // This call is recursive
```

Which turns the tree into:
```
├>""
├>br
└>"BBB"
```
And when we then ask for nodes in range [0,0] both the first empty TextNode and the line break appear as leaf nodes in the index 0 of the Dom.

### DIFF
--- a/crates/wysiwyg/src/composer_model/replace_text.rs
+++ b/crates/wysiwyg/src/composer_model/replace_text.rs
@@ -16,7 +16,7 @@ use crate::dom::action_list::{DomAction, DomActionList};
 use crate::dom::nodes::DomNode;
 use crate::dom::unicode_string::{UnicodeStrExt, UnicodeStringExt};
 use crate::dom::{DomHandle, DomLocation, Range};
-use crate::{ComposerModel, ComposerUpdate, Location, ToTree, UnicodeString};
+use crate::{ComposerModel, ComposerUpdate, Location, UnicodeString};
 
 impl<S> ComposerModel<S>
 where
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -50,6 +50,8 @@ where
     fn do_enter(&mut self) -> ComposerUpdate<S> {
         let (s, e) = self.safe_selection();
 
+        let tree = self.state.dom.to_tree().to_string();
```
andybalaam, line 53: Debug code?

crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -91,7 +93,32 @@ where
...
+                        &leaf.node_handle,
+                        vec![DomNode::new_line_break()],
+                    );
+                } else {
+                    self.state.dom.remove(&leaf.node_handle);
+                }
+            }
+            self.state.start += 1;
+            self.state.end = self.state.start;
+            self.create_update_replace_all()
```
andybalaam, line 121: This function is getting long, so I'd prefer this to be a separate function.

But if you're rushing to get this out, I'm happy for you to make that change later.

andybalaam (approved):
A couple of small comments to fix now or in a future change.

crates/wysiwyg/src/composer_model/replace_text.rs
jmartinesp: 🙈

PR #210, opened 2022-10-14
[Android] Refactor to consolidate hardware/software key handling

jonnyandrew:


platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
```diff
@@ -109,13 +108,16 @@ class EditorEditTextInputTests {
...
 
     @Test
-    @FlakyTest(detail = "Sometimes the pressKey event doesn't reach the view.")
-    fun testHardwareKeyMovementNotIntercepted() {
+    fun testMovementIsIntercepted() {
         onView(withId(R.id.rich_text_edit_text))
             .perform(ImeActions.setComposingText("Test"))
             .perform(pressKey(KeyEvent.KEYCODE_DPAD_LEFT))
             .perform(pressKey(KeyEvent.KEYCODE_DPAD_LEFT))
             .check(matches(selectionIsAt(3)))
```
jmartinesp, line 116: I just realised there's probably an error in this test. When you type 'Test', the cursor is at index 4, so this will fail because tapping on the left arrow key twice will get you to index 2, not 3. So we either change  `selectionIsAt(3))` to 2 instead or we remove the repeated `pressKey` action (I'd say this is better). 

What's surprising is this test sometimes passed, so maybe it's still flaky and the emulator sometimes ignores one of the key presses.

jmartinesp (changes requested):
I added a couple of comments to discuss.

platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
jonnyandrew: I noticed the first left arrow key press never changes the cursor position. I think the way that we're setting the intitial text in the test is not giving focus to the text input.

Demonstrated here (this test will pass):
```
        onView(withId(R.id.rich_text_edit_text))
            .perform(ImeActions.setComposingText("Test"))
            .check(matches(doesNotHaveFocus()))
            .perform(pressKey(KeyEvent.KEYCODE_DPAD_LEFT))
            .check(matches(hasFocus()))
 ```
 
But when I changed the test to use `typeText("Test")` rather than `setComposingText("Test")`, it seems to work as expected.
```
        onView(withId(R.id.rich_text_edit_text))
            .perform(typeText("Test"))
            .check(matches(hasFocus()))
            .perform(pressKey(KeyEvent.KEYCODE_DPAD_LEFT))
            .perform(pressKey(KeyEvent.KEYCODE_DPAD_LEFT))
            .check(matches(selectionIsAt(2)))
```

Is there any reason we couldn't use `typeText()` in the tests?

Comment (jmartinesp) [reply platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt:116]: > I think the way that we're setting the intitial text in the test is not giving focus to the text input.

Ah, that might be the cause of the issue. However, the current implementation (2 left presses) consistently fails for me, so it's weird that it only works for you like that. I'm using emulators with APIs 32 and 33.

> Is there any reason we use the custom setComposingText() action rather than typeText() in the tests?

For this test, no, we could use `typeText` instead: `typeText` emulates hardware keyboard interaction, so that's why we're using the custom `setComposingText` action instead for most of the tests as we want to simulate IME input, but for this test, since we're actually verifying HW keyboard input, `typeText` is probably a better choice.

File: platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt [hunk 2]
```diff
@@ -66,39 +64,27 @@ class EditorEditText : TextInputEditText {
...
-                false
-            } else if (event.metaState != 0 && event.unicodeChar == 0) {
-                // Is a modifier key
-                false
-            } else {
-                inputConnection?.sendHardwareKeyboardInput(event)
-                true
-            }
+
+    override fun onKeyDown(keyCode: Int, event: KeyEvent): Boolean =
```
jonnyandrew, line 75: According to [the docs](https://developer.android.com/reference/android/widget/TextView#onKeyDown(int,%20android.view.KeyEvent)) this is not a reliable source of software key events.

jonnyandrew:
Closing because the input processor methods (`commitText`, `deleteSurroundingText` etc.) are no longer called when using the software keyboard; composition is broken; and `onKeyDown` is not a reliable source of software key events.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -66,39 +64,27 @@ class EditorEditText : TextInputEditText {
...
 
-    /**
-     * We wrap the internal [EditableInputConnection] as it's not public, but its internal behavior
-     * is probably needed to work properly with the EditText.
-     */
     override fun onCreateInputConnection(outAttrs: EditorInfo): InputConnection {
-        val baseInputConnection = requireNotNull(super.onCreateInputConnection(outAttrs))
         val inputConnection =
-            InterceptInputConnection(baseInputConnection, this, inputProcessor)
+            InterceptInputConnection(this, inputProcessor)
```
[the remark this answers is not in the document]
jmartinesp: I think `baseInputConnection` is the one that's actually handling completion and autocorrect under the hood. If we don't use it, it seems like we'd have to implement them ourselves.

PR #211, opened 2022-10-17
Fix inserting a new line before a new line at index 0 crashes

jmartinesp:
Solved 2 corner cases:
* Inserting a new line before a new line with no TextNode before it.
* Inserting text before a new line when there was no TextNode before it.


andybalaam (approved):
Brilliant, thanks!

PR #214, opened 2022-10-17
feat(wasm): Add a `dev-build` NPM script

Hywan:
This patch adds a new NPM script: `dev-build`, which will build a “debug“/“profiling” build, instead of an optimized, small, no debug symbol, build. This is helpful when trying to debug something.


andybalaam (approved):
Looks great, thanks!

PR #215, opened 2022-10-17
iOS: Use rsync in release script so that deleted files get deleted

langleyd:


release_ios.sh
```diff
@@ -20,13 +20,13 @@ fi
 make ios
 
 BUILD_DIR=build
-REPO_PATH="${BUILD_DIR}/matrix-wysiwyg"
+REPO_PATH="${BUILD_DIR}/matrix-wysiwyg-composer-swift"
```
Velin92, line 23: shouldn't this be the source repo?

langleyd: Is the name of the SPM repo that we clone in to.

Velin92 (approved):
LGTM

PR #216, opened 2022-10-17
feat(wasm): Call a `start` function when a module is instantiated

Hywan:
By using `wasm_bindgen(start)`, we ensure that the `start` function is called everytime a Wasm module is instantiated. It means that we are sure that `console_error_panic_hook::set_once` will be called… once, and only once per Wasm instance.


andybalaam (approved):
Fantastic! I was not aware of this, and it's much better - thank you.

PR #218, opened 2022-10-17
Subhead is the default font size

Velin92:
According to design, 15.0 which is equivalent to subhead should be the default font size, and not body (which is 17.0)


langleyd (approved):
LGTM

PR #219, opened 2022-10-17
Allow the user to modify input events before processing, and listen for Ctrl+Enter events

andybalaam:


platforms/web/lib/useWysiwyg.ts
```diff
@@ -77,20 +79,23 @@ export function useWysiwyg(wysiwygProps?: WysiwygProps) {
         composerModel,
     );
 
+    const formattingFunctions = useFormattingActions(ref);
```
florianduros, line 82: We should also rename the hook

platforms/web/lib/useFormattingActions.ts
```diff
@@ -16,33 +16,32 @@ limitations under the License.
...
-                underline: () => sendEvent('formatUnderline'),
-                undo: () => sendEvent('historyUndo'),
-                redo: () => sendEvent('historyRedo'),
-                orderedList: () => sendEvent('insertOrderedList'),
-                unorderedList: () => sendEvent('insertUnorderedList'),
-                inlineCode: () => sendEvent('formatInlineCode'),
-                clear: () => sendEvent('clear'),
-            };
-        }, [editorRef]);
+    const formattingActions: FormattingFunctions = useMemo(() => {
```
florianduros, line 23: My mistake here, we should use the `useMemo` generic type: 

`useMemo<FormattingFunctions>`

platforms/web/src/App.tsx
```diff
@@ -51,11 +52,34 @@ function Button({ onClick, imagePath, alt, state }: ButtonProps) {
...
+        e: WysiwygInputEvent,
+        wysiwyg: Wysiwyg,
+    ): WysiwygInputEvent | null => {
+        if (e instanceof ClipboardEvent) {
+            return e;
+        } else if (enterToSend && e.inputType === 'insertParagraph') {
+            console.log(`SENDING: ${wysiwyg.content()}`);
+            e.preventDefault();
+            e.stopPropagation();
+            wysiwyg.actions.clear();
```
florianduros, line 67: I wonder if your proposal with changing the `inputType` to customise the behaviour is better here

platforms/web/src/App.tsx
```diff
@@ -51,11 +52,34 @@ function Button({ onClick, imagePath, alt, state }: ButtonProps) {
...
+    };
+
+    const { ref, isWysiwygReady, formattingStates, wysiwyg, debug } =
         useWysiwyg({
             isAutoFocusEnabled: true,
+            inputEventProcessor,
         });
 
+    const onEnterToSendChanged = () => {
+        setEnterToSend(!enterToSend);
```
florianduros, line 80: When using a value from the state to create a new state, we should use directly the function parameter from `setState`

`setEnterToSend(_enterToSend => !_enterToSend)`

platforms/web/src/App.tsx
andybalaam: You mean add a new `inputType` that tells us to clear? I don't think there is an existing one that means that.

florianduros: We have a custom one, you can take a look in `processInput`

andybalaam: But you think we should create a new custom event to clear the composer, instead of calling `.clear()` here? Why is that better?

florianduros: To be honest, I don't know.

Better to keep your implementation, it's clearer and it works

florianduros (approved):

andybalaam:
@florianduros I added one evil commit for you to review when you get a chance.

florianduros:
> @florianduros I added one evil commit for you to review when you get a chance.

Oh, I was not expecting that ^^

It's fine until we have a clean type generation i think

PR #220, opened 2022-10-18
[Android] Refactor to consolidate hardware/software key handling

jonnyandrew:
## What?
Ensure that the logic triggered by hardware keys is shared as much as possible with the logic triggered by software keys.

_Note: although some logic is now shared, this is achieved by imitating the software keyboard's interactions with the `InputConnection`. Hence it is by no means a perfect solution and behaviour may still vary between input types._ 

[`PSU-907`](https://element-io.atlassian.net/browse/PSU-907)

## Why?
Handling hardware and software keys separately has a few risks:
- manual testing in an emulator using a computer keyboard may not reflect the true behaviour when using a touch screen input.
- the surface area for bugs is greater.

### DIFF
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/utils/ImeActions.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/utils/ImeActions.kt
@@ -5,7 +5,6 @@ import android.view.inputmethod.EditorInfo
 import androidx.test.espresso.UiController
 import androidx.test.espresso.ViewAction
 import androidx.test.espresso.matcher.ViewMatchers.isDisplayed
-import io.element.android.wysiwyg.EditorEditText
 import io.element.android.wysiwyg.InterceptInputConnection
 import org.hamcrest.Matcher
 
@@ -68,7 +67,7 @@ object Ime {
             if (view == null) return
             val editorInfo = EditorInfo()
             val inputConnection = view.onCreateInputConnection(editorInfo)
-            (inputConnection as? InterceptInputConnection)?.backspace()
+            (inputConnection as? InterceptInputConnection)?.onHardwareBackspaceKey()
         }
     }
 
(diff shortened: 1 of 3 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/InterceptInputConnection.kt
```diff
@@ -204,30 +165,41 @@ class InterceptInputConnection(
...
+        } else
+            "\n"
+
+        return commitText(newText, 1)
+    }
+
+    private fun onHardwareCharacterKey(newChars: String): Boolean {
+        // The current composition may not be up to date at this point so we do not attempt to
+        // append to the current composition with the hardware keyboard.
+        finishComposingText()
```
jonnyandrew, line 200: Please note

platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/utils/ImeActions.kt
```diff
@@ -126,9 +125,7 @@ object Ime {
             if (view == null) return
             val editorInfo = EditorInfo()
             val inputConnection = view.onCreateInputConnection(editorInfo) as? InterceptInputConnection ?: return
-            val end = inputConnection.getCurrentComposition().second
-            inputConnection.setComposingRegion(end, end)
-            inputConnection.commitText("\n", 1)
+            inputConnection.onHardwareEnterKey()
```
jmartinesp, line 128: `ImeAction`s were implemented as the software keyboard counterpart of the `pressKey` actions that use the hardware keyboard input, so calling a method  `onHardware___Key` here feels a bit weird, but since the implementations should match now it should be fine.

platforms/android/library/src/main/java/io/element/android/wysiwyg/InterceptInputConnection.kt
```diff
@@ -275,7 +247,7 @@ class InterceptInputConnection(
         return baseInputConnection.requestCursorUpdates(cursorUpdateMode)
     }
 
-    fun getCurrentComposition(): Pair<Int, Int> {
+    fun getCurrentCompositionOrSelection(): Pair<Int, Int> {
```
jmartinesp, line 250: I should have renamed this long ago 😅 . Thanks!

platforms/android/library/src/main/java/io/element/android/wysiwyg/InterceptInputConnection.kt
jmartinesp: Nice catch!

jmartinesp (approved):
LGTM!

PR #224, opened 2022-10-19
[Android] Fix state restoration for EditorEditText

jmartinesp:


jmartinesp:
This approach causes some issues: the internal ComposerModel gets re-created for each configuration change, and this becomes problematic quickly as not only do we have to restore the HTML it contained, but also the previous and next Dom states.

I'll open another PR with a new strategy, trying to use a VM to keep a reference to the ComposerModel instance.

PR #225, opened 2022-10-19
feat(wasm): Reduce the Wasm module size by 76%

Hywan:
With #214 and [Twiggy](https://github.com/rustwasm/twiggy), I've noticed that a large portion of the Wasm module comes from [`html5ever`](https://github.com/servo/html5ever/). This crate is used to parse HTML fragments and to build an `html5ever` specific DOM tree. This specific DOM tree is converted to a “`PaDom`” tree, which is then converted to a final `Dom` tree.

This PR adds a new Cargo feature to the `wysiwyg` crate: `js`. This feature instructs the library to **not** use `html5ever`, but to use a Web API to parse HTML fragment: `DOMParser`. Indeed, there is this neat function: [`DOMParser.parseFromString`](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser/parseFromString), which does exactly what we want! Instead of embedding a full HTML5 parser, inside a Wasm module, which will be executed inside an HTML browser, well… we can use the Web API provided by this same browser to do the same job.

Let's see how it goes so far. Before this patch, the Wasm module (compiled by `npm run build` from `bindings/wysiwyg-wasm/`) has a size of 522Kb. After this patch, the same Wasm module has a size of 123Kb.

~~This is still a draft PR because, at the time of writing, not all nodes are parsed: not because it's hard, but just because I wanted to test my approach :-).~~ 

To run the tests, use:

```sh
$  wasm-pack test --release --firefox --headless -- --no-default-features --features js -- parser::parse
```

I wanted to reuse the existing `parse::parse` tests, but `speculoos` is used, which depends on `num`, which depends on `rustc-serialize`, which doesn't compile to Wasm properly, sadly.

## Progress

- [x] Implement a new `js` feature on `wysiwyg`
- [x] Use the `DOMParser` Web API inside `dom::parser::parse`
- [x] Map to `Dom` for all nodes
- [x] Write tests
- [x] Update CI

### DIFF
--- a/crates/wysiwyg/src/dom/dom_struct.rs
+++ b/crates/wysiwyg/src/dom/dom_struct.rs
@@ -72,6 +72,15 @@ where
         self.document().children()
     }
 
+    #[cfg(feature = "js")]
+    pub(crate) fn take_children(self) -> Vec<DomNode<S>> {
+        if let DomNode::Container(container) = self.document {
+            container.take_children()
+        } else {
+            panic!("Document should always be a Container!")
+        }
+    }
+
     pub fn append_child(&mut self, child: DomNode<S>) -> DomHandle {
         self.document_mut().append_child(child)
     }
--- a/crates/wysiwyg/src/dom/parser.rs
+++ b/crates/wysiwyg/src/dom/parser.rs
@@ -37,21 +37,22 @@ mod panode_text;
 mod paqual_name;
 mod parse;
 
+// Group all re-exports for `feature = "sys"`.
 #[cfg(feature = "sys")]
-use padom::PaDom;
+mod sys {
+    use super::*;
+
+    pub(super) use padom::PaDom;
+    pub(super) use padom_creation_error::PaDomCreationError;
+    pub(super) use padom_creator::PaDomCreator;
+    pub(super) use padom_handle::PaDomHandle;
+    pub(super) use padom_node::PaDomNode;
+    pub(super) use panode_container::PaNodeContainer;
+    pub(super) use panode_text::PaNodeText;
+    pub(super) use paqual_name::paqual_name;
+}
+
 #[cfg(feature = "sys")]
-use padom_creation_error::PaDomCreationError;
-#[cfg(feature = "sys")]
-use padom_creator::PaDomCreator;
-#[cfg(feature = "sys")]
-use padom_handle::PaDomHandle;
-#[cfg(feature = "sys")]
-use padom_node::PaDomNode;
-#[cfg(feature = "sys")]
-use panode_container::PaNodeContainer;
-#[cfg(feature = "sys")]
-use panode_text::PaNodeText;
-#[cfg(feature = "sys")]
-use paqual_name::paqual_name;
+use sys::*;
 
 pub use parse::parse;
--- a/crates/wysiwyg/src/dom/parser.rs
+++ b/crates/wysiwyg/src/dom/parser.rs
@@ -19,22 +19,39 @@
 //! [super::Dom]. All instances of classes within this module are thrown away
 //! when parsing finishes.
 
+#[cfg(feature = "sys")]
 mod padom;
+#[cfg(feature = "sys")]
 mod padom_creation_error;
+#[cfg(feature = "sys")]
 mod padom_creator;
+#[cfg(feature = "sys")]
 mod padom_handle;
+#[cfg(feature = "sys")]
 mod padom_node;
+#[cfg(feature = "sys")]
 mod panode_container;
+#[cfg(feature = "sys")]
 mod panode_text;
+#[cfg(feature = "sys")]
 mod paqual_name;
 mod parse;
 
+#[cfg(feature = "sys")]
 use padom::PaDom;
+#[cfg(feature = "sys")]
 use padom_creation_error::PaDomCreationError;
+#[cfg(feature = "sys")]
 use padom_creator::PaDomCreator;
+#[cfg(feature = "sys")]
 use padom_handle::PaDomHandle;
+#[cfg(feature = "sys")]
 use padom_node::PaDomNode;
+#[cfg(feature = "sys")]
 use panode_container::PaNodeContainer;
+#[cfg(feature = "sys")]
 use panode_text::PaNodeText;
+#[cfg(feature = "sys")]
 use paqual_name::paqual_name;
+
 pub use parse::parse;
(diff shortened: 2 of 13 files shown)


Hywan:
The CI is failing because running `cargo check` from the workspace seems to enable all the features (and the `sys` + `js` features are mutually exclusive, so it fails). I don't why there is this behavior, that's surprising. cc @jplatte

andybalaam:
> I wanted to reuse the existing parse::parse tests, but speculoos is used, which depends on num, which depends on rustc-serialize, which doesn't compile to Wasm properly, sadly.

Happy to switch to something other than speculoos if there is a decent alternative.

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -12,256 +12,513 @@
...
+        PaDomCreator::parse(html)
+            .map(padom_to_dom)
+            .map_err(padom_creation_error_to_dom_creation_error)
     }
 
-    /// Create a br node
-    fn new_line_break<S>() -> DomNode<S>
+    /// Convert a [PaDom] into a [Dom].
+    ///
+    /// [PaDom] is purely used within the parsing process (using html5ever) - in it,
```
andybalaam, line 52: By the way, this whole PaDom structure - if you can see a better way to do it, I'd be really interested!

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -12,256 +12,513 @@
...
-    } else {
-        panic!("Document was not a document!");
-    }
+            for nth in 0..number_of_nodes {
+                let node = nodes.get(nth as _).unwrap();
 
-    ret
-}
+                match node.node_name().as_str() {
+                    "BR" => {
```
andybalaam, line 355: OK, this is much better than I thought, because we retain fine-grained control over which tags we accept and how we convert them into Dom objects.

jplatte:
> The CI is failing because running `cargo check` from the workspace seems to enable all the features (and the `sys` + `js` features are mutually exclusive, so it fails). I don't why there is this behavior, that's surprising. cc @jplatte

Well there is one crate in the workspace that activates `sys` and another one that activates `js`. Cargo unifies features, so it doesn't have to compile the same crate twice. It can do this because features are meant to be additive. You will have to update CI to test the two bindings crates individually, but I also worry that this will break usage of rust-analyzer in this workspace. I don't have a good idea on how to fix that.

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -12,256 +12,513 @@
...
+            );
+
+            let dom = parse.unwrap();
+            let html_again = restore_whitespace(&dom.to_html().to_string());
+
+            assert_eq!(html, html_again);
+        }
+
+        #[wasm_bindgen_test]
+        fn formatting() {
```
andybalaam, line 495: I'd be really keen to share identical tests between the two features, if we can find a way.

andybalaam:
This is much better than I thought - we retain control over which tags we create and how we import them. Thank you for looking into this - I think it's the right way to go! Anything we can do to reduce the complexity of having 2 code paths would be great, but it's worth it for the code size gain.

andybalaam:
> It can do this because features are meant to be additive.

Hmm. Is this a fatal flaw in the approach? Can we find a way to make them additive?

jplatte:
Well, many crates are getting away just fine with mutually exclusive features (matrix-rust-sdk included). This is a special case in that there is a workspace around the crate with dependent crates that each want to activate one of the mutually exclusive features.

crates/wysiwyg/src/dom/parser.rs
```diff
@@ -19,22 +19,39 @@
...
 mod padom_node;
+#[cfg(feature = "sys")]
 mod panode_container;
+#[cfg(feature = "sys")]
 mod panode_text;
+#[cfg(feature = "sys")]
 mod paqual_name;
 mod parse;
 
+#[cfg(feature = "sys")]
```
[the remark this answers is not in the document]
Hywan: Done, https://github.com/matrix-org/matrix-wysiwyg/pull/225/commits/c1f888184c599d5b8a60752268d4046559715fb4.

crates/wysiwyg/src/dom/parser/parse.rs
Hywan: Exactly. It maps to what is already done with `html5ever`.

crates/wysiwyg/src/dom/parser/parse.rs
Hywan: Yes, I would like to even have the same tests shared between both!

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -72,6 +72,15 @@ where
         self.document().children()
     }
 
+    #[cfg(feature = "js")]
```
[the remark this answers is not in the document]
Hywan: If we remove this `#[cfg]`, when the `js` feature is turn off, the compiler will raise a warning (saying it's unused). I don't like warnings. If someone would like to use it for something else than `js`, this someone is free to remove the `#[cfg]`.

I've added this method to avoid having to clone the `&Vec<DomNode<S>>` returned by `children()`.

crates/wysiwyg/src/dom/parser/parse.rs
Hywan: Let me address that in another PR maybe :-).

Hywan:
> > The CI is failing because running `cargo check` from the workspace seems to enable all the features (and the `sys` + `js` features are mutually exclusive, so it fails). I don't why there is this behavior, that's surprising. cc @jplatte
> 
> Well there is one crate in the workspace that activates `sys` and another one that activates `js`. Cargo unifies features, so it doesn't have to compile the same crate twice. It can do this because features are meant to be additive. You will have to update CI to test the two bindings crates individually, but I also worry that this will break usage of rust-analyzer in this workspace. I don't have a good idea on how to fix that.

Ohhh, thanks for the explanations. It was driving me nuts.

Hywan:
> > It can do this because features are meant to be additive.
> 
> Hmm. Is this a fatal flaw in the approach? Can we find a way to make them additive?

The tags are no more mutually exclusives, except in the `parse` function. So the error has disappeared.

Hywan:
Regarding my last commit, I've opened a new issue to explain why the bugs have been fixed, but they are still present in other targets that use `html5ever`, https://github.com/matrix-org/matrix-wysiwyg/issues/242.

andybalaam:
Nice work, thank you!

PR #226, opened 2022-10-19
[Android] State restoration with ViewModel

jmartinesp:
* Replaces  `InputProcessor` with `EditorViewModel`.
* Renames `ComposerModel.dumpState()` to `ComposerModel.getCurrentDomState()` and creates `ComposerModel.getCurrentMenuState()` to get the current menu state.
* The initial menu state is computed when the ComposerModel is created now.
* Creates `ResourcesProvider` interface so the `HtmlToSpanParser` doesn't need a `Context` to work.


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -5,32 +5,34 @@ import android.content.ClipboardManager
...
     private var inputConnection: InterceptInputConnection? = null
-    private val inputProcessor = InputProcessor(
-        context,
-        menuStateCallback = { menuStateChangedListener?.menuStateChanged(it) },
-        // Using the returned ComposerModel automatically loads the native libraries and will crash
-        // layout preview and other tools. We're making it nullable as a workaround for that.
-        composer = if (isInEditMode) null else newComposerModel()
-    )
+    private val viewModel: EditorViewModel by lazy {
+        val viewModelStoreOwner = findViewTreeViewModelStoreOwner()
```
jmartinesp, line 31: I had this working before by getting the `ViewModelStoreOwner` of the `context` and using a custom key to get the right ViewModel instance. This worked as soon as the EditText was instantiated.

The new approach with `findViewTreeViewModelStoreOwner()` is simpler, but it will only work once the view has been attached to the current window.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -39,6 +41,8 @@ class EditorEditText : TextInputEditText {
         }
     }
 
+    private var isInitialized = false
```
jmartinesp, line 44: Used to understand when the view is ready to be used.

platforms/android/library/src/main/java/io/element/android/wysiwyg/viewmodel/EditorViewModel.kt
```diff
@@ -1,36 +1,49 @@
...
-    private val composer: ComposerModelInterface?,
-) {
+internal class EditorViewModel : ViewModel() {
+
+    private var composer: ComposerModel? = null
+    private var menuStateCallback: ((MenuState) -> Unit)? = null
+    private lateinit var resourcesProvider: ResourcesProvider
+
+    fun initialize(context: Context, isInEditMode: Boolean) {
+        resourcesProvider = AndroidResourcesProvider(context)
```
jonnyandrew, line 29: Should we try to avoid holding a reference to the context in the ViewModel?

platforms/android/library/src/main/java/io/element/android/wysiwyg/viewmodel/EditorViewModel.kt
```diff
@@ -1,36 +1,42 @@
...
-import kotlin.math.absoluteValue
+import uniffi.wysiwyg_composer.newComposerModel
 
-internal class InputProcessor(
-    private val context: Context,
-    private val menuStateCallback: (MenuState) -> Unit,
-    private val composer: ComposerModelInterface?,
-) {
+internal class EditorViewModel(
+    private val resourcesProvider: ResourcesProvider,
```
jonnyandrew, line 21: We'd still be holding a reference to the view context here which could leak it. If we need to do this, I think it could be better to use [`AndroidViewModel`](https://developer.android.com/reference/androidx/lifecycle/AndroidViewModel) that only holds the application context (which there is only one instance of).

jonnyandrew (approved):

PR #233, opened 2022-10-21
Further attempt to reduce flakiness

andybalaam:


platforms/web/cypress/e2e/clipboard/cut.spec.ts
```diff
@@ -20,7 +20,8 @@ describe('Cut', () => {
     const editor = '.editor:not([disabled])[contenteditable="true"]';
 
     it('remove text that is cut to clipboard', { browser: 'electron' }, () => {
-        cy.visit('/').wait(300);
+        cy.visit('/');
+        cy.get(editor).wait(500);
```
florianduros, line 24: I wonder if we can wait until the `contentEditable` field is at true ?

```ts
cy.get('#editor',  { timeout: 1000 }).should('have.attr', 'contentEditable', true)
```

(I'm not familiar with the cypress syntax, so it should not work out of box ^^)

andybalaam: The `editor` variable already says that `[contenteditable="true"]` so it should not match the `get` until it is editable.

I don't know why some of the characters don't type :-(

andybalaam: Unless my CSS selector is wrong?

florianduros (approved):

platforms/web/cypress/e2e/clipboard/cut.spec.ts
florianduros: The selectors seems to be good

andybalaam: Merging, in case it helps :-)

andybalaam: I did some local testing and it seems slightly more reliable so fingers crossed.

PR #235, opened 2022-10-25
[Android] Add plain text APIs

jonnyandrew:
[`PSU-780`](https://element-io.atlassian.net/browse/PSU-780)

Adds two a new public APIs to the `EditorEditText`:

### `EditorEditText#setPlainText`
Set the text without interpretting HTML formatting. In future, this can be extended to accept markdown formatted plain text. 

### `EditorEditText#getPlainText`
Get the text without any HTML formatting. In future, this can be extended to return markdown formatted plain text.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlConverter.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlConverter.kt
@@ -4,9 +4,12 @@ import androidx.core.text.HtmlCompat
 
 internal interface HtmlConverter {
     fun fromHtmlToPlainText(html: String): String
+    fun fromHtmlToSpans(html: String): CharSequence
 }
 
-internal object AndroidHtmlConverter: HtmlConverter {
+internal class AndroidHtmlConverter(
+    private val provideHtmlToSpansParser: (html: String) -> HtmlToSpansParser
+): HtmlConverter {
     /**
      * Get the content with formatting removed.
      * TODO: Return markdown formatted plaintext instead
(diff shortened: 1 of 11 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlConverter.kt
```diff
@@ -0,0 +1,18 @@
...
+    fun fromHtmlToPlainText(html: String): String
+}
+
+internal object AndroidHtmlConverter: HtmlConverter {
+    /**
+     * Get the content with formatting removed.
+     * TODO: Return markdown formatted plaintext instead
+     */
+    override fun fromHtmlToPlainText(html: String): String =
+        HtmlCompat.fromHtml(
```
jmartinesp, line 15: We could probably use `HtmlToSpansParser` here since it's our custom Html parser.

jonnyandrew: I'm not sure the HTML to spans implementation necessarily needs to overlap with the HTML to plain text implementation as eventually converting HTML to plain text will be done completely separately (using the Rust model) anyway. I could see value in combining the two interfaces though. What are your thoughts?

jonnyandrew: I've refactored a bit to combine the HTML utilities under one interface: [`feecd52`](https://github.com/matrix-org/matrix-wysiwyg/pull/235/commits/feecd5269d8943a89d3cc57f33bd0a1ef931a0e6).

jmartinesp (approved):
LGTM! Great job!

PR #238, opened 2022-10-26
Rust & Android: Fix crash when deleting emoji using HW keyboard

jmartinesp:
Steps:
- Write an emoji.
- Press backspace in HW keyboard when the cursor is right past it.

If it doesn't crash anymore, it's fixed.

### DIFF
--- a/crates/wysiwyg/src/composer_model/delete_text.rs
+++ b/crates/wysiwyg/src/composer_model/delete_text.rs
@@ -167,7 +167,7 @@ where
         let graphemes = str.find_graphemes_at(pos);
         // Take the grapheme before the position
         if let Some(last_grapheme) = graphemes.0 {
-            S::from(last_grapheme.as_str()).len()
+            last_grapheme.len()
         } else {
             // Default length for characters
             1
@@ -179,7 +179,7 @@ where
         let graphemes = str.find_graphemes_at(pos);
         // Take the grapheme after the position
         if let Some(first_grapheme) = graphemes.1 {
-            S::from(first_grapheme.as_str()).len()
+            first_grapheme.len()
         } else {
             // Default length for characters
             1
--- a/crates/wysiwyg/src/dom/unicode_string.rs
+++ b/crates/wysiwyg/src/dom/unicode_string.rs
@@ -59,6 +59,7 @@ pub trait UnicodeStr:
     + Index<RangeTo<usize>, Output = Self>
 {
     type CodeUnit: Copy + From<u8> + PartialEq;
+    type StringType: UnicodeString;
 
     // Should really be `-> Self::Chars<'a>`, but that requires GATs
     fn chars(&self) -> Box<dyn Iterator<Item = char> + '_>;
@@ -78,6 +79,7 @@ impl UnicodeString for String {
 
 impl UnicodeStr for str {
     type CodeUnit = u8;
+    type StringType = String;
 
     fn chars(&self) -> Box<dyn Iterator<Item = char> + '_> {
         Box::new(self.chars())
@@ -99,6 +101,7 @@ impl UnicodeString for Utf16String {
 
 impl UnicodeStr for Utf16Str {
     type CodeUnit = u16;
+    type StringType = Utf16String;
 
     fn chars(&self) -> Box<dyn Iterator<Item = char> + '_> {
         Box::new(self.chars())
@@ -120,6 +123,7 @@ impl UnicodeString for Utf32String {
 
 impl UnicodeStr for Utf32Str {
     type CodeUnit = u32;
+    type StringType = Utf32String;
 
     fn chars(&self) -> Box<dyn Iterator<Item = char> + '_> {
         Box::new(self.chars())
@@ -152,7 +156,7 @@ pub trait UnicodeStrExt: UnicodeStr {
     fn find_graphemes_at(
         &self,
         index: usize,
-    ) -> (Option<String>, Option<String>);
+    ) -> (Option<Self::StringType>, Option<Self::StringType>);
     fn u8_map_index(&self, pos: usize) -> usize;
 }
 
@@ -170,7 +174,7 @@ impl<S: UnicodeStr + ?Sized> UnicodeStrExt for S {
     fn find_graphemes_at(
         &self,
         index: usize,
-    ) -> (Option<String>, Option<String>) {
+    ) -> (Option<Self::StringType>, Option<Self::StringType>) {
         let u8_str = self.to_string();
         let u8_index = self.u8_map_index(index);
         let graphemes = u8_str.grapheme_indices(true);
@@ -178,11 +182,11 @@ impl<S: UnicodeStr + ?Sized> UnicodeStrExt for S {
         let mut next = None;
         for g in graphemes {
             if g.0 == u8_index {
-                next = Some(g.1.to_string())
+                next = Some(Self::StringType::from(g.1))
             }
             let length = g.1.len();
             if g.0 + length == u8_index {
-                prev = Some(g.1.to_string())
+                prev = Some(Self::StringType::from(g.1))
             }
         }
         (prev, next)


jonnyandrew (commented):
Nice find! This logic is a pretty complex though. Could we extract and unit test?

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -170,7 +172,17 @@ internal class InterceptInputConnection(
...
+        var previousCharLength = 1
+        // Look for graphemes, not UTF-16 indexes
+        if (start == end) {
+            val boundary = BreakIterator.getCharacterInstance()
+            boundary.setText(editable.toString())
+            if (boundary.isBoundary(end)) {
+                boundary.following(end)
+                previousCharLength = boundary.current() - boundary.previous()
+            }
+        }
```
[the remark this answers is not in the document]
jmartinesp: They both work, but your suggestion is simpler. Thanks!

jmartinesp:
I made quite a few changes to the approach I used. I was trying to fix this issue for Android but it's actually present on all platforms, so instead I changed the Rust code so it would automatically detect emojis when trying to delete text, either forwards or backwards.

I also had to do some changes to the Android code to work well with these changes in Rust.

crates/wysiwyg/src/tests/test_deleting.rs
```diff
@@ -222,3 +222,37 @@ fn deleting_a_newline_deletes_it() {
...
+    let mut model = cm("😄|😅");
+    model.backspace();
+    assert_eq!(tx(&model), "|😅");
+}
+
+#[test]
+fn test_backspace_complex_emoji() {
+    let mut model = cm("Test😮‍💨|😅");
+    model.backspace();
+    assert_eq!(tx(&model), "Test|😅");
```
andybalaam, line 237: Are we sure we want this behaviour?

In the github editor, if I paste in 😮‍💨 and then backspace once I get 😮. So it only deleted one codepoint.

I think the crash came because we tried to delete a partial codepoint, not a partial grapheme (but I'm not sure)

andybalaam: If we want to delete individual codepoints (which I think we do) we can use `String.chars` instead of `graphemes`.

If we want to delete graphemes, I think we should use https://crates.io/crates/unicode-segmentation which has a `graphemes` method. The code we have here only supports zero-width-joiners, but there are lots of "combining" characters that don't need a joiner - they just combine automatically (e.g. Combining Acute Accent (U+0301)). I assume unicode-segmentation handles those.

crates/wysiwyg/src/dom/unicode_string.rs
```diff
@@ -59,6 +59,42 @@ pub trait UnicodeStr:
...
+    /// Splits the current UnicodeStr into graphemes (visual characters), from start to end indexes
+    fn graphemes(&self, start: usize, end: usize) -> Vec<&Self> {
+        let mut ret = Vec::new();
+        let mut start = start;
+        let mut offset = 0;
+        let mut chars = self.chars().peekable();
+        while chars.peek().is_some() {
+            let c = chars.next().unwrap();
+            let char_len = self.char_len(&c);
+            let new_offset = offset + char_len;
```
andybalaam, line 75: I think the code works (I tried to break it with a test and couldn't), but I can't understand why we don't need `new_offset += char_len;` here. Surely every time around the loop we should move further to the right?

andybalaam: Ah, I found where we update `offset` now, ignore me!

crates/wysiwyg/src/composer_model/delete_text.rs
```diff
@@ -71,8 +72,24 @@ where
...
-            // Go forward 1 from the current location
-            self.state.end += 1;
+            let mut next_char_len: isize = 1;
+            let (s, e) = self.safe_selection();
+            let range = self.state.dom.find_range(s, e);
+            let leaves: Vec<&DomLocation> = range.leaves().collect();
+            if s == e && leaves.len() == 1 {
+                let leaf = leaves[0];
+                if let DomNode::Text(text_node) =
+                    self.state.dom.lookup_node(&leaf.node_handle)
```
andybalaam, line 82: This chunk of code is repeated between here and do_backspace - I think it should be in a shared function.

andybalaam: It's also a bit awkward - we actually just want "which text node is at this location?" - maybe we should write a function to answer that question. (Bear in mind sometimes there is no text node - e.g. when the cursor is between 2 BRs)

crates/wysiwyg/src/dom/unicode_string.rs
```diff
@@ -59,6 +59,42 @@ pub trait UnicodeStr:
...
+    fn char_len(&self, char: &char) -> usize;
+
+    /// Splits the current UnicodeStr into graphemes (visual characters), from start to end indexes
+    fn graphemes(&self, start: usize, end: usize) -> Vec<&Self> {
+        let mut ret = Vec::new();
+        let mut start = start;
+        let mut offset = 0;
+        let mut chars = self.chars().peekable();
+        while chars.peek().is_some() {
+            let c = chars.next().unwrap();
```
andybalaam, line 73: Can we not do `for char in chars` ?

andybalaam (changes requested):
Really nice that you found this bug but some questions about the fix.

crates/wysiwyg/src/tests/test_deleting.rs
jmartinesp: > Are we sure we want this behaviour?

On Android at least, that's the expected behaviour. If I paste `😮‍💨` here (GH editor) and I backspace the whole emoji gets deleted, so maybe it's a per OS implementation?

> I think the crash came because we tried to delete a partial codepoint

Yes, that seems to be the case.

> If we want to delete graphemes, I think we should use https://crates.io/crates/unicode-segmentation which has a graphemes method.

`unicode-segmentation` only works for UTF-8 strings. I tried using it at first but as soon as I passed a UTF-16 string everything crashed. That's why I implemented this other solution, trying to use it as an inspiration.

> The code we have here only supports zero-width-joiners, but there are lots of "combining" characters that don't need a joiner

To be honest, I haven't actually tried those. I'll do and come back with my results.

crates/wysiwyg/src/dom/unicode_string.rs
jmartinesp: I tried doing that at first, but we need to know both the current and the next character (checking for ZWJ) to understand if the current slice should be added as a grapheme or not. The easiest way I found to do that was using `peek()`.

crates/wysiwyg/src/tests/test_deleting.rs
jmartinesp: As you thought, the code only handles emoji properly, for those combining characters, each character is treated as a different cluster.

I took a look at the `UnicodeSegmentation` trait, trying to figure out if we could implement it for `Utf16Str` and `Utf32Str`, but it turns out it's hardcoded to return a `Graphemes` struct that contains a `str`, so the crate was only thought to work with Rust strings.

I could *try* to mimick the implementation of the crate or even reuse some of its utils to improve the logic used here, but it would take quite some time and I think there are more pressing issues at the moment. Would we be happy using this to fix the emoji crashes with a TODO to properly support graphemes in the future?

andybalaam: I think I'd prefer to just delete individual code points, rather than provide something that looks like it works, but then doesn't in some other circumstances. What do you think?

andybalaam: If we're really committed to graphemes, I'd want to use UnicodeSegmentation, and convert back and forth to/from UTF-8.

crates/wysiwyg/src/dom/unicode_string.rs
andybalaam: Ah, I missed that you were using both, sorry.

crates/wysiwyg/src/tests/test_deleting.rs
jmartinesp: > I think I'd prefer to just delete individual code points, rather than provide something that looks like it works, but then doesn't in some other circumstances. What do you think?

I think that's probably the best compromise at the moment.

> If we're really committed to graphemes, I'd want to use UnicodeSegmentation, and convert back and forth to/from UTF-8.

I guess this is possible too, but we'd have to map the indexes between the encodings very carefully. I can give this a try and if I see that it takes too much time, go back to just handling codepoints.

andybalaam: I would suggest just code points for now. If we want to do graphemes we probably need to discuss with product. We might want to match platform behaviour, or we might want to be consistent.

crates/wysiwyg/src/dom/unicode_string.rs
```diff
@@ -61,6 +62,42 @@ pub trait UnicodeStr:
 
     // Should really be `-> Self::Chars<'a>`, but that requires GATs
     fn chars(&self) -> Box<dyn Iterator<Item = char> + '_>;
+
+    /// Returns the length of the char in indices of the current encoding
+    fn char_len(&self, char: &char) -> usize;
+
+    /// Splits the current UnicodeStr into graphemes (visual characters), from start to end indexes
+    fn graphemes_us(&self, start: usize, end: usize) -> Vec<&Self> {
```
andybalaam, line 70: Is this function used?

crates/wysiwyg/src/dom/unicode_string.rs
```diff
@@ -142,4 +197,124 @@ impl<S: UnicodeStr + ?Sized> UnicodeStrExt for S {
     fn len(&self) -> usize {
         self.as_ref().len()
     }
+
+    /// Assuming [index] is a boundary between graphemes, returns a pair with the previous and next
+    /// graphemes, if present.
+    fn find_graphemes_at(
```
andybalaam, line 203: Could `find_graphemes_at` take the `pos` in its natural units (i.e. UTF-16 code units for us), and then call `u8_mapped_indexes` inside itself? This would make the interface much less confusing I think.

andybalaam (changes requested):
Getting there, but hoping we can simplify this a little more!

crates/wysiwyg/src/dom/unicode_string.rs
jmartinesp: No, my bad 😅 . This is the previous function, I left it there while implementing the new version and forgot to remove it.

crates/wysiwyg/src/dom/unicode_string.rs
jmartinesp: I'll take a look.

crates/wysiwyg/src/dom/unicode_string.rs
```diff
@@ -132,6 +149,11 @@ impl<S: UnicodeString> UnicodeStringExt for S {
 pub trait UnicodeStrExt: UnicodeStr {
     fn is_empty(&self) -> bool;
     fn len(&self) -> usize;
+    fn find_graphemes_at(
+        &self,
+        index: usize,
+    ) -> (Option<String>, Option<String>);
```
andybalaam, line 155: Sorry, one last thing! Can this return `Self`, so we do the conversion back to S inside the implementation?

andybalaam (changes requested):

PR #243, opened 2022-10-26
[iOS] Add unit test for plain text mode

aringenbach:
* Updates the plainText property so the content is immediately relevant (otherwise it relied on having an existing UITextView bound to the model)
* Add a unit test representing expected results when the Markdown/HTML conversions will be supported


Velin92 (approved):
LGTM!

PR #245, opened 2022-10-27
iOS: Fixing missing keystroke bug + long press delete issue

Velin92:
Reworked the code to allow the delegates only to update the content of the textview to avoid the missing keystroke issue.
Also fixed the delete long press issue.
Wrote UI tests for these issues.

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
@@ -21,15 +21,14 @@ import SwiftUI
 public struct WysiwygComposerView: UIViewRepresentable {
     // MARK: - Internal
 
-    public var viewModel: WysiwygComposerViewModel
-    
+    private var viewModel: WysiwygComposerViewModelProtocol
     private var tintColor = Color.accentColor
     private var placeholderColor = Color(UIColor.placeholderText)
     private var placeholder: String?
     @Binding public var focused: Bool
 
     public init(focused: Binding<Bool>,
-                viewModel: WysiwygComposerViewModel) {
+                viewModel: WysiwygComposerViewModelProtocol) {
         _focused = focused
         self.viewModel = viewModel
     }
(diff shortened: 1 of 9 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
```diff
@@ -21,25 +21,17 @@ import SwiftUI
 public struct WysiwygComposerView: UIViewRepresentable {
     // MARK: - Internal
 
-    public var content: WysiwygComposerContent
-    public var replaceText: (NSAttributedString, NSRange, String) -> Bool
-    public var select: (NSAttributedString, NSRange) -> Void
-    public var didUpdateText: (UITextView) -> Void
+    public var viewModel: WysiwygComposerViewModel
```
aringenbach, line 24: Should we move this to a protocol including all functions that are actually relevant for the UITextView ? Right now it will have access to a bunch of functions that are not necessarily useful from the UITextView perspective

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
```diff
@@ -105,20 +94,16 @@ public struct WysiwygComposerView: UIViewRepresentable {
...
         }
 
         public func textViewDidChangeSelection(_ textView: UITextView) {
             Logger.textView.logDebug([textView.logSelection],
                                      functionName: #function)
-            select(textView.attributedText, textView.selectedRange)
+            // Fixes long press delete issue
+            DispatchQueue.main.async {
+                self.select(textView.attributedText, textView.selectedRange)
+            }
```
aringenbach, line 106: I'm a bit worried about this part, but I think it's pretty hard (impossible ?) to make UI tests that will check for cursor positions to make sure it is always correct

aringenbach (approved):
LGTM

Raised two points that might be worth discussing now or later.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
Velin92: Good idea, will work on that!

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
Velin92: sadly this is the only way I found to fix the long press delete skipping some text issue.
I tested the view myself and dispatching the selection on the next UI update seems to not affect the user experience in any way.
But yeah would be nice to have a test for this, but I have no idea how to write it.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
Velin92: Added the protocol

aringenbach (approved):
LGTM

PR #247, opened 2022-10-27
[Android] Add tests for view model

jonnyandrew:
- Add unit tests for the view model core logic
- Refactor to combine view model methods that are always used together

[`PSU-945`](https://element-io.atlassian.net/browse/PSU-945)


jmartinesp (approved):
LGTM! Thanks for the changes!

PR #248, opened 2022-10-28
[Android] Fix text watchers called with empty string on every change

jonnyandrew:
This is needed because the editor currently uses `Editable.replace` to replace the entire contents of the editor when the model changes. This method does not replace spans that are not contained within the range resulting in spans which cover the whole range being duplicated.

[`PSU-935`](https://element-io.atlassian.net/browse/PSU-935) [`PSU-910`](https://element-io.atlassian.net/browse/PSU-910)

### DIFF
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
@@ -3,7 +3,6 @@ package io.element.android.wysiwyg.test
 import android.graphics.Typeface
 import android.text.Editable
 import android.text.Spannable
-import android.text.TextWatcher
 import android.text.style.BulletSpan
 import android.text.style.StyleSpan
 import android.text.style.UnderlineSpan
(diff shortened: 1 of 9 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -43,6 +45,7 @@ internal class HtmlToSpansParser(
 
     fun convert(): Spanned {
         parser.parse(InputSource(StringReader(html)))
+        if (BuildConfig.DEBUG) text.assertOnlyAllowedSpans()
```
jmartinesp, line 48: 👌

jmartinesp (approved):
LGTM! It seems to work pretty well and the approach is quite elegant.

platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
```diff
@@ -308,6 +311,31 @@ class EditorEditTextInputTests {
         }
     }
 
+    @Test
+    @Ignore("Failing in CI environment")
+    fun testTextWatcher() {
```
jonnyandrew, line 316: This test is failing only in the CI environment and I haven't managed to work out why.

@jmartinesp, I'll leave the PR open while I'm away in case you have any ideas how to fix this. Otherwise feel free to merge it as-is and I'll take a look when I'm back.

PR #249, opened 2022-10-28
[iOS] Fix bold+italic content in the composer on iOS 14/15

aringenbach:
Fixes the display of Bold+Italic content by moving from `subheadline` to `body` font


Velin92 (approved):
this PR is insanely long 😝 
LGTM!
Maybe in the future we could have some enum var that translates each system font metric to the corresponding html font name, so that we can customise the text size for the textView, depening on such enum var.

aringenbach:
> this PR is insanely long 😝 LGTM! Maybe in the future we could have some enum var that translates each system font metric to the corresponding html font name, so that we can customise the text size for the textView, depening on such enum var.

Yeah definitely, there is quite a few things we could/should be doing for better HTML customisation.

PR #256, opened 2022-11-03
chore: Remove warnings when `js` feature is used without a `wasm32` target architecture

Hywan:
Yup, everything is in the title.


andybalaam (approved):
Great, thanks.

PR #257, opened 2022-11-03
doc(wysiwyg): Explain when Markdown cannot be converted back to HTML

Hywan:
I hope it clarifies when and how compiling HTML to Markdown and then back to HTML is not possible.


andybalaam (approved):
Nice, thanks.

PR #258, opened 2022-11-03
feat(wysiwyg): Markdown compiles `<u>` to… `<u>` (raw HTML)

Hywan:
So that it improves the roundtrip from HTML to Markdown back to HTML.

Raw HTML is allowed by the spec, https://github.github.com/gfm/#raw-html.


andybalaam (approved):
Great, thanks!

PR #259, opened 2022-11-03
Don't use node-style assertions in Web code

andybalaam:


jmartinesp:
I tried the code and the backwards selection is not crashing, but why does this change fix the issue?

andybalaam:
When I backwards-select some text (in Web) and then click Bold, I get an error in the console. Do you not? @jmartinesp 

Error:

```
Uncaught ReferenceError: assert is not defined
    sr dom.ts:136
```

andybalaam:
This change avoids the use of `assert`, which is a node-specific function, not available in the browser.

jmartinesp:
> When I backwards-select some text (in Web) and then click Bold, I get an error in the console. Do you not?

Not really, I tried it with the assert version in both Chrome and Firefox and it seems to be working fine.

andybalaam:
Firefox 106.0.2

jmartinesp (approved):
LGTM!

jmartinesp:
> Firefox 106.0.2

It also happened on Chrome, so I don't think it's related to any concrete version.

PR #260, opened 2022-11-04
Expose get/set model with markdown to clients

aringenbach:
* Starting code for a Markdown/HTML parser that could be exposed later
* Bindings for WASM & UniFFI
    * Getting current model content as Markdown
    * Replacing all content with Markdown input


bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -27,6 +27,24 @@ impl ComposerModel {
         ))
     }
 
+    pub fn replace_all_text(
```
andybalaam, line 30: I think I would prefer the name `replace_all_markdown`. Was there a reason you went for "text"?

andybalaam (approved):
Looks great, with one suggested change.

andybalaam (approved):
Looks good!

PR #261, opened 2022-11-04
Provide a map of action to state in WASM

andybalaam:


platforms/web/lib/useListeners/event.ts
```diff
@@ -146,7 +96,12 @@ export function handleInput(
     testUtilities: TestUtilities,
     formattingFunctions: FormattingFunctions,
     inputEventProcessor?: InputEventProcessor,
-) {
+):
+    | {
+          content: string | undefined;
```
florianduros, line 101: `content?: string`

florianduros:
In the JS part, nothing to say :)

platforms/web/lib/useListeners/event.ts
andybalaam: Done, and also I added a rename, which I'd be grateful if you could review too please @florianduros

platforms/web/lib/useListeners/utils.ts
```diff
@@ -14,14 +14,29 @@ See the License for the specific language governing permissions and
...
-import { FORMATTING_ACTIONS } from '../constants';
-import { FormattingState, FormattingActions } from '../types';
+import { ACTION_TYPES } from '../constants';
+import { ActionTypes, ActionState, AllActionStates } from '../types';
 
-export function getDefaultFormattingStates() {
-    return FORMATTING_ACTIONS.reduce<
-        Record<FormattingActions, FormattingState>
-    >((acc, action) => {
+export function defaultActionStates(): AllActionStates {
```
florianduros, line 20: I prefer when the functions have a verb in their name to be more explicit

florianduros (changes requested):

platforms/web/lib/useListeners/utils.ts
florianduros: Yes but syntactically it's a function and you can't use it as `const myState = defaultActionStates`

andybalaam: Cool, renamed to createDefaultActionStates

florianduros (approved):

PR #262, opened 2022-11-04
[iOS] Enable switching plain text with markdown

aringenbach:
* Enables markdown/HTML translations when switching from/to plain text mode.
* Remove expected failures from now passing iOS tests.

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
@@ -62,12 +62,12 @@ public class WysiwygComposerViewModel: WysiwygComposerViewModelProtocol, Observa
 
     /// The composer content for plain text mode.
     public var plainTextModeContent: WysiwygComposerContent {
-        // TODO: convert plain text to HTML
-        WysiwygComposerContent(plainText: plainText,
-                               html: "",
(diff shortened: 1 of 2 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -62,12 +62,12 @@ public class WysiwygComposerViewModel: WysiwygComposerViewModelProtocol, Observa
 
     /// The composer content for plain text mode.
     public var plainTextModeContent: WysiwygComposerContent {
-        // TODO: convert plain text to HTML
-        WysiwygComposerContent(plainText: plainText,
-                               html: "",
-                               attributed: NSAttributedString(string: plainText),
-                               attributedSelection: .init(location: plainText.utf16Length,
-                                                          length: 0))
+        _ = model.setContentFromMarkdown(markdown: plainText)
```
Velin92, line 65: maybe we could add `@discardableResult` to make this line a bit nicer to read?

Velin92 (approved):
LGTM one small comment

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
aringenbach: I don't think we can, because this is a generated func from UniFFI

PR #266, opened 2022-11-07
[iOS] Disable hardware keyboard for UITests

aringenbach:
Set a non-strict expected failure and update assertion to an XCTAssertEqual.


aringenbach:
Looks like the CI is having an even harder time than usual:

```
WysiwygSharedTests.swift:130: Expected failure in -[WysiwygUITests.WysiwygUITests testTypingFast]: XCTAssertEqual failed: ("Optional("Some long text that I am going to type very fast")") is not equal to ("Optional("Sonthagnoy rft")")Reason: (Typing fast might fail on CI)
```

Velin92 (approved):
LGTM
The important thing is that the test is at least passing locally, I checked online and this might be happening because Xcode 14 enforces on sims the connect hardware keyboard flag always on, while we can set it off on sims.
I'll look if there is a way to enforce in the xcscheme to always have this flag disabled

PR #268, opened 2022-11-07
Refactor WysiwygComposerContent into two entities

aringenbach:
WysiwygComposerContent is now divided into two objects. This will avoid storing and publishing unrequired parts of the content on every update. 

* A published object containing the attributed text and the selection, which is everything that the TextView actually requires. This bit could probably be reworked to be completely internal (as well as the protocol between the TextView and the ViewModel) but there is an issue preventing it right now:
    * Workaround for links that needs to manually reset the selection (which might be required on clients as well, depending on the implementation)
* A computed property returning both the Markdown and the HTML content. This is only needed for example when sending the message, or when saving a partial message (this also removes having to create a half-dummy content for plain text mode when it's needed)

This is a breaking change for iOS implementations and will require an update on client to use the new APIs.

### DIFF
--- a/platforms/ios/example/Wysiwyg/Views/ContentView.swift
+++ b/platforms/ios/example/Wysiwyg/Views/ContentView.swift
@@ -37,11 +37,7 @@ struct ContentView: View {
             viewModel.plainTextMode.toggle()
         }
         Button("Send") {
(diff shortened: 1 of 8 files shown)


Velin92 (approved):
LGTM

PR #271, opened 2022-11-08
Fix typo on iOS workflow for GitHub Actions

aringenbach:
This typo prevented the iOS workflow to be re-triggered on PR updates.


andybalaam (approved):
Looks good, confirmed at https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

PR #272, opened 2022-11-08
[iOS] Fix clearing the content of the text view when plain text mode is enabled

aringenbach:
* Clear the content of the text view in plain text mode
* Removes unnecessary property storing the content of the textView in plain text mode


Velin92 (approved):
LGTM

PR #290, opened 2022-11-11
Add automated release with tags for Android & Web

jmartinesp:
I'll change the docs once we're sure that the jobs work as expected.


.github/workflows/publish.yml
```diff
@@ -1,8 +1,63 @@
...
+          rustup target add aarch64-linux-android
+
+      - name: 🔨 Install Uniffi
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          # keep in sync with uniffi dependency in Cargo.toml's
+          args: uniffi_bindgen --version ${{ env.UNIFFI_VERSION }}
+
+      - name: 🚀 Publish to Sonatype
```
andybalaam, line 49: Should this be "Publish to Maven Central?"

update_version.sh
```diff
@@ -0,0 +1,30 @@
...
+  SED_CMD="sed -i"
+fi
+
+VERSION=$1
+
+echo "Updating Rust"
+$SED_CMD "s/^version\s*=\s*\".*\"/version = \"$VERSION\"/g" bindings/wysiwyg-ffi/Cargo.toml
+$SED_CMD "s/^version\s*=\s*\".*\"/version = \"$VERSION\"/g" bindings/wysiwyg-wasm/Cargo.toml
+$SED_CMD "s/^version\s*=\s*\".*\"/version = \"$VERSION\"/g" crates/wysiwyg/Cargo.toml
+$SED_CMD "s/\"version\":\s*\".+\"/\"version\": \"$VERSION\"/g" bindings/wysiwyg-wasm/package.json
```
andybalaam, line 24: This is part of Web, so should be moved down

andybalaam (approved):
Very minor comments - looks great - thanks!

.github/workflows/publish.yml
jmartinesp: Technically, we're publishing to [Sonatype](s01.oss.sonatype.org) that later synchronises with Maven Central.

PR #297, opened 2022-11-14
[Android] Add RustCollector to catch and deal with Rust errors.

jmartinesp:
Just added a simple callback to catch the Rust errors from the client that integrates the library. It will also need some extra work to integrate with EA/Sentry.


andybalaam (approved):
Looks good thanks!

PR #298, opened 2022-11-14
Menu states are always returned after overwriting content

andybalaam:


bindings/wysiwyg-ffi/src/ffi_composer_update.rs
```diff
@@ -20,3 +20,103 @@ impl ComposerUpdate {
...
 }
+
+#[cfg(test)]
+mod test {
+    use std::{collections::HashMap, sync::Arc};
+
+    use crate::{ActionState, ComposerAction, ComposerModel, MenuState};
+
+    #[test]
+    fn initial_menu_update_is_populated() {
```
aringenbach, line 31: This one feels a bit weird in that particular context, but I suppose if we explicitly ask for that change it's correct to consider it can be undone. :)

aringenbach (approved):
Thanks, LGTM !

PR #301, opened 2022-11-14
Implement action states map in Rust

andybalaam:


aringenbach (approved):
LGTM

Out of curiosity, what's up with package-lock.json, are all those node modules supposed to be there ?

andybalaam:
> Out of curiosity, what's up with package-lock.json, are all those node modules supposed to be there ?

It looks like all these packages got removed in 6d603072070a39884435c677d5c412889cead056 and my change just adds them back.

@jmartinesp - any idea what happened there?

andybalaam:
No idea. Let's keep an eye - maybe a difference in our dev setups is causing churn.

PR #314, opened 2022-11-16
Add ZWSP to all list items

aringenbach:
Add or remove ZWSP when inserting or removing from a list

### DIFF
--- a/crates/wysiwyg/src/composer_model/lists.rs
+++ b/crates/wysiwyg/src/composer_model/lists.rs
@@ -202,10 +202,29 @@ where
         }
     }
 
+    fn remove_leading_zwsp(&mut self, list_item_handle: &DomHandle) {
+        let list_item_node = self.state.dom.lookup_node(list_item_handle);
+        let first_text_handle =
+            list_item_node.iter_text().next().unwrap().handle();
+        let first_text = self.state.dom.lookup_node_mut(&first_text_handle);
+        if let DomNode::Text(t) = first_text {
+            let mut text = t.data().to_string();
+            if text.starts_with("\u{200b}") {
+                text.remove(0);
+                t.set_data(text.into());
+            } else {
+                panic!(
+                    "First text of a list item should always start with ZWSP"
+                )
+            }
+        }
+    }
+
     fn move_list_item_content_to_list_parent(
         &mut self,
         list_item_handle: &DomHandle,
     ) -> ComposerUpdate<S> {
+        self.remove_leading_zwsp(list_item_handle);
         let list_item_node = self.state.dom.lookup_node(list_item_handle);
         if let DomNode::Container(list_item) = list_item_node {
             let list_item_children = list_item.children().clone();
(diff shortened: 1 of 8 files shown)


crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -274,11 +277,25 @@ where
...
+                            let mut owned_text: S = "\u{200b}".into();
+                            owned_text.push(text);
+                            owned_text
+                        } else {
+                            text.to_owned()
+                        })],
                     ));
+
+                if add_zwsp {
+                    // FIXME: is there a case where ZWSP is inserted in the middle of the selection ?
```
jmartinesp, line 294: Could you elaborate this a bit? Since we're sure that there's only 1 leaf node and it's a TextNode I can't think of any case, but I'm not sure I follow what you had in mind for this FIXME.

crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -274,11 +277,25 @@ where
             if let DomNode::Text(t) = node {
                 let text = t.data();
                 let index_in_parent = handle.index_in_parent();
+                let add_zwsp = !text.to_string().starts_with("\u{200b}");
```
jmartinesp, line 280: Since you have `TextNode.add_leading_zwsp`, couldn't that be used instead?

crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -253,9 +254,11 @@ where
                 list_type,
                 vec![DomNode::Container(ContainerNode::new_list_item(
                     "li".into(),
-                    vec![DomNode::new_text(S::default())],
+                    vec![DomNode::new_text("\u{200b}".into())],
```
jmartinesp, line 257: It would be nice if we could create a constant for this somewhere, but we'd have to have one for String, another one for UnicodeString... maybe it's not worth it.

jmartinesp (approved):
LGTM, with a few remarks, all optional, and a question.

The Android integration doesn't work 100% of the time, but that's probably something we need to fix in the platform.

crates/wysiwyg/src/composer_model/lists.rs
aringenbach: Yeah you're right, I'll remove this, I was thinking about cross node selections, but this is not this case yet. FIXME should be "implement for cross node selection" 😅

crates/wysiwyg/src/composer_model/lists.rs
aringenbach: Yeah I plan to use that when it is structured a bit better, right now the TextNode is re-created instead of moved and we have a non-mut context for it. I'll check if it is an easy fix, otherwise keeping it for later 😉

crates/wysiwyg/src/dom/unicode_string.rs
```diff
@@ -45,6 +45,11 @@ pub trait UnicodeString:
     type Str: UnicodeStr<CodeUnit = Self::CodeUnit, Owned = Self> + ?Sized;
 
     fn insert(&mut self, idx: usize, s: &Self::Str);
+
+    /// Creates a new unicode string consisting of a single ZWSP.
+    fn zwsp() -> Self {
```
jmartinesp, line 50: 👌

jmartinesp (approved):
LGTM

Issue #277, opened 2022-11-09
Spell checker / auto correct not working


jmartinesp:
Some users have mentioned that the automatic spell checking is not working.

- https://github.com/vector-im/element-android/issues/7463

jonnyandrew:
## Fndings (Android)

The spell check is running inside the editor but sometimes the spell check errors disappear and reappear.

## Problem
This seems to be related to how the rich text editor works internally, replacing the entire contents of the `EditText` on every key press and clearing the `Span`s applied by the spell checker.

## Solution
One solution could be to return incremental updates from the Rust model so that the app needs only to replace changed text and can thus preserve spell check suggestions. 

Related to-do:
https://github.com/matrix-org/matrix-rich-text-editor/blob/70a0e5074f106604e7eafc8e78df7e18527116de/crates/wysiwyg/src/composer_model/replace_text.rs#L183-L185

langleyd:
Markup and correction for autocorrection now works well on the latest.

PR #316, opened 2022-11-16
[iOS] Add a StringDiffer to help reconciliate the model with the content of the field

aringenbach:
Adds a StringDiffer that reconciliate content between the UITextView and the Rust model.

Fixes: 

* Magic replacements of the text view content from iOS, including:
  - Japanese auto-replacement from latin input (Romaji keyboard)
  - Korean replacement through combining symbols
  - Chinese keyboard suggestions (Pinyin keyboard)
  - Smart punctuation (e.g.: double space replaced by a dot)
  - iOS inconsistencies over characters (e.g.: `shouldChangeTextIn` providing U+0027 but actually using U+2019)
* Pending formats not applying during voice dictation
* Fixes backspace sometimes blocking on whitespaces
* Improves spell-checker/auto-correct since it lets the system handle more of the text view updates #277

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/CollectionDifference.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/CollectionDifference.swift
@@ -16,23 +16,32 @@
 
 import Foundation
 
+typealias UTF16Removals = [NSRange]
+typealias UTF16Insertions = [(range: NSRange, text: String)]
+
 extension CollectionDifference<Character> {
     /// Transforms the removals in the `CollectionDifference` into an array of ranges.
-    var removedRanges: [NSRange] {
+    /// Output ranges are defined with UTF16 locations and lengths.
+    ///
+    /// - Parameter string: String in which the characters have been removed.
+    /// - Returns: Array of removed ranges.
+    func utf16Removals(in string: String) -> UTF16Removals {
         removals.reduce([]) { partialResult, change in
             var partialRes = partialResult
             switch change {
-            case .remove(offset: let offset, element: _, associatedWith: _):
+            case .remove(offset: let offset, element: let element, associatedWith: _):
+                let index = string.index(string.startIndex, offsetBy: offset)
+                let utf16Offset = string[..<index].utf16.count
                 if let lastRange = partialRes.popLast() {
-                    if lastRange.upperBound == offset {
-                        partialRes.append(NSRange(location: lastRange.location, length: lastRange.length + 1))
+                    if lastRange.upperBound == utf16Offset {
+                        partialRes.append(NSRange(location: lastRange.location, length: lastRange.length + element.utf16.count))
                     } else {
                         partialRes.append(lastRange)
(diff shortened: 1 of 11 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Tools/StringDiffer.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        guard !difference.isEmpty else {
+            return nil
+        }
+
+        let removedRanges = difference.removedRanges
+        let textInsertions = difference.textInsertions
+
+        guard removedRanges.count < 2, textInsertions.count < 2 else {
+            throw StringDifferError.tooComplicated
```
Velin92, line 69: Could a case like this where the diff would return multiple removed ranges and text insertions actually happen?

aringenbach: Hopefully, it doesn't. We could be able to handle it, but I'd rather not waste time trying to match the removals to the correct insertions.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/CollectionDifference.swift
```diff
@@ -0,0 +1,83 @@
...
+    /// - Parameter string: String in which the characters have been removed.
+    /// - Returns: Array of removed ranges.
+    func utf16Removals(in string: String) -> UTF16Removals {
+        removals.reduce([]) { partialResult, change in
+            var partialRes = partialResult
+            switch change {
+            case .remove(offset: let offset, element: let element, associatedWith: _):
+                let index = string.index(string.startIndex, offsetBy: offset)
+                let utf16Offset = string[..<index].utf16.count
+                if let lastRange = partialRes.popLast() {
```
Velin92, line 35: `guard let lastRange = partialRes.popLast() else { return [NSRange(location: utf16Offset, length: element.utf16.count)] }`

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/CollectionDifference.swift
```diff
@@ -0,0 +1,83 @@
...
+    /// - Parameter string: String in which the characters have been inserted.
+    /// - Returns: Array of inserted ranges and text.
+    func utf16Insertions(in string: String) -> UTF16Insertions {
+        insertions.reduce([]) { partialResult, change in
+            var partialRes = partialResult
+            switch change {
+            case .insert(offset: let offset, element: let element, associatedWith: _):
+                let index = string.index(string.startIndex, offsetBy: offset)
+                let utf16Offset = string[..<index].utf16.count
+                if let lastRange = partialRes.popLast() {
```
Velin92, line 64: same comment as above

Velin92 (approved):
LGTM, just two small comments.
Also the fix actually also enable voice dictation to support the format toolbar right?

PR #317, opened 2022-11-17
Set the text view as a non optional ViewModel owned property

aringenbach:
Set the text view as non optional and move it to be directly owned by the view model. This will avoid a lot of unwrapping.


Velin92 (approved):
LGTM, can I ask you a favor? Could you also expose as a public(private set) `minHeight` (realised we use it in element-ios without knowing its true value, but always by assuming it)

PR #319, opened 2022-11-18
Fixing panic when adding a link to a nested formatted text - innermost approach

Velin92:
When trying to add a link to neste formatted text like the following: _test italic **test italic and bold**_ this caused a panic in the rust SDK. Now it's possible to handle these simple cases.
For now we are using a simple solution where we just add in place of the highlighted text node, a link node, even if this means adding the same link into multiple different nodes that live on different leaves.


jmartinesp (approved):
LGTM! (Although I'd change the 1st commit message 😛 ).

If we want to be more complete, maybe we could add an extra test for the future, when we properly support both lists and links using a more optimised approach:

```
cm("Te{st<ul><li>AB</li><li>C}|D</li></ul>")
```

Ideally, we'd also add one for a selection that already contains a link to check that it's replaced, but I think that's out of scope for the ticket. You could also create it and disable it using `#[ignore]`, so it's ready to be tested in the future.

PR #322, opened 2022-11-18
Change Existing Links + Min Rust version

Velin92:


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -292,6 +292,14 @@ where
             ),
         }
     }
+
+    pub(crate) fn set_link(&mut self, link: S) {
+        let ContainerNodeKind::Link(_) = self.kind else {
+            panic!("Setting list type to a non-list container is not allowed")
+        };
```
Velin92, line 299: This the new Rust guard only compiles on 1.65 Rust

bindings/wysiwyg-ffi/Cargo.toml
```diff
@@ -17,10 +17,10 @@ crate-type = ["cdylib", "staticlib"]
 # Keep the uniffi version here in sync with the installed version of
 # uniffi-bindgen that is called from
 # ../../examples/example-android/app/build.gradle
-uniffi = "=0.19.2"
-uniffi_macros = "=0.19.2"
+uniffi = "=0.21.0"
```
aringenbach, line 20: Can you change this value as well on both `.github/workflows/android.yml` and `.github/workflows/ios.yml` ? 

Plus, I suggest adding Android and iOS to make sure platform specific CI works as expected.

aringenbach (commented):
One small remark for when the Rust version issue is fixed

jmartinesp (approved):
LGTM! I would have split the PR in two (one for the Rust version, another for the link code), but it's alright.

Remember to squash and fix the commit messages!

PR #325, opened 2022-11-21
Inline code formatting for several nodes

jmartinesp:
Adds support for formatting several nodes using inline code. At first, I thought about adding an exception for `extend_format_in_multiple_nodes`, but the implementation is so different that I went with a whole different method instead.

### DIFF
--- a/crates/wysiwyg/src/composer_model/format.rs
+++ b/crates/wysiwyg/src/composer_model/format.rs
@@ -19,9 +19,7 @@ use crate::dom::action_list::DomActionList;
 use crate::dom::nodes::{ContainerNodeKind, DomNode};
 use crate::dom::unicode_string::UnicodeStrExt;
 use crate::dom::{Dom, DomHandle, DomLocation, Range};
-use crate::{
-    ComposerModel, ComposerUpdate, InlineFormatType, ToHtml, UnicodeString,
-};
+use crate::{ComposerModel, ComposerUpdate, InlineFormatType, UnicodeString};
 
 #[derive(Eq, PartialEq, Debug)]
 enum FormatSelectionType {
@@ -74,7 +72,7 @@ where
             )
         } else {
             let range = self.state.dom.find_range(s, e);
-            let leaves: Vec<&DomLocation> = range.leaves().collect();
+            let leaves = range.leaves();
             // We'll iterate through the leaves finding their closest structural node ancestor and
             // grouping these leaves based on the handles of these ancestors.
             let mut structure_ancestors = HashMap::new();
@@ -92,7 +90,7 @@ where
                 list.push(leaf.clone());
             }
 
-            // Order those ancestors
+            // Order those ancestors (important to avoid node replacement & conflicts of handles)
             let mut keys: Vec<&Vec<usize>> =
                 structure_ancestors.keys().collect();
             keys.sort();
@@ -104,6 +102,8 @@ where
                 let mut cur_text = S::default();
                 // Where we'll insert the result of merging the text contents
                 let mut insert_text_at: Option<DomHandle> = None;
+                // Nodes to be added to the Dom, might contain both TextNodes and LineBreaks
+                let mut nodes_to_add = Vec::new();
                 // Iterate the leaves backwards to avoid modifying the previous Dom structure
                 for leaf in leaves.into_iter().rev() {
                     // Find the immediate child of the common ancestor containing this leaf as its descendant
@@ -166,25 +166,12 @@ where
                         }
                         DomNode::LineBreak(_) => {
                             // We should split inline code nodes at line breaks
-                            let text = cur_text.clone().to_string();
                             if !cur_text.is_empty() {
-                                self.state.dom.insert(
-                                    &ancestor_child.next_sibling(),
-                                    DomNode::new_formatting(
-                                        InlineFormatType::InlineCode,
-                                        vec![DomNode::new_text(cur_text)],
-                                    ),
-                                );
+                                nodes_to_add
+                                    .insert(0, DomNode::new_text(cur_text));
                             }
-                            // If leaf line break is not a direct child of the common ancestor,
-                            // move it to the parent ancestor
-                            if ancestor_child != leaf.node_handle {
-                                self.state.dom.remove(&leaf.node_handle);
-                                self.state.dom.insert(
-                                    &ancestor_child.next_sibling(),
-                                    DomNode::new_line_break(),
-                                );
-                            }
+                            nodes_to_add.insert(0, DomNode::new_line_break());
+                            self.state.dom.remove(&leaf.node_handle);
                             // Update insertion point and reset text
                             insert_text_at = Some(ancestor_child.clone());
                             cur_text = S::default();
@@ -195,20 +182,24 @@ where
                     }
                 }
 
-                // If there is still some collected text not added to the Dom, insert it inside an
-                // inline code node.
+                // Insert the nodes into the Dom inside an inline code node
                 if insert_text_at.is_some() {
                     let insert_at = insert_text_at.unwrap();
 
+                    // If there is still some collected text add it to he list of nodes to insert
                     if !cur_text.is_empty() {
-                        self.state.dom.insert(
-                            &insert_at,
-                            DomNode::new_formatting(
-                                InlineFormatType::InlineCode,
-                                vec![DomNode::new_text(cur_text.clone())],
-                            ),
-                        );
+                        nodes_to_add
+                            .insert(0, DomNode::new_text(cur_text.clone()));
                     }
+
+                    // Insert the inline code node
+                    self.state.dom.insert(
+                        &insert_at,
+                        DomNode::new_formatting(
+                            InlineFormatType::InlineCode,
+                            nodes_to_add,
+                        ),
+                    );
 
                     // Merge inline code nodes for clean up
                     self.merge_formatting_node_with_siblings(&insert_at);
@@ -833,7 +824,7 @@ mod test {
     fn format_inline_code_same_row_with_line_breaks() {
         let mut model = cm("<b>{bold</b><br /><i>text}|</i>");
         model.inline_code();
-        assert_eq!(tx(&model), "<code>{bold</code><br /><code>text}|</code>");
+        assert_eq!(tx(&model), "<code>{bold<br />text}|</code>");
     }
 
     #[test]
@@ -863,7 +854,7 @@ mod test {
         model.inline_code();
         assert_eq!(
             tx(&model),
-            "<b><u>bo</u></b><code>{ld</code><br /><code>te}|</code><i>xt</i>"
+            "<b><u>bo</u></b><code>{ld<br />te}|</code><i>xt</i>"
         );
     }
 
@@ -874,7 +865,7 @@ mod test {
         model.inline_code();
         assert_eq!(
             tx(&model),
-            "<b><u>bo</u></b><code>{ld</code><br /><code>te}|</code><i>xt</i>"
+            "<b><u>bo</u></b><code>{ld<br />te}|</code><i>xt</i>"
         );
     }
 
@@ -906,7 +897,7 @@ mod test {
         model.inline_code();
         assert_eq!(
             tx(&model),
-            "Text <code>{before</code><br /><ul><li><code>bo}|</code><b>ld</b></li><li><i>text</i></li></ul>"
+            "Text <code>{before<br /></code><ul><li><code>bo}|</code><b>ld</b></li><li><i>text</i></li></ul>"
         );
     }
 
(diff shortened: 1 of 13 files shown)


crates/wysiwyg/src/lib.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+extern crate core;
```
andybalaam, line 15: What is this?

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -90,6 +90,27 @@ where
...
+            self.lookup_node_mut(parent_handle)
+        {
+            container
+        } else {
+            panic!("Parent is not a container node");
+        };
+        parent.append_child(child)
+    }
+
+    pub fn insert(&mut self, node_handle: &DomHandle, node: DomNode<S>) {
```
andybalaam, line 108: Maybe insert_before would be a good name? It took me a while to figure out what this did. Maybe also a doc comment.

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -90,6 +90,27 @@ where
         self.document_mut().append_child(child)
     }
 
+    pub fn append_child_to_parent(
```
andybalaam, line 93: Maybe append_inside would be a good name?

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
         model.bold();
         assert_eq!(tx(&model), "AAA&nbsp;|");
     }
+
+    #[test]
+    fn format_inline_code_same_row() {
```
andybalaam, line 810: inline_code_replacing_formatting_removes_formatting

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b>{bold</b><i>text}|</i>");
+        model.inline_code();
+        assert_eq!(tx(&model), "<code>{boldtext}|</code>");
+    }
+
+    #[test]
+    fn format_inline_code_same_row_partial() {
```
andybalaam, line 817: inline_code_replacing_partial_formatting_removes_overlapping_formatting

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b>bo{ld</b><i>te}|xt</i>");
+        model.inline_code();
+        assert_eq!(tx(&model), "<b>bo</b><code>{ldte}|</code><i>xt</i>");
+    }
+
+    #[test]
+    fn format_inline_code_same_row_with_line_breaks() {
```
andybalaam, line 824: inline_code_with_formatting_preserves_line_breaks

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b>{bold</b><br /><i>text}|</i>");
+        model.inline_code();
+        assert_eq!(tx(&model), "<code>{bold<br />text}|</code>");
+    }
+
+    #[test]
+    fn format_inline_code_different_rows() {
```
andybalaam, line 831: inline_code_replacing_complex_formatting_removes_formatting

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b><u>{bold</u></b><i>text}|</i>");
+        model.inline_code();
+        assert_eq!(tx(&model), "<code>{boldtext}|</code>");
+    }
+
+    #[test]
+    fn format_inline_code_different_rows_nested() {
```
andybalaam, line 838: inline_code_replacing_formatting_removes_formatting

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b><u>{bold</u><i>italic</i></b><i>text}|</i>");
+        model.inline_code();
+        assert_eq!(tx(&model), "<code>{bolditalictext}|</code>");
+    }
+
+    #[test]
+    fn format_inline_code_different_rows_partial() {
```
andybalaam, line 845: inline_code_partially_replacing_formatting_removes_overlap

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b><u>bo{ld</u></b><i>te}|xt</i>");
+        model.inline_code();
+        assert_eq!(tx(&model), "<b><u>bo</u></b><code>{ldte}|</code><i>xt</i>");
+    }
+
+    #[test]
+    fn format_inline_code_different_rows_partial_with_line_break() {
```
andybalaam, line 852: inline_code_on_partial_nested_line_break_removes_formatting

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        let mut model = cm("<b><u>bo{ld</u></b><br /><i>te}|xt</i>");
+        model.inline_code();
+        assert_eq!(
+            tx(&model),
+            "<b><u>bo</u></b><code>{ld<br />te}|</code><i>xt</i>"
+        );
+    }
+
+    #[test]
+    fn format_inline_code_different_rows_partial_with_line_break_inside_parent()
```
andybalaam, line 862: inline_code_on_partial_nested_line_break_within_parent_removes_formatting

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
...
+        model.inline_code();
+        assert_eq!(
+            tx(&model),
+            "Text {before<br /><ul><li>bo}|<b>ld</b></li><li><i>text</i></li></ul>"
+        );
+    }
+
+    // TODO: might need to re-visit it if we add ZWSP ant the end of inline code tags
+    #[test]
+    fn disable_inline_code_then_write_text() {
```
andybalaam, line 952: Shouldn't we have lots more disabling tests, assuming this works?

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -621,4 +805,154 @@ mod test {
         model.bold();
         assert_eq!(tx(&model), "AAA&nbsp;|");
     }
+
```
andybalaam, line 808: Awesome tests!

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -51,7 +53,189 @@ where
...
+        } else {
+            if parent.handle().has_parent() {
+                self.find_structure_ancestor(&parent.handle())
+            } else {
+                None
+            }
+        }
     }
 
     fn format_or_unformat(
```
andybalaam, line 241: I'd prefer if utility functions like format_or_unformat were at the bottom, below your new tests.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -51,7 +53,189 @@ where
...
+        if s == e {
+            self.toggle_zero_length_format(&format);
+            ComposerUpdate::update_menu_state(
+                self.compute_menu_state(MenuStateComputeType::KeepIfUnchanged),
+            )
+        } else {
+            let range = self.state.dom.find_range(s, e);
+            let leaves = range.leaves();
+            // We'll iterate through the leaves finding their closest structural node ancestor and
+            // grouping these leaves based on the handles of these ancestors.
```
andybalaam, line 77: This should probably be a separate method

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -51,7 +53,189 @@ where
...
+                        .node_handle
+                        .raw()
+                        .split_at(ancestor_handle.len() + 1)
+                        .clone();
+                    let ancestor_child =
+                        DomHandle::from_raw(path_to_child.to_vec());
+
+                    let node =
+                        self.state.dom.lookup_node(&leaf.node_handle).clone();
+                    match node {
```
andybalaam, line 120: Can we break out the arms of the match into methods, or similar? There is a lot of code here.

andybalaam (changes requested):
This is awesome!  Small changes requested and some new tests maybe.

crates/wysiwyg/src/composer_model/format.rs
jmartinesp: Yup. Will help @Velin92 with the link implementation too.

crates/wysiwyg/src/composer_model/format.rs
andybalaam: Sorry, I misread the diff! Ignore me.

crates/wysiwyg/src/composer_model/format.rs
andybalaam: Yeah, that sounds good.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -51,7 +53,63 @@ where
...
+            let list = structure_ancestors
+                .entry(ancestor_handle)
+                .or_insert(Vec::new());
+            // Add the DomHandle of the leaf to the list of grouped handles by this ancestor
+            list.push(leaf.clone());
+        }
+        structure_ancestors
+    }
+
+    pub(crate) fn remove_and_clean_up_empty_nodes_until(
```
andybalaam, line 88: I think you should re-use something like delete_nodes in crates/wysiwyg/src/composer_model/delete_text.rs here. It handles some corner cases I think.

crates/wysiwyg/src/composer_model/format.rs
```diff
@@ -51,7 +53,63 @@ where
...
+            && self.state.dom.parent(cur_handle).children().is_empty()
+        {
+            self.remove_and_clean_up_empty_nodes_until(
+                &cur_handle.parent_handle(),
+                top_handle,
+            )
+        }
+    }
+
+    fn find_structure_ancestor(&self, handle: &DomHandle) -> Option<DomHandle> {
```
andybalaam, line 104: Is this the same as find_struct_parent in crates/wysiwyg/src/composer_model/join_nodes.rs ?

crates/wysiwyg/src/composer_model/format_inline_code.rs
```diff
@@ -0,0 +1,337 @@
...
+                            cur_text = S::default();
+                        }
+                        _ => panic!(
+                            "Leaf should be either a line break or a text node"
+                        ),
+                    }
+                }
+
+                // Insert the nodes into the Dom inside an inline code node
+                if insert_text_at.is_some() {
```
andybalaam, line 96: Use `if let Some(insert_at) = insert_text_at` here instead of needing to unwrap

andybalaam (changes requested):
Almost there I think, couple of places we might be able to re-use code.

crates/wysiwyg/src/composer_model/format.rs
jmartinesp: Ouch, I knew I had already written something like this before, but I couldn't find it. The new version seems to be a better implementation though, I'll try to de-duplicate it.

crates/wysiwyg/src/composer_model/format.rs
jmartinesp: Yup, same as below, I had a feeling that we already had something like this but I couldn't really find it. I'll see if it works here too.

jmartinesp: Actually, it seems like it doesn't. A parent structure node can be temporarily empty and this method would actually remove it too, while previously we'd just stop at it if we found it.

crates/wysiwyg/src/composer_model/format.rs
jmartinesp: It seems like it's also not exactly what we need here, this function actually finds *the parent of an ancestor structure node* if it exists, otherwise, that structure node (if we found a ListItem, it would return its parent List). I should probably rename it and add some docs.

crates/wysiwyg/src/composer_model/format.rs
andybalaam: OK, no need to re-use it then but maybe a comment about why it is different.

andybalaam (approved):
Looks good, thanks!

PR #326, opened 2022-11-21
Hopefully fix https://github.com/vector-im/element-web/issues/23649

andybalaam:
by handling insertCompositionText as if it were insertText.


florianduros (approved):
Look goods, thanks

PR #329, opened 2022-11-23
implement ctrl/cmd + backspace/delete

alunturner:
First stab at some rust code. Diff is approximately half tests, half code.

Quite a few questions to address, specifically:

- use of enums, is that a sensible approach? where they should actually live in the codebase if it is the right approach?
    - _yes, approach is fine, can stay in that file if not used elsewhere, but names can change to be something more descriptive_
- should the `get_end_index_of_run` method be moved outside the `impl` block? I think it probably should, not sure if it stays in the same file or not
    - _if it can move outside, move it outside and use it as a util_ 
- finally I think it could do with a refactor due to the repetition between `delete_word` and `backspace_word`, this is pretty clearly required as it stands **but** there may also be some value in keeping them separate in case we want to change the behaviour of only one of the methods 
    - _hold off on the refactor to make a single remove word method, need to sort dom traversal and codepoint navigation first_


alunturner:
Superseded by different branch implementing this functionality

PR #330, opened 2022-11-23
Enable DomIterator on any subtree

aringenbach:
Provides a way to start a DomIterator from a DomNode rather than the entire Dom. Result will include all the nodes of the subtree at its location, including itself.


jmartinesp (approved):
LGTM! I guess it also works with `.rev()`?

EDIT: it seems like that needs us to implement `DoubleEndedIterator` too, never mind.

PR #331, opened 2022-11-23
Test: Verifying  the Link Action is in reversed state when the cursors is at the end of a link.

Velin92:


crates/wysiwyg/src/tests/test_menu_state.rs
```diff
@@ -164,6 +164,14 @@ fn selecting_after_a_line_break_inside_formatting_nodes_reversed_actions() {
     assert!(model.action_is_enabled(ComposerAction::StrikeThrough));
 }
 
+#[test]
+fn cursor_at_the_end_of_link_reverses_the_link_action() {
+    let model = cm("non_link_text <a href=\"https://element.io\">link_text|</a> non_link_text");
+    assert!(model.action_is_reversed(ComposerAction::Link));
+    let model2 = cm("non_link_text <a href=\"https://element.io\">link_text</a>| non_link_text");
```
andybalaam, line 171: No need for the second case - it is identical to the first.

andybalaam (approved):
Looks good with small change.

PR #334, opened 2022-11-24
Adding text when cursor is at the end of the link, always places it outside of it

Velin92:
Suggested tests:

- [x] - selection starting before a link and ending afterwards
- [x] - selection starting in one link and ending in another
- [x] - selection starting in a link inside a list item
- [x] - selection starting inside a bold tag inside an italic tag inside a link
- [x] - as above but cursor at the end of all 3

### DIFF
--- a/crates/wysiwyg/src/composer_model/replace_text.rs
+++ b/crates/wysiwyg/src/composer_model/replace_text.rs
@@ -168,7 +168,9 @@ where
 
             if range.is_empty() {
                 if !new_text.is_empty() {
-                    self.state.dom.append_child(DomNode::new_text(new_text));
+                    self.state
+                        .dom
+                        .append_at_end_of_document(DomNode::new_text(new_text));
                 }
                 start = 0;
             } else if let Some(starting_link_handle) = starting_link_handle {
@@ -222,7 +224,8 @@ where
         let action_list = self.replace_in_text_nodes(range.clone(), new_text);
 
         let (to_add, to_delete, _) = action_list.grouped();
-        let to_delete = to_delete.into_iter().map(|a| a.handle).collect();
+        let to_delete: Vec<DomHandle> =
+            to_delete.into_iter().map(|a| a.handle).collect();
 
         // We only add nodes in one special case: when the selection ends at
         // a BR tag. In that case, the only nodes that might be deleted are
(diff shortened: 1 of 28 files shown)


crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -176,6 +188,54 @@ where
...
+            sibling_text_node.set_data(data);
+        } else if !new_text.is_empty() {
+            let new_child = DomNode::new_text(new_text);
+            let parent = self.state.dom.parent_mut(&node_handle);
+            let index = node_handle.index_in_parent() + 1;
+            parent.insert_child(index, new_child);
+        }
+    }
+
+    fn first_sibling_text_node_mut(
```
jmartinesp, line 210: `fn find_next_text_node_mut`?

crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -176,6 +188,54 @@ where
...
+                Some(sibling_text_node)
+            } else {
+                None
+            }
+        } else {
+            None
+        }
+    }
+
+    fn first_link_node_handle(&self, range: &Range) -> Option<DomHandle> {
```
jmartinesp, line 229: `fn find_link_node_to_shrink_at_end`? I can't find a good name for this function. @andybalaam , any ideas?

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -286,6 +290,13 @@ where
         }
     }
 
+    pub(crate) fn is_link_node(&self) -> bool {
```
jmartinesp, line 293: I'm pretty sure this could be just a `matches!(self.kind, ContainerNodeKind::Link(_))` or something similar.

jmartinesp (commented):
Added some comments, mainly about naming functions.

crates/wysiwyg/src/composer_model/replace_text.rs
Velin92: maybe `find_next_sibling_text_node_mut` I feel that not specifying sibling, makes that next mistakable for maybe the next node in the hierarchy or so on...

jmartinesp: Yes, that makes more sense.

jmartinesp (approved):
LGTM!

PR #337, opened 2022-11-25
Prevent inline formatting and links when inline code is selected

jmartinesp:


crates/wysiwyg/src/composer_model/menu_state.rs
```diff
@@ -194,6 +194,25 @@ where
...
         }
+        let contains_inline_code_node = locations
+            .iter()
+            .find(|l| {
+                self.state
+                    .dom
+                    .lookup_node(&l.node_handle)
+                    .is_formatting_node_of_type(&InlineFormatType::InlineCode)
+            })
+            .is_some();
```
aringenbach, line 205: Don't know if we are already doing the same kind of check elsewhere but I guess we definitely want some helper func `contains_node_of_type()` for ranges/locations (can be done later though)

aringenbach (approved):
LGTM

PR #338, opened 2022-11-25
Add `kind` property to DomLocations

jmartinesp:
This should save us from doing unnecessary calls to `self.state.dom.lookup_node(...)` to just check a location's node kind.

### DIFF
--- a/crates/wysiwyg/src/dom/range.rs
+++ b/crates/wysiwyg/src/dom/range.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::dom::dom_handle::DomHandle;
+use crate::dom::nodes::dom_node::DomNodeKind;
 use std::cmp::Ordering;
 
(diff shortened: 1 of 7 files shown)


crates/wysiwyg/src/dom/range.rs
```diff
@@ -51,6 +52,9 @@ pub struct DomLocation {
     /// True if this is a node which is not a container i.e. a text node or
     /// a text-like node like a line break.
     pub is_leaf: bool,
```
aringenbach, line 54: I'm guessing this can be computed from `DomNodeKind` now ?

aringenbach (approved):
LGTM

crates/wysiwyg/src/dom/range.rs
jmartinesp: Yes, nice catch!

PR #339, opened 2022-11-25
Fix replacing a link with text when the selection begins at the start of the link

jmartinesp:
This is actually a not-so-great workaround, so I'm not even sure if we *should* actually use it.

Suggestions to improve it are **truly** welcome.

### DIFF
--- a/crates/wysiwyg/src/composer_model/join_nodes.rs
+++ b/crates/wysiwyg/src/composer_model/join_nodes.rs
@@ -351,7 +351,9 @@ where
         (ret, moved_handles)
     }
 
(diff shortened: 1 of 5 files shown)


jmartinesp:
It seems like once combined with the new text replacement behaviour for Links, this makes the code crash in some cases, such as:

```
let mut model =
        cm("test {<a href=\"https://element.io\">test_link_1</a> <a href=\"https://matrix.org\">test_link_2}|</a> test");
    model.replace_text(utf16("added_text")); // crash on joining nodes code
```
I tried adding more cases, making sure we haven't modified the leaves in the range before applying the `join_nodes` post-processing, etc., but every new change broke something else, so I think it would be better to close this.

PR #343, opened 2022-11-28
Rework list_item slicing with clone & do_replace_text_in

aringenbach:
List item slicing now works with 3 simple steps

1. Clone the list_item and insert the clone afterwards
2. Remove text from slice location to the end of the first list item
3. Replace text from the start of the second list item to the location matching the slice with ZWSP

Makes the list item slicing work with any type of formatting.
Adds some tests for edge cases with `do_replace_text_in` that don't work yet.

### DIFF
--- a/crates/wysiwyg/src/composer_model/replace_text.rs
+++ b/crates/wysiwyg/src/composer_model/replace_text.rs
@@ -76,10 +76,7 @@ where
                 let list_item_end_offset = range
                     .locations
                     .into_iter()
-                    .filter(|loc| {
-                        self.state.dom.lookup_node(&loc.node_handle).kind()
-                            == ListItem
-                    })
+                    .filter(|loc| loc.kind == ListItem)
                     .next()
                     .unwrap()
                     .end_offset;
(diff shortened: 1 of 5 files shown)


crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -67,16 +67,26 @@ where
...
+                location.position + location.start_offset;
             let handle = &location.node_handle;
             let parent_list_item_handle =
                 self.state.dom.find_parent_list_item_or_self(handle);
             if let Some(parent_list_item_handle) = parent_list_item_handle {
+                let list_item_end_offset = range
+                    .locations
+                    .into_iter()
+                    .filter(|loc| {
+                        self.state.dom.lookup_node(&loc.node_handle).kind()
```
jmartinesp, line 80: I think you can just use `loc.kind() == DomNodeKind::ListItem` or `matches!(loc.kind(), ...)` for this now.

jmartinesp (approved):
LGTM, just a possible improvement. 

To be honest, when I first saw this approach it seemed like we were doing the slicing in a very crude and not efficient way, but with the "right" approach we'd still have to duplicate everything inside the list items, so it's actually quite clever!

aringenbach:
Yeah actually I was trying to do the "right way" (we can see it a bit in commit history), but still had more non-working edge cases than this solution. We could still go back to it when there is more structure to the Rust base code to make it easier anyway.

PR #344, opened 2022-11-29
Create sub-iterators for Dom, forward and backwards

jmartinesp:
Draft PR for the DOM sub-iterators. Also adds `Dom.next_node(&DomHandle)` and `Dom.prev_node(&DomHandle)` to get the next and previous nodes in the DOM.


aringenbach (approved):
Code LGTM, pretty sure the best way to be sure it's always reliable is to merge this and start using it everywhere 🙂

PR #345, opened 2022-11-29
iOS: Updating max compressed height now triggers a recalculation of compressed height

Velin92:


andybalaam (approved):
Looks good, probably :-)

PR #347, opened 2022-11-29
Use make setup to avoid duplications on uniffi version in mobile workflows

aringenbach:
Removes some of the duplications on uniffi-version


jonnyandrew (approved):
Nice improvement!

PR #348, opened 2022-11-29
Code blocks

jmartinesp:
Adds code blocks (`<pre>`) support for the text editor. It currently has the following issues:
1. When a code block covers a list item, it turns it into a simple text node and adds a new line after it. This new line is always added, even when not needed.
2. At the moment it's not possible to add only part of a list to a code block: if several ListItem nodes are selected, the whole list will get into the code block.
3. The HTML parser parses `<pre>\nSomething</pre>` as `<pre>Something</pre>` since when using code blocks in HTML, the first line can be used to just give some structure to the HTML. This could be a problem if we create an empty code block and add a new line in the clients, it could either create an index mismatch or just don't process the new line at all.

I'm thinking about possible solutions to these 3 issues, but I'd rather first merge the current version and then attempt to fix them one by one.

Also, the `split_sub_tree` and `slice_before/slice_after` implementations are kind of similar, so it might be a good idea to try to get them merged into a single method in the future.

### DIFF
--- a/crates/wysiwyg/src/composer_model/code_block.rs
+++ b/crates/wysiwyg/src/composer_model/code_block.rs
@@ -43,57 +43,96 @@ where
         let Some((parent_handle, idx_start, idx_end)) = self.find_nodes_to_wrap_in_block(s, e) else {
             return ComposerUpdate::keep();
         };
-        let mut dom = &mut self.state.dom;
-        let mut leaves_to_add: Vec<DomNode<S>> = Vec::new();
+        let mut code_block: DomNode<S> = DomNode::new_code_block(Vec::new());
+        code_block.set_handle(DomHandle::root());
         let mut nodes_visited = HashSet::new();
         let start_handle = parent_handle.child_handle(idx_start);
         let up_to_handle = parent_handle.child_handle(idx_end + 1);
-        let iter = dom.iter_from_handle(&start_handle);
-        for node in iter {
-            if node.handle() >= up_to_handle {
-                break;
+        let ancestor_to_split = self.find_ancestor_to_split(&start_handle);
+        let start_depth = ancestor_to_split.raw().len();
+        let mut path: Vec<usize> = Vec::new();
+        for depth in start_depth..start_handle.raw().len() {
+            let mut cur_child = &mut code_block;
+            let mut path_to_traverse = path.clone();
+            for i in &path_to_traverse {
+                if let DomNode::Container(c) = cur_child {
+                    cur_child = c.get_child_mut(*i).unwrap();
+                }
             }
-            match node.kind() {
-                Text | Formatting(_) | Link => {
-                    nodes_visited.insert(node.handle());
-                    if nodes_visited.contains(&node.handle().parent_handle()) {
-                        continue;
+
+            let DomNode::Container(cur_container) = cur_child else { break; };
+            let path_len = parent_handle.raw().len();
+            if depth == path_len {
+                let iter = self.state.dom.iter_from_handle(&start_handle);
+                for node in iter {
+                    if node.handle() == up_to_handle {
+                        break;
                     }
-                    leaves_to_add.push(node.clone());
-                }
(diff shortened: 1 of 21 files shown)


crates/wysiwyg/src/composer_model/code_block.rs
```diff
@@ -0,0 +1,523 @@
...
+                        continue;
+                    }
+                    leaves_to_add.push(node.clone());
+                }
+                LineBreak => {
+                    leaves_to_add.push(DomNode::new_text("\n".into()));
+                }
+                List | ListItem => {
+                    let mut needs_to_add_line_break =
+                        node.handle().index_in_parent() > 0;
```
jonnyandrew, line 68: Is it right that the code block is coupled so tightly to lists?

I'm wondering if we can reduce code block's knowledge of the other node types so it doesn't need to be updated each time we add or change a node type? If for example, all nodes types implement a `to_code()` method, then code block can simply call `node.to_code()` on any compatible node type and concentrate on the wrapping / unwrapping the result.

crates/wysiwyg/src/composer_model/code_block.rs
```diff
@@ -0,0 +1,523 @@
...
+                .locations
+                .iter()
+                .filter(|l| l.kind == ListItem)
+                .count()
+                > 1;
+            for node in rev_iter {
+                if !node.is_block_node()
+                    && node.kind() != LineBreak
+                    && (node.kind() != ListItem
+                        || selection_contains_several_list_items)
```
jonnyandrew, line 183: Similarly, there is a lot of code that is tightly coupled to lists here. Should code block need to know how lists work?

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -388,6 +388,83 @@ where
             panic!("Parent node was not a container!");
         }
     }
+
+    /// Checks if the passed [handle] exists in the DOM.
+    pub fn exists(&self, handle: &DomHandle) -> bool {
```
jonnyandrew, line 393: I think this checks if the DOM contains an element. Should this method then be called `contains`?

jonnyandrew (commented):
Nice! I haven't tested but I had a quick look at the code.

crates/wysiwyg/src/composer_model/code_block.rs
jmartinesp: I added a `ComposerModel.format_node_for_code_block` function, but we still need to check the kind of the DomNode to understand where we need to insert new line breaks.

crates/wysiwyg/src/composer_model/code_block.rs
jmartinesp: It needs for now, but when I fix the issue with whole lists being included into code blocks instead of only some of their list items I think I might be able to avoid these.

crates/wysiwyg/src/composer_model/menu_state.rs
```diff
@@ -214,7 +207,28 @@ where
...
+            )
+        })
+        .is_some()
+}
+
+fn contains_code_block(locations: &Vec<DomLocation>) -> bool {
+    locations
+        .iter()
+        .find(|l| l.kind == DomNodeKind::CodeBlock)
+        .is_some()
```
aringenbach, line 233: There is a `any` operator that is the equivalent of `find` + `is_some`

(we can clean this up in a separate PR because I'm sure we have done this in several other places 🙂 )

crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -227,7 +229,7 @@ where
                 )
             {
                 // FIXME: need to join other types of node as well
```
aringenbach, line 231: This function do join other nodes than text now

aringenbach (approved):
LGTM, let's indeed merge this big change, it will be easier to iterate over it 👍

PR #349, opened 2022-11-30
Provide a method to assert the Dom is in a known state

andybalaam:


crates/wysiwyg/src/dom/iter.rs
```diff
@@ -54,6 +63,18 @@ where
...
+    /// order
+    pub fn iter_containers(&self) -> impl Iterator<Item = &ContainerNode<S>> {
+        self.iter().filter_map(|node| {
+            if let DomNode::Container(c) = node {
+                Some(c)
+            } else {
+                None
+            }
+        })
+    }
```
aringenbach, line 77:
```suggestion
    pub fn iter_containers(&self) -> impl Iterator<Item = &ContainerNode<S>> {
        self.iter().filter_map(DomNode::as_container)
    }
```

aringenbach (approved):
Looks like a pretty good start for fixing our consistency issues 🙏

PR #355, opened 2022-12-01
Get Link Action function

Velin92:
A function that will return the current action that can be performed on a link


Velin92:
> I'd actually add a few more tests, for existing links:
> 
> * One where the cursor is inside the link or the selection only partially covers it.
> * Another for selecting some external nodes and the link (totally or partially).
> * Another for selecting several link nodes, making sure the 1st one is returned.

just added them right now!

jmartinesp (approved):
LGTM!

PR #363, opened 2022-12-01
Set Link With Text

Velin92:
A function that should be used only when there is no highlighted text, that allows the client to add text and then a link to it.

### DIFF
--- a/bindings/wysiwyg-wasm/src/lib.rs
+++ b/bindings/wysiwyg-wasm/src/lib.rs
@@ -65,7 +65,7 @@ trait IntoFfi {
 impl IntoFfi for &HashMap<wysiwyg::ComposerAction, wysiwyg::ActionState> {
     fn into_ffi(self) -> js_sys::Map {
         let ret = js_sys::Map::new();
-        for (k, v) in self.into_iter() {
+        for (k, v) in self.iter() {
             ret.set(&k.as_ref().into(), &v.as_ref().into());
         }
         ret
@@ -75,6 +75,12 @@ impl IntoFfi for &HashMap<wysiwyg::ComposerAction, wysiwyg::ActionState> {
 #[wasm_bindgen]
 pub struct ComposerModel {
     inner: wysiwyg::ComposerModel<Utf16String>,
+}
+
+impl Default for ComposerModel {
+    fn default() -> Self {
+        Self::new()
+    }
 }
 
 #[wasm_bindgen]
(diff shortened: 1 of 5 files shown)


andybalaam (approved):
Looks great, nice work.

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -39,15 +39,27 @@ where
...
         }
         self.push_state_to_history();
+        self.do_replace_text(text.clone());
+        e += text.len();
+        let range = self.state.dom.find_range(s, e);
+        self.set_link_range(range, link)
+    }
 
+    pub fn set_link(&mut self, link: S) -> ComposerUpdate<S> {
+        // push_state_to_history is after this check:
```
jmartinesp, line 60: This probably no longer applies?

bindings/wysiwyg-wasm/src/lib.rs
```diff
@@ -77,6 +77,12 @@ pub struct ComposerModel {
     inner: wysiwyg::ComposerModel<Utf16String>,
 }
 
+impl Default for ComposerModel {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
jmartinesp, line 85: I don't think we need these default methods? In fact, I think we should never have an empty `ComposerModel`.

jmartinesp (approved):
LGTM, just a couple of comments.

crates/wysiwyg/src/composer_model/hyperlinks.rs
Velin92: whoops need to remove this comment

bindings/wysiwyg-wasm/src/lib.rs
Velin92: I know... the thing is that clippy complains that if you have an empty new implementation like we do, we either remove that implementation or we create also the Default impl for it.
Can either remove both this and the `new()` if we never use it or keep this

jmartinesp: I don't think we enforce clippy warnings on the repo at the moment, but maybe then it's best to rename the `new` function to `default`, as that seems to be the Rust way. Is that right @andybalaam ?

Velin92: you should use default like this
`let mut model: ComposerModel = Default::default()`
in place of using the new, of course a struct contains sub structs that implement default, you can also use `#derive(Default)`

Velin92: You know what? Let's keep the warning for now, I'll remove the default implementation.
The cleanest thing to do would be to go through all our structs, make them implement default through a derive, so that we can actually have more complex structs derive from default and use it in place of the empty `new()`

But yeah all this work is worth doing into another PR

andybalaam: Yeah, implementing or deriving `Default` is good when there is a sensible default/empty value for an object.

No, we don't enforce clippy and we should!

PR #364, opened 2022-12-02
Set default toolchain via rust-toolchain.toml

jplatte:
This allows users to have their own overrides while defaulting to the specified toolchain. The same config file could also be users to have rustup automatically download standard libraries for specific targets (e.g. wasm) automatically, see the docs for details:

https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,6 +90,5 @@ coverage:
 	cargo llvm-cov --open
 
 setup:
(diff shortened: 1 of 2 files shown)


andybalaam (approved):
Great, thank you!

PR #365, opened 2022-12-02
[Android] Fix crashes in ElementX on config changes

jmartinesp:
There are several crashes on Element X because the lazy ViewModel is instantiated when it's still not attached to the parent view. This happens both in `AndroidView.factory` and the `EditText.onRestoreInstanceState()` method when it's inside an `AndroidView` in Compose component.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/ViewLazyViewModelExtension.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/ViewLazyViewModelExtension.kt
@@ -12,17 +12,14 @@ import androidx.lifecycle.findViewTreeViewModelStoreOwner
 
 @MainThread
 inline fun <reified VM : ViewModel> View.viewModel(
-    useViewTreeOwner: Boolean = false,
     noinline viewModelInitializer: (() -> VM)? = null,
 ): Lazy<VM> {
     return ViewModelLazy(
         viewModelClass = VM::class,
         storeProducer = {
-            if (useViewTreeOwner) {
-                findViewTreeViewModelStoreOwner()!!.viewModelStore
-            } else {
-                (getBaseContext(context) as ViewModelStoreOwner).viewModelStore
-            }
+            // We no longer use `findViewTreeViewModelStoreOwner` here because it can cause crashes
+            // in Compose AndroidViews. See https://github.com/matrix-org/matrix-rich-text-editor/pull/365.
+            context.getViewModelStoreOwner().viewModelStore
         },
         factoryProducer = {
             object : ViewModelProvider.Factory {
@@ -36,13 +33,13 @@ inline fun <reified VM : ViewModel> View.viewModel(
     )
 }
 
-fun getBaseContext(context: Context): Context {
-    var currentContext = context
+fun Context.getViewModelStoreOwner(): ViewModelStoreOwner {
+    var currentContext = this
     while (currentContext !is ViewModelStoreOwner) {
         if (currentContext is ContextWrapper) {
             currentContext = currentContext.baseContext
         } else {
-            error("There is not base context that is a ViewModelStoreOwner")
+            error("There is no base context that is a ViewModelStoreOwner")
         }
     }
     return currentContext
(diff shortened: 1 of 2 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/ViewLazyViewModelExtension.kt
```diff
@@ -1,19 +1,29 @@
...
+    noinline viewModelInitializer: (() -> VM)? = null,
 ): Lazy<VM> {
     return ViewModelLazy(
         viewModelClass = VM::class,
-        storeProducer = { findViewTreeViewModelStoreOwner()!!.viewModelStore },
+        storeProducer = {
+            if (useViewTreeOwner) {
+                findViewTreeViewModelStoreOwner()!!.viewModelStore
+            } else {
+                (getBaseContext(context) as ViewModelStoreOwner).viewModelStore
```
jmartinesp, line 24: Add a new (default) way of getting the `ViewModel` that won't crash when using Compose and having config changes (like screen rotation).

@jonnyandrew I see few other alternatives, but if you do find one please modify this code. Also, feel free to remove the old code if you think it's not worth keeping it around.

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/ViewLazyViewModelExtension.kt
```diff
@@ -22,6 +32,18 @@ inline fun <reified VM : ViewModel> View.viewModel(
                         ?: modelClass.newInstance()
                 }
             }
-        }
+        },
     )
 }
+
+fun getBaseContext(context: Context): Context {
```
bmarty, line 39: Maybe return a `ViewModelStoreOwner` and change to `Context.getViewModelStoreOwner()`

```
    fun Context.getViewModelStoreOwner(): ViewModelStoreOwner {
        var currentContext = this
        while (currentContext !is ViewModelStoreOwner) {
            if (currentContext is ContextWrapper) {
                currentContext = currentContext.baseContext
            } else {
                error("There is no base context that is a ViewModelStoreOwner")
            }
        }
        return currentContext
    }
```

jonnyandrew (approved):

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/ViewLazyViewModelExtension.kt
```diff
@@ -22,6 +32,18 @@ inline fun <reified VM : ViewModel> View.viewModel(
...
+    var currentContext = context
+    while (currentContext !is ViewModelStoreOwner) {
+        if (currentContext is ContextWrapper) {
+            currentContext = currentContext.baseContext
+        } else {
+            error("There is not base context that is a ViewModelStoreOwner")
```
bmarty, line 45:
```suggestion
            error("There is no base context that is a ViewModelStoreOwner")
```

bmarty (commented):
Thanks for this change. I have made some minor remarks. I let @jonnyandrew approve the PR.

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/ViewLazyViewModelExtension.kt
```diff
@@ -22,6 +29,18 @@ inline fun <reified VM : ViewModel> View.viewModel(
                         ?: modelClass.newInstance()
                 }
             }
-        }
+        },
     )
 }
+
+fun Context.getViewModelStoreOwner(): ViewModelStoreOwner {
```
bmarty, line 36: (sorry for the late remark). Could be private (or internal), but may be useful for library users, hence this is not the aim of the library to expose this.

jmartinesp: Fixed.

jmartinesp: Actually, I just remembered why it wasn't private/internal. It seems like it can't be when used from an inline function. I'll revert the commit.

bmarty: Strange, but OK

PR #368, opened 2022-12-02
Change Rust panic behaviour from abort to unwind

jonnyandrew:
## What?
Change the behaviour of Rust panics so that, instead of aborting, they unwind.

## Why?
When Rust panics, if it is configured to abort, the host application has no way to continue and must crash.

Changing this behaviour to instead unwind allows Rust panics to be caught at the FFI boundary so that the host application catch and rethrow the error in the host language (`InternalException` in Kotlin for example), and continue without crashing.

Note this imposes a small performance/size impact and it could be worth reverting this once the library is stable.


andybalaam (approved):
Thanks!

PR #371, opened 2022-12-05
Char extension for ZWSP

aringenbach:
* Add extension to character to create ZWSP easily
* Use it for UnicodeString ZWSP string creation
* Replace most isolated usages
* Fix some clippy warnings on `text_node.rs`

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/text_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/text_node.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::char::CharExt;
 use crate::composer_model::delete_text::Direction;
 use crate::composer_model::example_format::SelectionWriter;
 use crate::dom::dom_handle::DomHandle;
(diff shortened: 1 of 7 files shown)


crates/wysiwyg/src/dom/nodes/text_node.rs
```diff
@@ -128,10 +126,7 @@ where
         direction: &Direction,
     ) -> Option<CharType> {
         let char = self.char_at_offset(offset, direction);
-        match char {
-            Some(c) => Some(get_char_type(c)),
-            None => None,
-        }
+        char.map(get_char_type)
```
alunturner, line 129: I thought there must have been a better place to express this when I wrote it

alunturner (approved):
Looks good to me (and neater)

crates/wysiwyg/src/dom/nodes/text_node.rs
aringenbach: @alunturner  You can add `"rust-analyzer.checkOnSave.command": "clippy",` in e.g. VS Code json settings, that's what point me towards this.

PR #372, opened 2022-12-05
Remove list item tag name from constructors

aringenbach:
Removed an old and unnecessary item_name parameter on ListItem constructor


alunturner (approved):
Looks good to me

PR #374, opened 2022-12-05
Fix crash when deleting all text left in a model within a tag

aringenbach:
Crash happened within any type of node (link, formatting, etc)

### DIFF
--- a/crates/wysiwyg/src/dom/dom_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_methods.rs
@@ -193,14 +193,16 @@ where
 
         // Delete the nodes marked for deletion
         let deleted_handles = if !to_delete.is_empty() {
-            self.delete_nodes(to_delete.clone())
+            self.delete_nodes(to_delete)
         } else {
             Vec::new()
         };
 
-        // If our range covered multiple text-like nodes, join together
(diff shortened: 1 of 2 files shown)


crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -193,14 +193,16 @@ where
...
         } else {
             Vec::new()
         };
 
-        // If our range covered multiple text-like nodes, join together
-        // the two sides of the range.
-        if range.leaves().count() > 1 {
+        if self.children().is_empty() {
+            // Nothing to join if the deletion left the DOM empty.
+        } else if range.leaves().count() > 1 {
```
Velin92, line 203: This should be a bit easier to read
```
if range.leaves.count() > 1 && !self.children().is_empty
```

Velin92 (approved):
LGTM just one small comment to avoid an empty if statement

crates/wysiwyg/src/dom/dom_methods.rs
aringenbach: Doesn't work because we have to opt out of both the `if` and the `else if` if that statement is true, might be better to extract this to a function though to return more explicitly.

PR #378, opened 2022-12-06
make find range by work symmetrically

alunturner:
Previously, we would return anything left of the cursor, but only containers to the right if at a boundary. Now we return all things on both sides. See below for an illustration.

* `<br />|<br />`
    * before returns single location
    * now returns both locations
*  `<em>something</em>|<em>else</em>`
    * before returns 3 locations, left container, left text node, right container
    * now returns 4 locations, left container, left text node, right container, right text node

### DIFF
--- a/crates/wysiwyg/src/composer_model/replace_text.rs
+++ b/crates/wysiwyg/src/composer_model/replace_text.rs
@@ -87,7 +87,6 @@ where
                     );
                     return self.create_update_replace_all();
                 } else {
-                    println!("working on {:?}", leaf);
                     let current_cursor_global_location =
                         leaf.position + leaf.start_offset;
                     let handle = &leaf.node_handle;
--- a/crates/wysiwyg/src/composer_model/replace_text.rs
+++ b/crates/wysiwyg/src/composer_model/replace_text.rs
@@ -87,6 +87,7 @@ where
                     );
                     return self.create_update_replace_all();
                 } else {
+                    println!("working on {:?}", leaf);
                     let current_cursor_global_location =
                         leaf.position + leaf.start_offset;
                     let handle = &leaf.node_handle;
@@ -121,7 +122,7 @@ where
         &mut self,
         range: Range,
     ) {
-        let leaves: Vec<&DomLocation> = range.leaves().collect();
+        let leaves = range.leaves();
         let empty_text_leaves: Vec<&DomLocation> = leaves
             .into_iter()
             .filter(|l| {
--- a/crates/wysiwyg/src/composer_model/replace_text.rs
+++ b/crates/wysiwyg/src/composer_model/replace_text.rs
@@ -63,50 +63,65 @@ where
         &mut self,
         range: Range,
     ) -> ComposerUpdate<S> {
-        let leaves: Vec<&DomLocation> = range.leaves().collect();
-        if leaves.len() == 1 {
-            let location = leaves[0];
-            let current_cursor_global_location =
-                location.position + location.start_offset;
-            let handle = &location.node_handle;
-            let parent_list_item_handle =
-                self.state.dom.find_parent_list_item_or_self(handle);
-            if let Some(parent_list_item_handle) = parent_list_item_handle {
-                let list_item_end_offset = range
-                    .locations
-                    .into_iter()
-                    .filter(|loc| loc.kind == ListItem)
-                    .next()
-                    .unwrap()
-                    .end_offset;
-                self.do_enter_in_list(
-                    &parent_list_item_handle,
-                    current_cursor_global_location,
-                    list_item_end_offset,
-                )
-            } else {
-                self.do_enter_in_text(handle, location.start_offset)
+        let position = range.start();
+        let leaf_at_cursor: Option<&DomLocation> = range.leaves().find(|loc| {
+            loc.position <= position && position <= loc.position + loc.length
+        });
+
+        match leaf_at_cursor {
+            None => {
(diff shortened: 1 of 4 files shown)


crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -63,50 +63,65 @@ where
...
+                                &handle,
+                                current_cursor_global_location,
+                                list_item_end_offset,
+                            )
+                        }
+                        None => {
+                            self.do_enter_in_text(handle, leaf.start_offset)
+                        }
+                    }
+                }
```
alunturner, line 115: This is the same branching logic as before, but now we need to account for multiple locations so can't use the length check we were previously using

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -312,19 +327,11 @@ mod test {
...
         );
-        assert_eq!(
-            find_pos(&d, &d.document_handle(), 3, 3),
-            found_single_node(DomHandle::from_raw(vec![0]), 0, 3, 3, 3)
-        );
-        assert_eq!(
-            find_pos(&d, &d.document_handle(), 3, 4),
-            found_single_node(DomHandle::from_raw(vec![1]), 3, 0, 1, 3)
-        );
-        // TODO: break up this test and name parts!
```
alunturner, line 323: Tests rearranged to remove this todo and slim this single large test down. Tests are equivalent apart from one new test.

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -345,12 +348,25 @@ mod test {
...
+        let d = dom(&[tn("foo"), tn("bar")]);
+        assert_eq!(
+            find_pos(&d, &d.document_handle(), 3, 3),
+            FindResult::Found(vec![
+                make_single_location(DomHandle::from_raw(vec![0]), 0, 3, 3, 3),
+                make_single_location(DomHandle::from_raw(vec![1]), 3, 0, 0, 3)
+            ])
+        );
+    }
+
```
alunturner, line 362: This is the new test for the boundary condition

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -502,36 +518,6 @@ mod test {
...
-                        start_offset: 8,
-                        end_offset: 8,
-                        position: 0,
-                        length: 30,
-                        kind: DomNodeKind::Formatting(Italic)
-                    },
-                ]
-            }
-        )
+        assert_eq!(range.locations.len(), 4)
```
aringenbach, line 521: Previous code also asserted that the length of the DomLocations created is correct, can we have a test for this ? 🙏

crates/wysiwyg/src/dom/find_range.rs
```diff
@@ -502,36 +518,6 @@ mod test {
...
-                        length: 8,
-                        kind: DomNodeKind::Text
-                    },
-                    DomLocation {
-                        node_handle: DomHandle::from_raw(vec![0, 1]),
-                        start_offset: 0,
-                        end_offset: 0,
-                        position: 8,
-                        length: 21,
-                        kind: DomNodeKind::Formatting(Italic)
```
aringenbach, line 523: If we remove this I believe you ave unused import of InlineFormatType::Italic

crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -63,50 +63,65 @@ where
...
+
+                    match parent_list_item_handle {
+                        Some(handle) => {
+                            let list_item_end_offset = range
+                                .locations
+                                .into_iter()
+                                .filter(|loc| loc.kind == ListItem)
+                                .next()
+                                .unwrap()
+                                .end_offset;
```
aringenbach, line 104:
```suggestion
                            let list_item_end_offset = range
                                .locations
                                .into_iter()
                                .find(|loc| loc.kind == ListItem)
                                .unwrap()
                                .end_offset;
```

crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -63,50 +63,65 @@ where
...
+                self.state.end = self.state.start;
+                return self.create_update_replace_all();
+            }
+            Some(leaf) => {
+                if leaf.length == 0 {
+                    // Special case, there is an empty text node at the cursor position
+                    self.enter_with_zero_length_selection_and_empty_text_nodes(
+                        range,
+                    );
+                    return self.create_update_replace_all();
```
aringenbach, line 88:
```suggestion
                    self.create_update_replace_all()
```

crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -63,50 +63,65 @@ where
...
+        match leaf_at_cursor {
+            None => {
+                // Selection doesn't contain any text like nodes. We can assume it's an empty Dom.
+                self.state
+                    .dom
+                    .document_mut()
+                    .append_child(DomNode::new_line_break());
+                self.state.start += 1;
+                self.state.end = self.state.start;
+                return self.create_update_replace_all();
```
aringenbach, line 80:
```suggestion
                self.create_update_replace_all()
```

aringenbach (changes requested):
Looks good, just a few suggestions to apply :)

alunturner:
> Looks good, just a few suggestions to apply :)

Thanks for the feedback - all those changes now applied

aringenbach (approved):
LGTM

PR #379, opened 2022-12-06
Lookup and join ancestor on join nodes

aringenbach:
* Fixes some delete text cases by looking up ancestors from an entirely removed text node until finding a non-removed ancestor.
* Now joins all compatible nodes in the ancestor and not only text nodes
* Add some `push` functions to append the content of another mutable node into a dom node and ease up joining nodes. (since we can't borrow multiple mutable nodes of the tree by design, it should be rather safe to use because the node to join should removed from the tree first, and will be inserted without any duplication - unless explicitly cloning).
* Add a `can_push` helper to check if a node can safely be removed from the tree in order to be pushed into its (usually) previous neighbour.
* Also, fixes some clippy issues in updated files.

### DIFF
--- a/crates/wysiwyg/src/tests/test_deleting.rs
+++ b/crates/wysiwyg/src/tests/test_deleting.rs
@@ -361,3 +361,22 @@ fn deleting_selection_of_a_container_in_multiple_containers() {
     model.backspace();
     assert_eq!(tx(&model), "<i>| test</i>");
 }
+
+#[test]
+fn deleting_selection_of_a_container_with_text_node_neighbors() {
+    let mut model = cm(
+        "<em>abc<del>{def}|</del>ghi</em>",
+    );
+    model.backspace();
+    assert_eq!(tx(&model), "<em>abc|ghi</em>");
+}
+
+
+#[test]
+fn deleting_selection_of_a_container_with_matching_neighbors() {
+    let mut model = cm(
+        "<em><strong>abc</strong><del>{def}|</del><strong>ghi</strong></em>",
+    );
+    model.backspace();
+    assert_eq!(tx(&model), "<em><strong>abc|ghi</strong></em>");
+}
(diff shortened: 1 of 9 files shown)


crates/wysiwyg/src/tests/test_deleting.rs
```diff
@@ -361,3 +361,19 @@ fn deleting_selection_of_a_container_in_multiple_containers() {
...
+    assert_eq!(tx(&model), "<em>abc|ghi</em>");
+}
+
+#[test]
+fn deleting_selection_of_a_container_with_matching_neighbors() {
+    let mut model = cm(
+        "<em><strong>abc</strong><del>{def}|</del><strong>ghi</strong></em>",
+    );
+    model.backspace();
+    assert_eq!(tx(&model), "<em><strong>abc|ghi</strong></em>");
```
Velin92, line 378: maybe is worth checking the tree to for these two tests?

crates/wysiwyg/src/tests/test_remove_links.rs
```diff
@@ -108,3 +108,11 @@ fn remove_multiple_partially_selected_links_in_different_containers() {
         "<b>test_{link_bold</b> <i>test}|_link_italic</i>"
     );
 }
+
+#[test]
+fn remove_link_between_text_nodes_joins() {
+    let mut model = cm("abc{<a href=\"https://matrix.org\">def</a>}|ghi");
+    model.remove_links();
+    assert_eq!(tx(&model), "abc{def}|ghi");
+    model.state.dom.explicitly_assert_invariants();
```
Velin92, line 117: same as above

Velin92 (approved):
LGTM, two suggestions

crates/wysiwyg/src/tests/test_deleting.rs
aringenbach: I just added `explicitly_assert_invariants()` to all deleting tests that did not have it before and they all pass. Having this test pass and the expected HTML should be pretty much the same as having the tree in the form we expect :)

crates/wysiwyg/src/tests/test_remove_links.rs
aringenbach: Should be fine with the last assert, see comment above ^

PR #380, opened 2022-12-06
Change link interface to use string

jonnyandrew:
## What?

Update the interface for link actions to use `String` instead of `Vec<u16>`.

## Why?

So that the Kotlin/Swift bindings generated for this class use the native string types and do not need conversion.

After this change, the Kotlin binding looks like this:

```
sealed class LinkAction {
...
    
    data class Edit(
        val `link`: String
    ) : LinkAction()
}


Velin92 (approved):
LGTM

PR #381, opened 2022-12-06
Add link APIs for Android

jonnyandrew:
Add public APIs to the Android library for the following:
- Get the current 'link action' (so that the caller knows what is possible when calling the new link APIs on the editor)
- Set a link on the current selection
- Insert new text with a link


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -252,8 +253,46 @@ class EditorEditText : TextInputEditText {
...
+     * @see [setLink]
+     */
+    fun removeLink() = setLink(null)
+
+    /**
+     * Create text with a link.
+     *
+     * @param link The link to set (or null to remove).
+     */
+    fun insertLink(link: String, text: String) {
```
jonnyandrew, line 294: On the naming, I thought `insertLink()` could make a bit more sense for the public API than `setLinkWithText()` (to make it clearer that this will insert new text). @Velin92 what do you think? And should we consider it for the internal interfaces too?

Velin92: on the iOS side I actually named it `createLink(urlString: String, text: String)`
Since the setLink is also usable for both creating a link from scratch and for the edit, while the first is pretty much just for creation/insertion, so I guess is okay!

aringenbach (approved):
LGTM

One small functional question, should we be able to change both the link text and the url when editing (with the same UI as creating from zero-length selection creation, pre-filled with current text)

jonnyandrew:
> should we be able to change both the link text and the url when editing (with the same UI as creating from zero-length selection creation, pre-filled with current text) ?

@aringenbach yes that is the end goal according to our designs, but the composer model doesn't yet support this.

It looks like this feature will be implemented by [PSU-1002](https://element-io.atlassian.net/browse/PSU-1002) (at least for unformatted text).

PR #383, opened 2022-12-07
Remove word functionality

alunturner:
This PR adds functionality for cmd/opt backspace/delete. This _does not_ include full list behaviour yet.

The diff looks big, but 450 of those lines are tests.

Changes

- add 8 methods to `delete_text.rs` to apply the functionality
- add extensive tests for this functionality
- add starter tests for list behaviour
- remove unused utils committed last week for this work
- amends the `CharType` enum to remove `LineBreak` and add `ZWSP` types as the first was not required and the second will be important for list functionality


crates/wysiwyg/src/composer_model/delete_text.rs
```diff
@@ -87,6 +114,213 @@ where
...
+                        None => ComposerUpdate::keep(),
+                        Some(details) => {
+                            let (_location, _) = details;
+                            self.remove_word(
+                                // use the original start type from the previous call
+                                start_type, direction, _location,
+                            )
+                        }
+                    }
+                }
```
alunturner, line 239: I think that the logic here can be improved due to the common code between the blocks, but I can't figure out the borrowing

crates/wysiwyg/src/composer_model/delete_text.rs
```diff
@@ -87,6 +114,209 @@ where
...
+        match selected_location {
+            None => None,
+            Some(location) => {
+                let char_type =
+                    self.get_char_type_from_location(&location, direction);
+
+                match char_type {
+                    None => None,
+                    Some(char_type) => Some((location, char_type)),
+                }
```
aringenbach, line 292:
```suggestion
                char_type.map(|char_type| (location, char_type))
```

crates/wysiwyg/src/composer_model/delete_text.rs
```diff
@@ -87,6 +114,209 @@ where
...
+    fn get_char_type_from_location(
+        &self,
+        location: &DomLocation,
+        direction: &Direction,
+    ) -> Option<CharType> {
+        let node = self.state.dom.lookup_node(&location.node_handle);
+        match node {
+            DomNode::Container(_) => {
+                // we should never get a container type
+                panic!("hit container in get_details_from_range_and_direction")
```
aringenbach, line 308:
```suggestion
                panic!("hit container in get_char_type_from_location")
```

aringenbach (commented):
Good to see that much tests 🙏 

Maybe you can try adding `model.state.dom.explicitly_assert_invariants();` to the end of your tests to detect if we have some join nodes mechanic to trigger in these cases (which I guess we probably do need somewhere)

aringenbach (approved):
LGTM

PR #384, opened 2022-12-07
Extended selection/range for DOM

aringenbach:
Add some helper to the find_range mechanism that allows iterating towards the next linebreak/structure node/list item in order to build an extended range.

### DIFF
--- a/crates/wysiwyg/src/dom/find_extended_range.rs
+++ b/crates/wysiwyg/src/dom/find_extended_range.rs
@@ -1,0 +1,144 @@
+// Copyright 2022 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/dom/find_extended_range.rs
```diff
@@ -0,0 +1,221 @@
...
+        for node in iter {
+            if node.is_text_node() {
+                offset += node.text_len();
+            } else if node.is_line_break() {
+                offset += node.text_len();
+                break;
+            } else if node.is_block_node() || node.is_list_item() {
+                break;
+            }
+        }
```
alunturner, line 77: Maybe personal preference but I wonder if something like the following would be easier to read
```suggestion
        for node in iter {
            if node.is_text_node() || node.is_line_break() {
                offset += node.text_len();
            } 
            
            if node.is_line_break() || node.is_block_node() || node.is_list_item() {
                break;
            }
        }
```

crates/wysiwyg/src/dom/find_extended_range.rs
```diff
@@ -0,0 +1,221 @@
...
+                if self.is_last_child_of_list(&node.handle()) {
+                    break;
+                }
+            } else if node.is_line_break() {
+                offset += node.text_len();
+                break;
+            } else if node.is_block_node() || node.is_list_item() {
+                break;
+            }
+        }
```
alunturner, line 109: Similar to above, I think legibility is improved by separating the offset increment and the break logic

PR #388, opened 2022-12-09
Add dom node slicing tools & add leading ZWSP from container level

aringenbach:
* Add some tools that allows slicing nodes at given offset
* Add a function to insert leading ZWSP from container level (effectively inserting in the first found text node)
* Add some tests to DomNode, ContainerNode & TextNode

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -317,6 +317,17 @@ where
         Some(link)
     }
 
+    /// Add a leading ZWSP char to this container.
+    /// Returns false if no updates was done.
+    /// e.g. first text-like node is a line break
+    /// or a text node that already starts with a ZWSP.
+    pub fn add_leading_zwsp(&mut self) -> bool {
+        let Some(first_child) = self.children.get_mut(0) else {
+            return false;
+        };
+        first_child.add_leading_zwsp()
+    }
+
     /// Push content of the given container node into self. Panics
     /// if given container node is not of the same kind.
     pub(crate) fn push(&mut self, other_node: &mut ContainerNode<S>) {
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -317,6 +319,29 @@ where
         Some(link)
     }
 
+    /// Creates a container with the same kind & attributes
+    /// as self, with given children and an unset handle.
+    fn clone_with_children(&self, children: Vec<DomNode<S>>) -> Self {
```
jmartinesp, line 324: I built almost exactly the same function for code blocks 😅 .

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -334,6 +359,108 @@ where
             self.append_child(child);
         }
     }
+
+    /// Slice this container after given position.
+    /// Returns a new container of the same kind with the
+    /// removed content, with both nodes keeping
+    /// expected hierarchy.
+    pub fn slice_after(&mut self, offset: usize) -> ContainerNode<S> {
```
jmartinesp, line 367: And also a very similar function to this one, to split the Dom into a sub-tree that can be inserted later. I guess we'll have to try merge them somehow later.

Also, could we change `offset` for `index`? Offset is usually used in the project for char offset and it can be a bit misleading.

jmartinesp (approved):
LGTM! Just a comment about naming.

crates/wysiwyg/src/dom/nodes/container_node.rs
aringenbach: I'm changing for `position` because this is actually a relative position inside the node more than an index or an offset, actually even my documentation says so 😅

PR #395, opened 2022-12-13
Add list_type to `ContainerNodeKind::List`

aringenbach:
* Ease up comparing ContainerNodeKind for lists
* Removed some unnecessary cloning
* Removed a few clippy issues
* Joining subsequent lists of same kind is allowed
* Effectively testing joining lists will be added by upcoming PRs

### DIFF
--- a/crates/wysiwyg/src/dom/range.rs
+++ b/crates/wysiwyg/src/dom/range.rs
@@ -357,7 +357,7 @@ mod test {
         let r = range_of(
             "\
             <ul><li>{a</li><li>b</li></ul>\
-            <ul><li>c</li><li>d</li><li>e</li></ul>fgh}|",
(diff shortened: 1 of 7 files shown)


crates/wysiwyg/src/dom/range.rs
```diff
@@ -357,7 +357,7 @@ mod test {
         let r = range_of(
             "\
             <ul><li>{a</li><li>b</li></ul>\
-            <ul><li>c</li><li>d</li><li>e</li></ul>fgh}|",
```
aringenbach, line 360: Small explanation on this change, the join nodes mechanism is triggered by the parser, which means we do join lists of same type here, but this is not what we want to test here, so I changed to non-matching list types to fix the test (instead of recalculating the handles that are transformed by joining nodes).

jmartinesp (approved):
LGTM!

PR #396, opened 2022-12-13
Add `DomLocationPosition` helper

aringenbach:
* Add a helper that computes a DomLocation relative position towards its Range
* Also replaced two filters with it that now don't need to compute the `range.end()` with an iterator anymore.
* Fixed some clippy issues within `range.rs`


jmartinesp (approved):
LGTM!

PR #398, opened 2022-12-13
Fix issues with line breaks at the start and end of code blocks

jmartinesp:
* Removes the last line break in a `<pre>` tag if not needed.
* Allows `ComposerModel` to properly set a selection in code blocks starting with a line break.

### DIFF
--- a/crates/wysiwyg/src/tests/test_paragraphs.rs
+++ b/crates/wysiwyg/src/tests/test_paragraphs.rs
@@ -196,7 +196,7 @@ fn enter_in_code_block_at_start_adds_the_line_break() {
 #[test]
 fn enter_in_code_block_at_start_with_previous_line_break_moves_it_outside_the_code_block(
 ) {
-    // The initial line break will be removed, so it's the as having a single line break at the start
+    // The initial line break will be removed, so it's the same as having a single line break at the start
     let mut model = cm("<pre>\n\n|Test</pre>");
     model.enter();
     assert_eq!(tx(&model), "<br />|<pre>Test</pre>")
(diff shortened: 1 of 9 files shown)


crates/wysiwyg/src/tests/test_paragraphs.rs
```diff
@@ -187,13 +187,29 @@ fn enter_in_code_block_in_text_node_adds_line_break_as_text() {
...
+) {
+    // The initial line break will be removed, so it's the as having a single line break at the start
+    let mut model = cm("<pre>\n\n|Test</pre>");
     model.enter();
     assert_eq!(tx(&model), "<br />|<pre>Test</pre>")
 }
 
+#[test]
+fn enter_in_code_block_at_start_with_a_line_break_after_it_adds_another_one() {
+    // The initial line break will be removed, so it's the as having a single line break at the start
```
aringenbach, line 207:
```suggestion
    // The initial line break will be removed, so it's the same as having a single line break at the start
```

crates/wysiwyg/src/tests/test_paragraphs.rs
```diff
@@ -187,13 +187,29 @@ fn enter_in_code_block_in_text_node_adds_line_break_as_text() {
...
+fn enter_in_code_block_at_start_adds_the_line_break() {
+    let mut model = cm("<pre>|Test</pre>");
+    model.enter();
+    assert_eq!(tx(&model), "<pre>\n|Test</pre>")
+}
+
+#[test]
+fn enter_in_code_block_at_start_with_previous_line_break_moves_it_outside_the_code_block(
+) {
+    // The initial line break will be removed, so it's the as having a single line break at the start
```
aringenbach, line 199:
```suggestion
    // The initial line break will be removed, so it's the same as having a single line break at the start
```

aringenbach (approved):
LGTM

PR #399, opened 2022-12-14
Create lists from multiline selection

aringenbach:
Allows creating lists from multiline selection and extend range appropriately.
Also remove some of the legacy list code to use new functions

### DIFF
--- a/crates/wysiwyg/src/dom/dom_list_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_list_methods.rs
@@ -44,6 +44,8 @@ where
         let mut list_items = Vec::new();
         let mut line_break_positions = list_item.line_break_positions();
 
+        // Slice the list item on each line break position from last to
+        // first position and create a new list item for each slice.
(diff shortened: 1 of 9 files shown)


crates/wysiwyg/src/dom/dom_list_methods.rs
```diff
@@ -0,0 +1,147 @@
...
+        removed_nodes.reverse();
+
+        let mut list_item = ContainerNode::new_list_item(removed_nodes);
+        // Set an arbitrary handle allows us to transform this while detached from DOM.
+        list_item.set_handle(DomHandle::root());
+        let mut list_items = Vec::new();
+        let mut line_break_positions = list_item.line_break_positions();
+
+        while let Some(position) = line_break_positions.pop() {
+            let mut sliced = list_item.slice_after(position);
```
jmartinesp, line 48: Some comment here explaining how we're splitting the nodes by line breaks would be nice.

jmartinesp (approved):
LGTM!

PR #400, opened 2022-12-14
Add rich/plain conversion functions

alunturner:
We need to be able to convert between rich and plain modes consistently when moving between composers on web.

These functions allow us to get the rich/plain text in the format required for transitioning between the composers. They achieve this by creating a model purely to generate the output. The model is not used after this initial conversion.

### DIFF
--- a/platforms/web/lib/conversion.ts
+++ b/platforms/web/lib/conversion.ts
@@ -1,0 +1,52 @@
+/*
+Copyright 2022 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// rust generated bindings
+import {
+    // eslint-disable-next-line camelcase
+    new_composer_model,
+} from '../generated/wysiwyg.js';
+
+// const backslash = String.fromCharCode(0x005c);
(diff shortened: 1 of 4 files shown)


platforms/web/lib/conversion.ts
```diff
@@ -0,0 +1,51 @@
...
+*/
+
+// rust generated bindings
+import {
+    // eslint-disable-next-line camelcase
+    new_composer_model,
+} from '../generated/wysiwyg.js';
+
+export function richToPlain(richText: string) {
+    if (richText.length === 0) {
```
florianduros, line 24: If we can have a small test here for this case, it will be great, thanks!

platforms/web/lib/conversion.ts
```diff
@@ -0,0 +1,51 @@
...
+    // in plain text, newlines will be represented as \n for display
+    // as the display box can not interpret markdown
+    const markdown = model.get_content_as_markdown();
+    const plainText = markdown.replaceAll(/\\/g, '');
+
+    return plainText;
+}
+
+export function plainToRich(plainText: string) {
+    if (plainText.length === 0) {
```
florianduros, line 39: If we can have a small test here for this case, it will be great, thanks!

platforms/web/lib/conversion.ts
```diff
@@ -0,0 +1,51 @@
...
+        return '';
+    }
+    // in plain text, newlines will be represented as \n if input
+    // by the user pressing enter, so we want to convert these to
+    // valid markdown for parsing by MarkdownHTMLParser::to_html
+    const markdown = plainText.replaceAll(/\n/g, '<br />\n');
+    const model = new_composer_model();
```
florianduros, line 46: If we don't call init previously, it will not work.

For example, when we will keep the composer mode (in the local storage for example) and we have the plain text mode by default. The wasm will not be initialised and it will not work.

We need to share here and in the function above, the same singleton for wasm initialisation.

florianduros (changes requested):
Look goods except the missing wasm initialisation.

florianduros (approved):

PR #401, opened 2022-12-14
Add link function

florianduros:


platforms/web/lib/useWysiwyg.formatting.test.tsx
```diff
@@ -218,3 +218,46 @@ describe('insertText', () => {
...
+    });
+
+    it('Should transform the selected text into link', async () => {
+        // When
+        fireEvent.input(textbox, {
+            data: 'foobar',
+            inputType: 'insertText',
+        });
+        select(textbox, 0, 6);
+        await act(() => userEvent.click(buttonLink));
```
alunturner, line 256: I've had a bit of a play on your branch and I don't think that the `await act(...)` wrapper is required on any calls to userEvent. At least not in these new tests anyway. Personally I'd remove them as I think it increases the legibility of the tests, but up to you.

PR #403, opened 2022-12-14
[iOS] Fix for a bug where attributed string keeps format style for links and inline code after deletion

Velin92:
Happens when these formats are at the start of the textview and get deleted/fully cleared

Also added an init parameter that allows the TextView to either have links with the underline or not.

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
@@ -215,7 +215,8 @@ public extension WysiwygComposerViewModel {
             return true
         }
         
-        // This fixes a bug where the attributed string keeps link and inline code formatting when they are the last formatting to be deleted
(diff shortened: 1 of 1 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -116,12 +116,16 @@ public class WysiwygComposerViewModel: WysiwygComposerViewModelProtocol, Observa
...
+                linkColor: UIColor = .link,
+                shouldUnderlineLinks: Bool = false) {
         self.minHeight = minHeight
         self.maxCompressedHeight = maxCompressedHeight
         self.maxExpandedHeight = maxExpandedHeight
         self.textColor = textColor
         self.linkColor = linkColor
+        if !shouldUnderlineLinks {
+            textView.linkTextAttributes[.underlineStyle] = 0
+        }
```
Velin92, line 128: requested by Callum, was quite easy to implement so I added it in this PR.

aringenbach (approved):
LGTM

PR #404, opened 2022-12-15
[iOS] Inline code background colour is shown in the composer

Velin92:
Also, can be customised.

simulator_screenshot_EFFEC735-599E-4EB7-954A-BD513EA6FBD6

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/NSAttributedString+Attributes.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/NSAttributedString+Attributes.swift
@@ -78,7 +78,7 @@ extension NSAttributedString {
         }
         
         // Adding background to inline code
-        mutableAttributed.enumerateTypedAttribute(.font) { (font: UIFont, range: NSRange, _: UnsafeMutablePointer<ObjCBool>) in
+        mutableAttributed.enumerateTypedAttribute(.font) { (font: UIFont, range, _) in
             if font.familyName == "Courier" {
                 mutableAttributed.addAttributes([.backgroundColor: codeBackgroundColor], range: range)
             }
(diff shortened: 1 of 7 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/NSAttributedString+Attributes.swift
```diff
@@ -62,18 +62,27 @@ extension NSAttributedString {
...
         // This fixes an iOS bug where if some text is typed after a link, and then a whitespace is added the link color is overridden.
         mutableAttributed.enumerateAttribute(.link, in: NSRange(location: 0, length: mutableAttributed.length)) { value, range, _ in
             if value != nil {
                 mutableAttributed.addAttributes([.foregroundColor: linkColor], range: range)
             }
         }
+        
+        // Adding background to inline code
+        mutableAttributed.enumerateTypedAttribute(.font) { (font: UIFont, range, _) in
+            if font.familyName == "Courier" {
```
aringenbach, line 82: Maybe something like `font.fontDescriptor.symbolicTraits.contains(.traitMonoSpace)` is a more fitting test for this. 

But this is one more reason to integrate DTCoreText inside the library to make customisation easier I guess.

Velin92: Wait I will check if I can make this work by using CSS styling. I think we also want to change the font, because I think we use Menlo-Regular but iOS by default parses Courier from HTML.

If this CSS styling doesn't work yeah we might need this workaround for now and in the future integrating DTCoreText would make things a lot easier and more customisable

Velin92: DONE!

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Tools/HTMLParser.swift
```diff
@@ -30,23 +30,48 @@ final class HTMLParser {
     ///   - html: HTML to parse
     ///   - encoding: string encoding to use
     ///   - textColor: text color to apply to the result string
+    ///   - codeBackgroundColor: color to apply to the background of code blocks
```
aringenbach, line 33: could you add missing link color as well 🙏

aringenbach (approved):
LGTM

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Tools/HTMLParser.swift
Velin92: ahaha definitively

PR #406, opened 2022-12-15
Add minimal ZWSP support to code blocks.

jmartinesp:
It always adds a ZWSP TextNode at the start of the code block and removes it when the code block is removed, adjusting the selection


aringenbach (approved):
LGTM

PR #409, opened 2022-12-15
Fix: disable inline code and then type text re-orders nodes

jmartinesp:
I didn't realise that when we unformat we actually remove the whole formatting node and the re-apply formatting to the part that should be formatted.

That, and a bug when working with nodes in `depth = 0` caused this weird issue where:

```
<i>Test</i><code>code|</code>
```

Disabling inline code and then typing would result in:

```
<code>code</code><i>Test</i>New text|
```

### DIFF
--- a/crates/wysiwyg/src/composer_model/format.rs
+++ b/crates/wysiwyg/src/composer_model/format.rs
@@ -128,8 +128,12 @@ where
         format: &InlineFormatType,
     ) {
         assert!(start != end);
-        let range = self.state.dom.find_range(start, end);
-        self.format_several_nodes(&range, format);
+        if *format == InlineFormatType::InlineCode {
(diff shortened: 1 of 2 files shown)


jmartinesp:
> I would add an additional test like this:
> 
> ```
> let mut model = cm("");
>         model.bold();
>         model.replace_text("bold".into());
>         model.italic();
>         model.inline_code();
>         model.replace_text("code".into());
>         assert_eq!(tx(&model), "<b>bold<\b><code><\code>");
> ```
> 
> Essentially what happens if before code block also a formatting is already active and one is activated before the inline code block

Done!

Velin92 (approved):

Velin92 (approved):
LGTM

PR #411, opened 2022-12-15
Add `ZwspNode` as a new type of node

aringenbach:
* Create ZwspNode type
* Add it as a type of DomNode
* Add first utilities
* Add todos for cases where it is not handled yet

### DIFF
--- a/bindings/wysiwyg-wasm/src/lib.rs
+++ b/bindings/wysiwyg-wasm/src/lib.rs
@@ -475,7 +475,7 @@ pub struct DomHandle {
 
 #[wasm_bindgen]
 impl DomHandle {
-    /// Returns "container", "line_break" or "text" depending on the type of
+    /// Returns "container", "line_break", "text" or "zwsp" depending on the type of
     /// node we refer to.
     /// Panics if we are not a valid reference (because the model has changed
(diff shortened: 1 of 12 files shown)


bindings/wysiwyg-wasm/src/lib.rs
```diff
@@ -532,6 +534,7 @@ impl DomHandle {
             wysiwyg::DomNode::Container(node) => node.name().to_string(),
             wysiwyg::DomNode::LineBreak(node) => node.name().to_string(),
             wysiwyg::DomNode::Text(_) => String::from("-text-"),
+            wysiwyg::DomNode::Zwsp(_) => String::from("-text-"),
```
jmartinesp, line 537: Copy & paste error (`-text-`)?

jmartinesp (approved):
LGTM, just minor remark.

bindings/wysiwyg-wasm/src/lib.rs
aringenbach: Actually I wasn't sure if this string changes something or not for the web implementation. If it doesn't, we probably should set something like `"-zwsp-"` yeah.

PR #412, opened 2022-12-15
Replace old text node zwsp functions with new node variant

aringenbach:
* Replaces usage of zwsp char at the start of a TextNode by actual ZwspNode

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -370,8 +370,11 @@ where
             DomNode::Container(c) => c.add_leading_zwsp(),
             DomNode::Zwsp(_) => false,
             _ => {
-                self.children.insert(0, DomNode::new_zwsp());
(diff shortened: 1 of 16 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -366,7 +366,18 @@ where
...
-        first_child.add_leading_zwsp()
+        match first_child {
+            DomNode::Container(c) => c.add_leading_zwsp(),
+            DomNode::Zwsp(_) => false,
+            _ => {
+                if self.handle().is_set() {
+                    self.insert_child(0, DomNode::new_zwsp());
+                } else {
+                    self.children.insert(0, DomNode::new_zwsp());
+                }
```
aringenbach, line 377: Wrote this code, which is not really required on current context because we are on a detached state from Dom. So we are fine with unset handles. 
I wonder though if what we want as a solution is this kind of thing or rather setting up a fake handle when transforming items in detached state.

Velin92 (approved):

PR #414, opened 2022-12-19
Quotes & fixing code blocks

jmartinesp:
## Changes in this PR:

* Added support for Quotes.
* Code blocks and Quotes now use the same shared `split_sub_tree_between` method to retrieve the nodes that they need to wrap.
* `split_sub_tree` was re-written from scratch, it can now select only part of the contents of a block node (i.e. lists). Also, it's now private, having 2 ways to access it:
  * `split_sub_tree_from`: given a DomHandle, a start offset and an initial depth, it will split the Dom into two halves, returning the 'rightmost' part.
  * `split_sub_tree_between`: similar to the previous one, but you also pass `to_handle` and `end_offset`. With these, you don't split the Dom in 2 parts, but instead extract everything between `from_handle` and `to_handle` and return it.
* Quotes can now include code blocks, code blocks can wrap quotes and add their contents.
* Both nodes should now always have a ZWSP node as their first child.

## Things that are probably still broken:

* Selection in code blocks after applying or removing them. The current tests work, but I'm pretty sure the checks I've put into place don't cover all corner cases, so we might need to find better ones.
* Nested `quote > list > quote > list` hierarchies. I haven't really tested them, as it seemed quite complex.
* Deletion of first/last characters in code blocks and quotes, they might have weird behaviours.

I probably forgot other minor stuff that might come back to bite us later.


crates/wysiwyg/src/composer_model/quotes.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    #[test]
+    fn apply_quote_to_several_nodes_with_line_breaks() {
+        let mut model = cm("Some {text<br /><b>Next}| line</b>");
+        model.quote();
+        assert_eq!(
+            tx(&model),
+            "<blockquote>~Some {text<br /><b>Next}| line</b></blockquote>"
+        )
+    }
```
jonnyandrew, line 238: It looks like the block quote wraps the whole editor content. But if you've selected a part of the content, wouldn't you expect only the selected part to be quoted? I.e.
```suggestion
    fn apply_quote_to_several_nodes_with_line_breaks() {
        let mut model = cm("Some {text<br /><b>Next}| line</b>");
        model.quote();
        assert_eq!(
            tx(&model),
            "Some <blockquote>~{text<br /><b>Next</b>}|</blockquote><b> line</b>"
        )
    }
```

Similarly, on unformatted text wouldn't we expect the following behaviour:
```
    #[test]
    fn apply_quote_to_simple_text_start() {
        let mut model = cm("{Some}| text");
        model.quote();
        assert_eq!(tx(&model), "<blockquote>~Some|</blockquote> text")
    }
```

crates/wysiwyg/src/composer_model/code_block.rs
```diff
@@ -435,58 +247,97 @@ where
...
-                    }
-                    _ => continue,
-                }
-            }
-            panic!("Should never reach this point, one of the parents surely can be split.");
-        }
-    }
-
-    pub(crate) fn format_node_for_code_block(node: &DomNode<S>) -> DomNode<S> {
+    pub(crate) fn format_node_for_code_block(
```
jonnyandrew, line 250: It's a bit hard to follow what's happening in this method because of the deeply nested branches. Perhaps we could refactor a bit to flatten it and/or add some comments to explain how it works?

jonnyandrew (commented):
Nice one so far! Some of this is pretty complex making it difficult for me to review the code in detail. Could you add a few comments to explain how things work?

crates/wysiwyg/src/composer_model/quotes.rs
jonnyandrew: Actually, ignore this. I just tried in a few different editors and now I understand what we're going for.

jmartinesp:
I moved the code as you suggested and added some extra comments to try to make the code easier to understand.

jonnyandrew (approved):
Thanks for the changes! I've approved but I think a second review would be beneficial for this as it's such a big PR

crates/wysiwyg/src/composer_model/code_block.rs
```diff
@@ -435,149 +247,114 @@ where
...
                 for c in container.children() {
-                    children.push(Self::format_node_for_code_block(c));
+                    children.extend(self.format_node_for_code_block(
+                        c, range, first_leaf, last_leaf,
+                    ));
                 }
-                DomNode::Container(container.clone_with_new_children(children))
+                // Then post-process the containers to add line breaks and move selection if needed
+                if container.is_block_node() || container.is_list_item() {
+                    if container.is_list() {
```
Velin92, line 287: the content from 287 to  338 could be a function of it's since it's very nested (and maybe some of it's own nested pieces may be other functions?)

crates/wysiwyg/src/composer_model/code_block.rs
```diff
@@ -435,149 +247,114 @@ where
...
+                            } else if location.starts_inside() {
+                                if added_at_start {
+                                    self.state.start += 1;
+                                }
+                                self.state.end += 1;
+                            } else if location.ends_inside() {
+                                self.state.start += 1;
+                                self.state.end += 1;
+                            }
+                        }
```
Velin92, line 336: this definitely could be a separate function that handles the Quote case.

crates/wysiwyg/src/composer_model/quotes.rs
```diff
@@ -0,0 +1,319 @@
...
+
+    fn add_quote(&mut self) -> ComposerUpdate<S> {
+        let (s, e) = self.safe_selection();
+        let Some(wrap_result) = self.state.dom.find_nodes_to_wrap_in_block(s, e) else {
+            // No suitable nodes found to be wrapped inside the quote. The Dom should be empty
+            self.state.dom.append_at_end_of_document(DomNode::new_quote(vec![DomNode::new_text(S::zwsp())]));
+            self.state.start += 1;
+            self.state.end += 1;
+            return self.create_update_replace_all();
```
Velin92, line 40: why this and not returning `keep()`?

crates/wysiwyg/src/composer_model/quotes.rs
```diff
@@ -0,0 +1,319 @@
...
+            self.remove_quote()
+        } else {
+            self.add_quote()
+        }
+    }
+
+    fn add_quote(&mut self) -> ComposerUpdate<S> {
+        let (s, e) = self.safe_selection();
+        let Some(wrap_result) = self.state.dom.find_nodes_to_wrap_in_block(s, e) else {
+            // No suitable nodes found to be wrapped inside the quote. The Dom should be empty
```
Velin92, line 36: I would change this comment to:
```
// No wrappable nodes found.
// Adding an empty Quote block with an a single ZWSP
```

crates/wysiwyg/src/composer_model/quotes.rs
```diff
@@ -0,0 +1,319 @@
...
+            } else {
+                start_handle.sub_handle_up_to(parent_handle.depth() + 1)
+            };
+        if !start_handle_is_start_at_depth
+            && self.state.dom.contains(&insert_at_handle)
+        {
+            insert_at_handle = insert_at_handle.next_sibling();
+        } else if self.state.dom.document().children().is_empty() {
+            insert_at_handle = self.state.dom.document_handle().child_handle(0);
+        }
```
Velin92, line 72: similiar code already seen in code blocks... maybe we can make a function out of it?

crates/wysiwyg/src/composer_model/replace_text.rs
```diff
@@ -179,41 +184,100 @@ where
...
+                new_line_at_end: bool,
+            ) {
+                dom.insert_at(&at_handle, sub_tree);
+                let insert_new_line_at = if new_line_at_end {
+                    at_handle.next_sibling()
+                } else {
+                    at_handle.clone()
+                };
+                dom.insert_at(&insert_new_line_at, DomNode::new_line_break());
+            }
```
Velin92, line 213: Why not having this function as part of file instead of being inside the scope of another function?

Velin92 (approved):
LGTM, just pointed out some possible code improvements to help with the overall readability

crates/wysiwyg/src/composer_model/code_block.rs
jmartinesp: Done in 3b957ad1a90e53360a56816dde413d8954ee7a98

crates/wysiwyg/src/composer_model/code_block.rs
jmartinesp: Extracted to functions in 81750202bd4f1e5ebdfeebcefdd7eb4d9837fb87

crates/wysiwyg/src/composer_model/quotes.rs
jmartinesp: Changed it to something similar in 9f6f3bdf7afa49dfab185836bf73b47f7db0c8bb

crates/wysiwyg/src/composer_model/code_block.rs
```diff
@@ -36,188 +33,104 @@ where
...
-                insert_at_handle
+                start_handle.sub_handle_up_to(parent_handle.depth() + 1)
             };
+        if !start_handle_is_start_at_depth
+            && self.state.dom.contains(&insert_at_handle)
+        {
+            insert_at_handle = insert_at_handle.next_sibling();
+        } else if self.state.dom.document().children().is_empty() {
+            insert_at_handle = self.state.dom.document_handle().child_handle(0);
+        }
```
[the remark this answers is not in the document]
jmartinesp: Done that in 7d712a783764bc2a996bfd34a8070435a172c5c1. To be honest, I didn't expect it to work for both block nodes, but it does 🎊 !

crates/wysiwyg/src/composer_model/replace_text.rs
jmartinesp: It's internal to the function because it's only a helper for this function (we have to use it twice). I don't think it's worth it to have an `insert_X_node_at()` function for a concrete type.

Velin92: I see... could we at least place it either at the end of the enclosing scope or at the start of the enclosing scope?
Since having it in the middle of the enclosing scope kinda "breaks" it.

jmartinesp: Yes, it's true that it looks weird at its current position 😅 . Done in ed191dc.

PR #415, opened 2022-12-19
Add `insert_parent` DOM method

jonnyandrew:
## What?
Adds a DOM method to insert a parent node around a range.

## Why?
Currently, when adding a hyperlink to a range, it is done by applying the link to every leaf in the range. In order to wrap the leaves at a higher level in the tree, it'd be useful to have a method at the Dom level to insert a parent node around a rnage.

## What does this look like?
Take for example the following HTML with selection indicated by `{`, `}` and cursor by `|`.

```
<i><b>Hel{lo</b><u>formatted</u>wor}|ld</i>
```

It's now possible to insert a link around the range like so:

```mermaid
graph TD
classDef default stroke:#000
classDef hello fill:#ffa
classDef formatted fill:#faf
classDef world fill:#aff
classDef link fill:#cfa,stroke-width:4px

Dom[Dom before] --> Italic
Italic --> Bold
Italic --> Underline
Italic --> World:::world
Bold:::hello --> Hello
Underline:::formatted --> Formatted
Hello["&quot;Hello&quot;"]:::hello
Formatted["&quot;formatted&quot;"]:::formatted
World["&quot;world&quot;"]:::split

Dom2[Dom after] --> Italic2[Italic]
Italic2 --> Bold2a[Bold]
Bold2a:::hello --> Hello2a
Italic2 --> Link
Link:::link --> Bold2b[Bold]
Link --> Underline2
Link --> World2a
Italic2 --> World2b
Bold2b:::hello --> Hello2b
Underline2[Underline]:::formatted --> Formatted2
Hello2a["&quot;Hel&quot;"]:::hello
Hello2b["&quot;lo&quot;"]:::hello
Formatted2["&quot;formatted&quot;"]:::formatted
World2a["&quot;wor&quot;"]:::world
World2b["&quot;ld&quot;"]:::world
```

### DIFF
--- a/crates/wysiwyg/src/dom/dom_handle.rs
+++ b/crates/wysiwyg/src/dom/dom_handle.rs
@@ -90,6 +90,16 @@ impl DomHandle {
         DomHandle::from_raw(new_path.to_vec())
     }
 
+    /// Returns a DomHandle with the 'sub-path' down to the passed 'depth'.
+    /// i.e.: a DomHandle with path `[0, 1, 2, 3]` with `at_depth(2)` will return a DomHandle
+    /// with path `[3]`.
+    pub fn sub_handle_down_from(&self, depth: usize) -> DomHandle {
+        assert!(&self.path.is_some());
+        let path = self.path.clone().unwrap();
+        let (_, new_path) = path.split_at(depth);
+        DomHandle::from_raw(new_path.to_vec())
+    }
+
     /// Return true if this handle has a parent i.e. it is not the root. If
     /// this returns true, it is safe to call index_in_parent() and
     /// parent_handle().
--- a/crates/wysiwyg/src/dom/dom_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_methods.rs
@@ -17,7 +17,7 @@
 
 use crate::dom::nodes::ContainerNodeKind;
 use crate::dom::unicode_string::UnicodeStr;
-use crate::{DomHandle, DomNode, UnicodeString};
+use crate::{DomHandle, DomNode, ToHtml, UnicodeString};
 
 use super::action_list::{DomAction, DomActionList};
 use super::nodes::dom_node::DomNodeKind;
@@ -434,6 +434,73 @@ where
         self.assert_invariants();
     }
 
+    /// Returns two new subtrees as the result of splitting the Dom symetrically without mutating
+    /// itself. Also returns the new handles of node that was split.
+    ///
+    /// Only returns nodes that are modified by the split and ignores any nodes which were not
+    /// either split or contain a node that was split.
+    pub(crate) fn split_new_sub_trees(
+        &self,
+        from_handle: &DomHandle,
+        offset: usize,
+        depth: usize,
+    ) -> (DomNode<S>, DomHandle, DomNode<S>, DomHandle) {
+        let mut clone = self.clone();
+        let right = clone.split_sub_tree(from_handle, offset, depth);
+
+        // Remove unmodified children of the right split
+        // TODO: create a utility for this
+        let mut right = DomNode::Container(ContainerNode::new(
+            S::default(),
+            ContainerNodeKind::Generic,
+            None,
+            vec![right
+                .as_container()
+                .unwrap()
+                .children()
+                .first()
+                .unwrap()
+                .clone()],
+        ));
+        right.set_handle(DomHandle::root());
+
+        // Remove unmodified children of the left split
+        // TODO: create a utility for this
+        let mut left = DomNode::Container(ContainerNode::new(
+            S::default(),
+            ContainerNodeKind::Generic,
+            None,
+            vec![clone
+                .lookup_node(&from_handle.sub_handle_up_to(depth))
+                .as_container()
+                .unwrap()
+                .children()
+                .last()
+                .unwrap()
+                .clone()],
+        ));
+        left.set_handle(DomHandle::root());
+
+        // Reset the root unmodified siblings were removed
+        let mut right_handle =
+            from_handle.sub_handle_down_from(depth).raw().to_owned();
+        right_handle[0] = 0;
+        let right_handle = DomHandle::from_raw(right_handle);
+
+        let mut left_handle =
+            from_handle.sub_handle_down_from(depth).raw().to_owned();
+        left_handle[0] = 0;
+        let left_handle = DomHandle::from_raw(left_handle);
+
+        (left, left_handle, right, right_handle)
+    }
+
+    /// Splits the tree at a given node into two parts. The first part stays in the tree and this
+    /// function returns the second part.
+    ///
+    /// * `from_handle` - the position of the node to split.
+    /// * `offset` - the position within the given node to split.
+    /// * `depth` - the depth within the original tree at which to make the returned tree's root
     pub(crate) fn split_sub_tree(
         &mut self,
         from_handle: &DomHandle,
@@ -498,13 +565,14 @@ where
                             {
                                 // Do nothing
                             } else {
-                                let left_data =
-                                    text_node.data()[..offset].to_owned();
-                                let right_data =
+                                // TODO: Figure out how to make this all less verbose
+                                let to_remove =
                                     text_node.data()[offset..].to_owned();
-                                text_node.set_data(left_data);
+                                let to_keep =
+                                    text_node.data()[..offset].to_owned();
+                                text_node.set_data(to_keep);
                                 removed_nodes
-                                    .insert(0, DomNode::new_text(right_data));
+                                    .insert(0, DomNode::new_text(to_remove));
                             }
                         }
                         _ => {
@@ -632,6 +700,7 @@ fn adjust_handles_for_delete(
 mod test {
     use crate::dom::DomHandle;
     use crate::tests::testutils_composer_model::{cm, tx};
+    use crate::tests::testutils_conversion::utf16;
     use crate::ToHtml;
 
     use super::*;
@@ -702,6 +771,7 @@ mod test {
             2,
             0,
         );
+        assert_eq!(model.state.dom.to_html(), "Text<b>bo</b>");
         assert_eq!(ret.to_html().to_string(), "<b>ld</b><i>italic</i>");
     }
 
@@ -713,6 +783,7 @@ mod test {
             2,
             0,
         );
+        assert_eq!(model.state.dom.to_html(), "<u>Text<b>bo</b></u>");
         assert_eq!(ret.to_html().to_string(), "<u><b>ld</b><i>italic</i></u>");
     }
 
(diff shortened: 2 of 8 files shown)


crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -434,6 +434,73 @@ where
         self.assert_invariants();
     }
 
+    /// Returns two new subtrees as the result of splitting the Dom symetrically without mutating
+    /// itself. Also returns the new handles of node that was split.
+    ///
+    /// Only returns nodes that are modified by the split and ignores any nodes which were not
+    /// either split or contain a node that was split.
+    pub(crate) fn split_new_sub_trees(
```
jmartinesp, line 442: This code makes sense, but take into account that it can be expensive memory-wise, as you're duplicating the whole Dom. 

In the quotes PR I've modified `split_sub_tree` to also take a `to_handle: Option<DomHandle>` so you can extract just part of the tree, not halves of it, so it might be a good idea to replace this function in the future with it, or modify the `split_aub_tree` function to return an actual `Dom<S>` object instead of a container, along with the `DomHandle` of the `from_handle` and `to_handle` inside of it if needed?

crates/wysiwyg/src/dom/insert_parent.rs
```diff
@@ -0,0 +1,271 @@
...
+                    location.start_offset
+                };
+
+                let (left, left_handle, right, right_handle) =
+                    self.split_new_sub_trees(node, offset, shared_depth);
+
+                let mut inner = if location.ends_inside() {
+                    left.clone()
+                } else {
+                    right.clone()
```
jmartinesp, line 76: It would be better to not clone the sub-tree again if possible, but from the code below it seems like you need to use both `inner` and `right` as different nodes.

crates/wysiwyg/src/dom/insert_parent.rs
```diff
@@ -0,0 +1,271 @@
...
+                    continue;
+                }
+
+                let offset = if location.ends_inside() {
+                    location.end_offset
+                } else {
+                    location.start_offset
+                };
+
+                let (left, left_handle, right, right_handle) =
```
jmartinesp, line 70: This is actually what I was a bit afraid of. If I'm not mistaken, for both the start and end leaves, as well as their direct ancestors (since you're using `range.locations` instead of `range.leaves()`), you're duplicating the whole Dom, then splitting it.

crates/wysiwyg/src/dom/nodes/dom_node.rs
```diff
@@ -311,6 +311,49 @@ where
             DomNode::Zwsp(_) => panic!("Can't slice a zwsp"),
         }
     }
+
+    /// Find the node based on its handle.
+    /// Panics if the handle is unset or invalid
+    pub fn lookup_node(&self, node_handle: &DomHandle) -> &DomNode<S> {
```
jmartinesp, line 317: I don't know if it would actually be better to have `split_sub_tree` return a new `Dom` object or have `Dom.lookup_node` actually just call `Dom.container().lookup_node()`. What do you think?

Velin92 (approved):
LGTM! Great work so far!

crates/wysiwyg/src/dom/insert_parent.rs
```diff
@@ -0,0 +1,271 @@
...
+        // Check if the range has shared ancestry. The new parent node should not contain
+        // these nodes, so filter them from the range.
+        let shared_depth = range.shared_parent().depth();
+        let range = Range::new(range.locations_from_depth(shared_depth));
+
+        // Prepare the new parent node to have the selected range moved into it:
+        // - Set the new node's handle to the start of the range
+        // - Get the mutable container node to move the selected range into
+        let new_handle: DomHandle = range.node_handles().min().unwrap().clone();
+        let new_handle = if range.leaves().min().unwrap().is_covered() {
```
jmartinesp, line 31: You won't believe how long it took me to find a good way to decide where the new parent should be inserted when the tree is split (I can't exactly use range with block nodes), so good job here!

crates/wysiwyg/src/dom/dom_handle.rs
```diff
@@ -90,6 +90,16 @@ impl DomHandle {
         DomHandle::from_raw(new_path.to_vec())
     }
 
+    /// Returns a DomHandle with the 'sub-path' down to the passed 'depth'.
+    /// i.e.: a DomHandle with path `[0, 1, 2, 3]` with `at_depth(2)` will return a DomHandle
+    /// with path `[3]`.
+    pub fn sub_handle_down_from(&self, depth: usize) -> DomHandle {
```
jmartinesp, line 96: I had been thinking about building something like this for the code blocks and quotes features. Thanks!

crates/wysiwyg/src/dom/range.rs
```diff
@@ -254,10 +262,23 @@ impl Range {
             .filter(move |l| l.node_handle.depth() == depth)
     }
 
+    pub fn locations_from_depth(
+        &self,
+        depth: usize,
+    ) -> impl Iterator<Item = &DomLocation> {
+        self.locations
+            .iter()
+            .filter(move |l| l.node_handle.depth() > depth)
```
jmartinesp, line 271: I think `*l.node_handle.depth()` would also work here, instead of using `move`.

crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -434,6 +434,73 @@ where
...
+        let mut clone = self.clone();
+        let right = clone.split_sub_tree(from_handle, offset, depth);
+
+        // Remove unmodified children of the right split
+        // TODO: create a utility for this
+        let mut right = DomNode::Container(ContainerNode::new(
+            S::default(),
+            ContainerNodeKind::Generic,
+            None,
+            vec![right
```
jmartinesp, line 457: Wouldn't `let mut right = clone.split...` already give you what you need? If I'm not mistaken, right now `split_sub_tree` already returns a generic container.

jmartinesp (commented):
I have some questions and some discussions that we could have to decide how to use `split_sub_tree` in the future.

crates/wysiwyg/src/dom/dom_methods.rs
jonnyandrew: I've updated this to use the same container type that's returned from `split_sub_tree`: [`41029eb`](https://github.com/matrix-org/matrix-rich-text-editor/pull/415/commits/41029eb8bab9240631b2ab4789b9b4bbb4252fb7)

crates/wysiwyg/src/dom/range.rs
jonnyandrew: My understanding is that `move` is needed because `depth` will only live as long as the function but the closure that `filter` uses will need to run after the function returns. Will it have some bad/unintended effect other than this? 

(I have tried your suggestion but it didn't seem to work.)

crates/wysiwyg/src/dom/insert_parent.rs
jonnyandrew: We are iterating over `range.locations` but this code is only run on the leaves due to [lines 60-62](https://github.com/matrix-org/matrix-rich-text-editor/pull/415/files#diff-cfbcd9fb6c294410b18d63f5d384054ff8f86b4a956e5fe98518a346f3ea5861R60-R62):
```
if !self.lookup_node(node).is_leaf() {
    continue;
}
```

crates/wysiwyg/src/dom/insert_parent.rs
jonnyandrew: You're right, cloning the subtree again is not necessary - thank you for spotting this!

[`40be2d1`](https://github.com/matrix-org/matrix-rich-text-editor/pull/415/commits/40be2d1e2ece87068303169572873d96312f6ed2)

crates/wysiwyg/src/dom/dom_methods.rs
jonnyandrew: > In the quotes PR I've modified `split_sub_tree` to also take a `to_handle: Option<DomHandle>` so you can extract just part of the tree, not halves of it, so it might be a good idea to replace this function in the future with it,

Thank you for letting me know about this, once the modified `split_sub_tree` method is here I'll update this one to use this.

crates/wysiwyg/src/dom/nodes/dom_node.rs
jonnyandrew: I thought about this, but figured it'd be a bit confusing to have `Dom` contain sub-trees since they do not represent the whole document. But I may not be understanding the architecture here. Maybe @andybalaam can comment?

crates/wysiwyg/src/dom/insert_parent.rs
jmartinesp: Oh, I see. Sorry, I completely missed those lines 👍 .

crates/wysiwyg/src/dom/nodes/dom_node.rs
jmartinesp: Our `Dom` is just a structure that contains a tree and has helper methods for accessing it, so I think it would make sense to use one to temporarily host and modify the sub-tree. This would allow us to use helper `insert(&DomHandle)`, `remove(&DomHandle)`, etc., as well as look up, splitting, iterators, etc. 

As long as we make clear that the returned Dom is a sub-tree and not the original one I'd say it's probably fine to use a Dom object.

crates/wysiwyg/src/dom/range.rs
jmartinesp: You're right, I didn't realise you were returning the iterator, not a collected version of it, so the `move` keyword is needed.

crates/wysiwyg/src/dom/dom_methods.rs
jmartinesp: I was actually wrong here, I thought you could re-use the result of calling `split_sub_tree` but it's not possible to do so here. The new version is more readable, thanks!

jmartinesp (approved):
The new changes LGTM!

crates/wysiwyg/src/dom/nodes/dom_node.rs
jonnyandrew: Ok, that makes sense. I just tried to refactor this to replace `DomNode` with `Dom` but I was then unable to make it work without undoing the improvement in [`40be2d1`](https://github.com/matrix-org/matrix-rich-text-editor/pull/415/commits/40be2d1e2ece87068303169572873d96312f6ed2) (i.e. I needed to clone the subtree `Dom` children in order to move them into the new parent node).

If you know a way to do this work please let me know and I'll make a follow-up PR with this

jmartinesp: Oh, sorry. I didn't mean that you should do it now, I was just arguing that we should probably make that change in the future. Since I already modified `split_sub_tree` on my other PR I'll try to do this change myself, then adapt existing code to it.

PR #417, opened 2022-12-20
Amend conversion functions

alunturner:
Amends the functions used for converting between rich and plain text in web. 

Corrects issue with trailing newlines when switching between rich and plain with the setting enabled that makes enter insert newlines, as opposed to sending the message.


florianduros (approved):
LGFM

PR #418, opened 2022-12-20
[iOS] new attributed string to html mapping for lists with ZWSP

Velin92:


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/NSAttributedString+Range.swift
```diff
@@ -100,17 +92,23 @@ extension NSAttributedString {
...
+        let prefixes = listPrefixesRanges()
         var actualIndex: Int = attributedIndex
+        
+        guard actualIndex > 0 else {
+            return actualIndex
+        }
+        
+        if !prefixes.isEmpty {
+            actualIndex += 1
+        }
```
aringenbach, line 111: I think we need to add some documentation on where this extra offset comes from. 
And if it is what I suppose, I guess this code won't work properly if we have multiple separated lists ?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/NSAttributedString+Range.swift
```diff
@@ -129,13 +127,19 @@ extension NSAttributedString {
...
         var actualIndex: Int = htmlIndex
-
+        
+        guard actualIndex > 0 else {
+            return actualIndex
+        }
+        
+        if !prefixes.isEmpty {
+            actualIndex -= 1
+        }
```
aringenbach, line 141: Same here

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -406,8 +404,7 @@ private extension WysiwygComposerViewModel {
         do {
             let htmlSelection = NSRange(location: Int(start), length: Int(end - start))
             // FIXME: temporary workaround as trailing newline should be ignored but are now replacing ZWSP from Rust model
```
aringenbach, line 406: same here

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/NSAttributedString+Range.swift
Velin92: I think this is caused by the fact that iOS adds an extra newline character at the end of a list when this is created.
And yeah I am actually experiencing some issues when creating multiple lists one after another.
Will look a bit more into it, but the best thing would be to find a way to just remove this extra newline when a list is created.

Velin92: I honestly don't know what is actually causing the need for this offset, but it's actually making the test and the deleting and adding work, in the tests also where the scenario has multiple lists.

However there are problems:
A minor issue where iOS adds a newline after a list has been created at the end of the list itself.

Then there is a major issue:  essentially the attributed string wants to keep the list paragraph style even if I terminate the list with a double enter, for no apparent reason...
It first starts rendering the text as normal but as soon as you press enter it becomes a new list again (however the model doesn't see it as a list, is just the attributed paragraph style that is enforcing it)
I tried a lot of things changing the typedText value, removing the paragraph style from the last characters... and so on.

In terms of mapping the PR is fine, but we have these two separate issue, where the last one is looking like a hell to fix.
I am starting to think that maybe DTCoreText is the way to go...

Velin92: Actually could this error flagged by @jonnyandrew be the cause?
https://matrix.to/#/!UCBgpJmTgAHYpcJBoh:matrix.org/$6CTxShE6ZdoK1BfgjdhJUz64yPWj-AQxWuxDXrAD-T8?via=matrix.org&via=element.io&via=one.ems.host

PR #420, opened 2022-12-20
Fix edge case for `insert_parent`

jonnyandrew:
Fix an issue in `Dom.insert_parent` where the new container is inserted in the wrong position in the tree. The affected edge case was where the selection ends inside the first leaf.

- Related: https://github.com/matrix-org/matrix-rich-text-editor/pull/415


jmartinesp (approved):
LGTM!

PR #422, opened 2022-12-20
[iOS] RTE release workflow to open a Swift PM repo release PR

Velin92:


release_ios.sh
```diff
@@ -1,13 +1,20 @@
...
-else 
-  echo "Not on main branch. Exiting..."
-  exit 1
-fi
+#if [ $BRANCH_NAME == "main" ]; then 
+#  echo "On main branch."
+#else 
+#  echo "Not on main branch. Exiting..."
+#  exit 1
+#fi
```
Velin92, line 17: this needs to be restored when this is fully tested

release_ios.sh
```diff
@@ -25,12 +32,15 @@ WYSIWYG_COMPOSER_PATH="platforms/ios/lib/WysiwygComposer/"
 rm -rf $BUILD_DIR
 mkdir $BUILD_DIR
 git clone git@github.com:matrix-org/matrix-wysiwyg-composer-swift.git $REPO_PATH
-git checkout main
+# git checkout main
```
Velin92, line 35: this needs to be restored when fully tested

release_ios.sh
```diff
@@ -25,12 +32,15 @@ WYSIWYG_COMPOSER_PATH="platforms/ios/lib/WysiwygComposer/"
...
 rsync -a --delete --exclude=".git" $WYSIWYG_COMPOSER_PATH $REPO_PATH
 last_commit=$(git rev-parse --short HEAD);
 RELEASE_BRANCH="release_$last_commit"
 cd $REPO_PATH
 git checkout -b $RELEASE_BRANCH
 git add .
 git commit -m "release $last_commit"
+if ["$TAG"]; then 
+  git tag $TAG
+fi
```
Velin92, line 45: I don't know why but this line here doesn't seem to work.
Essentially I want to add a tag if the $TAG value exists, which should be passed by calling the the script with the -t tag option

release_ios.sh
```diff
@@ -1,13 +1,20 @@
 #!/usr/bin/env bash
 
+options=':t'
+while getopts $options option; do
+  case "$option" in
+    t) TAG=$OPTARG;;
+  esac
+done
```
Velin92, line 8: I don't know why but when I pass the -t argument in the scripts it's always an empty string.
For example if I run /.release_ios.sh -t 0.0.1 I get the list of tags since the git tag command is called with an empty string instead of creating one

PR #423, opened 2022-12-21
Update link behaviour to wrap existing elements

jonnyandrew:
## What?

Allow links to wrap other elements.

### Example
Given the following editor content:
```
<b><u>Hello</u> <i>world</i></b>
```
When applying a link to the whole range, previously you'd end up with:
```
<b><u><a href="">Hello</a></u> <i><a href="">world</a></i></b>
```
Now, only one link tag is created:
```
<a href=""><b><u>Hello</u> <i>world</i></b></a>
```
### Notes
- This change does not pay much attention to the case where the selection touches existing links (see [`PSU-997`](https://element-io.atlassian.net/browse/PSU-997)). But note that the behaviour is slightly different
  - links can now be split just like any other element (previously, the whole link was edited).
  - links that completely contain the selection, are edited.
- This change does not affect insertion (i.e. creating new text with a link)

## Why?
This simplifies the document model and reduces the number of duplicate tags, making the ability to edit links simpler.

## See
- [`PSU-997`](https://element-io.atlassian.net/browse/PSU-997)
- https://github.com/matrix-org/matrix-rich-text-editor/pull/415

### DIFF
--- a/crates/wysiwyg/src/composer_model/hyperlinks.rs
+++ b/crates/wysiwyg/src/composer_model/hyperlinks.rs
@@ -13,10 +13,13 @@
 // limitations under the License.
 
 use crate::dom::nodes::dom_node::DomNodeKind;
+use crate::dom::nodes::ContainerNodeKind;
 use crate::dom::nodes::{ContainerNodeKind::Link, DomNode};
 use crate::dom::unicode_string::UnicodeStrExt;
 use crate::dom::{DomLocation, Range};
-use crate::{ComposerModel, ComposerUpdate, LinkAction, UnicodeString};
+use crate::{
+    ComposerModel, ComposerUpdate, DomHandle, LinkAction, UnicodeString,
+};
 
 impl<S> ComposerModel<S>
 where
@@ -58,9 +61,63 @@ where
 
     pub fn set_link(&mut self, link: S) -> ComposerUpdate<S> {
         self.push_state_to_history();
-        let (s, e) = self.safe_selection();
-        let range = self.state.dom.find_range(s, e);
-        self.set_link_range(range, link)
+        let (mut s, mut e) = self.safe_selection();
+
+        let mut range = self.state.dom.find_range(s, e);
+
+        // Find container link that completely covers the range
+        if let Some(link) = self.find_parent_links(&range) {
+            // If found, update the range to the container link bounds
+            range = self.state.dom.find_range_by_node(&link);
+            (s, e) = (range.start(), range.end());
+        }
+
+        if s == e {
+            return ComposerUpdate::keep();
+        }
+
+        let inserted = self
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(link, vec![]));
+
+        // Ensure no duplication by deleting any links contained within the new link
+        self.delete_child_links(&inserted);
+
+        self.create_update_replace_all()
+    }
+
+    fn delete_child_links(&mut self, node_handle: &DomHandle) {
+        let node = self.state.dom.lookup_node(node_handle);
+
+        for handle in node
+            .iter_containers()
+            .filter(|n| matches!(n.kind(), ContainerNodeKind::Link(_)))
+            .map(|n| n.handle())
+            .filter(|h| *h != *node_handle)
+            .collect::<Vec<_>>()
+            .into_iter()
+            .rev()
+        {
+            self.state.dom.remove_and_keep_children(&handle);
+        }
+    }
+
+    fn find_parent_links(&mut self, range: &Range) -> Option<DomHandle> {
+        let mut parent_handle = range.shared_parent();
+        while !parent_handle.is_root() {
+            let node = self.state.dom.lookup_node(&parent_handle);
+            let container = match node {
+                DomNode::Container(container) => container,
+                _ => continue,
+            };
+            if matches!(container.kind(), ContainerNodeKind::Link(_)) {
+                return Some(node.handle());
+            }
+            parent_handle = parent_handle.parent_handle();
+        }
+
+        None
     }
 
     fn set_link_range(&mut self, range: Range, link: S) -> ComposerUpdate<S> {
--- a/crates/wysiwyg/src/composer_model/hyperlinks.rs
+++ b/crates/wysiwyg/src/composer_model/hyperlinks.rs
@@ -66,7 +66,7 @@ where
         let mut range = self.state.dom.find_range(s, e);
 
         // Find container link that completely covers the range
-        if let Some(link) = self.find_parent_links(&range) {
+        if let Some(link) = self.find_closest_ancestor_link(&range) {
             // If found, update the range to the container link bounds
             range = self.state.dom.find_range_by_node(&link);
             (s, e) = (range.start(), range.end());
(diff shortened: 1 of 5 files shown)


crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -58,9 +61,63 @@ where
...
+        let node = self.state.dom.lookup_node(node_handle);
+
+        for handle in node
+            .iter_containers()
+            .filter(|n| matches!(n.kind(), ContainerNodeKind::Link(_)))
+            .map(|n| n.handle())
+            .filter(|h| *h != *node_handle)
+            .collect::<Vec<_>>()
+            .into_iter()
+            .rev()
```
jonnyandrew, line 100: I wasn't sure how to make `DomNode.iter_containers()` return a double ended iterator, hence `collect()` followed by `rev()`

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -58,9 +61,63 @@ where
...
+            .dom
+            .insert_parent(&range, DomNode::new_link(link, vec![]));
+
+        // Ensure no duplication by deleting any links contained within the new link
+        self.delete_child_links(&inserted);
+
+        self.create_update_replace_all()
+    }
+
+    fn delete_child_links(&mut self, node_handle: &DomHandle) {
```
jonnyandrew, line 90: I wasn't sure if this method belonged in the composer layer or whether it might be better abstracted into the DOM layer. Open to ideas!

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -58,9 +61,63 @@ where
...
+        let (mut s, mut e) = self.safe_selection();
+
+        let mut range = self.state.dom.find_range(s, e);
+
+        // Find container link that completely covers the range
+        if let Some(link) = self.find_parent_links(&range) {
+            // If found, update the range to the container link bounds
+            range = self.state.dom.find_range_by_node(&link);
+            (s, e) = (range.start(), range.end());
+        }
```
jmartinesp, line 73: If there is already a parent link and we found it, couldn't we just modify it to update its link value instead? Or replace it with a copy that contains the same children but updated values.

I might be missing some corner case.

crates/wysiwyg/src/composer_model/hyperlinks.rs
jmartinesp: If we could pass the `ContainerNodeKind` to match it could make sense to make it `pub(crate)` and add it to the Dom layer instead as `remove_child_nodes_of_kind(...)` or something like that.

crates/wysiwyg/src/composer_model/hyperlinks.rs
jmartinesp: I don't think there's a better way at the moment, to be honest. In the quotes PR I made some changes to the iterators, but I don't know if we could avoid collecting those items even with those changes.

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -176,6 +178,22 @@ where
...
+
+        let locations = match result {
+            FindResult::Found(locations) => locations,
+            _ => panic!("Node does not exist"),
+        };
+
+        let leaves = locations.iter().filter(|l| l.is_leaf());
+
+        let s = leaves.clone().map(|l| l.position).min().unwrap();
+        let e = leaves.clone().map(|l| l.position + l.length).max().unwrap();
```
jmartinesp, line 192: You probably don't need a 2nd clone here?

crates/wysiwyg/src/dom/range.rs
```diff
@@ -310,10 +328,27 @@ impl Range {
...
 
         DomHandle::from_raw(shared_path)
     }
+
+    fn find_location(&self, node_handle: &DomHandle) -> &DomLocation {
+        self.locations
+            .iter()
+            .find(|l| *l.node_handle.raw() == *node_handle.raw())
+            .unwrap()
+    }
```
jmartinesp, line 351: It would probably be safer to return an `Option<&DomLocation>` instead and handle the `None` case in the calling function. Also, this is done a few times in the codebase, so having it as a `pub(crate)  fn` so we can replace those usages could be useful.

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -58,9 +61,63 @@ where
...
+            .filter(|h| *h != *node_handle)
+            .collect::<Vec<_>>()
+            .into_iter()
+            .rev()
+        {
+            self.state.dom.remove_and_keep_children(&handle);
+        }
+    }
+
+    fn find_parent_links(&mut self, range: &Range) -> Option<DomHandle> {
```
jmartinesp, line 106: This could be `find_closest_ancestor_link`? Since it returns an `Option<DomHandle>`, the name seems a bit misleading.

jmartinesp (commented):
I have a few questions and suggestions.

crates/wysiwyg/src/composer_model/hyperlinks.rs
jonnyandrew: That may be more efficient but I think it'd be safer (and simpler) to process this special case in the same way as other links (so that there is only one code path).

crates/wysiwyg/src/composer_model/hyperlinks.rs
jonnyandrew: I tried to refactor this into something a bit more general but it got a bit complicated when we start needing to handle non container nodes (and is too specific IMO to warrant refactoring if we handle just the containers).

crates/wysiwyg/src/dom/dom_struct.rs
jonnyandrew: [`a753a13`](https://github.com/matrix-org/matrix-rich-text-editor/pull/423/commits/a753a134040615916b28415de7c68bb257e9d686)

crates/wysiwyg/src/dom/range.rs
jonnyandrew: [`28e06db`](https://github.com/matrix-org/matrix-rich-text-editor/pull/423/commits/28e06dbf41aae9bb20763ee8abdbfb409259bfeb)

crates/wysiwyg/src/composer_model/hyperlinks.rs
jonnyandrew: [`8180d8d`](https://github.com/matrix-org/matrix-rich-text-editor/pull/423/commits/8180d8d1a3a92863137bdbead0df6c445c768b96)

crates/wysiwyg/src/composer_model/hyperlinks.rs
jmartinesp: I gave it a try too and It's way more complex than I thought, so no worries, this is fine.

crates/wysiwyg/src/composer_model/hyperlinks.rs
jmartinesp: I'd argue about it being simpler, it took me a while to understand why you were doing this. However, I think it's fine to leave it like this.

jmartinesp (approved):
LGTM!

PR #430, opened 2022-12-22
[iOS] Improved workflow

Velin92:
The workflow has been improved by pairing it with a workflow that is also present on the SwiftPM repo and by updating the release.md file

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -79,7 +79,7 @@ Remember to manually check the PR that will get opened on the [SWIFTPM Repo PR l
 
 Example: 
 1. The branch and PR name will always be something like release_v_X.Y.Z
-2. When clicking Merge customise the commit message to only include as title the same name release_v_X.Y.Z
+2. When clicking Merge customise the commit message to only include as title the same name release_v_X.Y.Z (substitute the "Merge pull request..." default commit message and put only the pr/branch name)
 3. This will trigger on merge a [workflow](https://github.com/matrix-org/matrix-wysiwyg-composer-swift/blob/main/.github/workflows/tag_on_release.yml) that will automatically tag main with the release value of X.Y.Z
 
 You can also manually do a revision release by just running `./release_ios.sh` on main which will not create a release PR but a revision PR that can have any commit message after the merge.
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -73,10 +73,15 @@ cd platforms/android && ./gradlew publish closeAndReleaseRepository`
 ```
   
 ### Swift/iOS:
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -73,10 +73,15 @@ cd platforms/android && ./gradlew publish closeAndReleaseRepository`
 ```
   
 ### Swift/iOS:
-This should be done automatically when a tag is uploaded however these are the steps to do it manually:
+When a tag is uploaded the `./release_ios.sh` will be called with the `-t <version>` (where `<version>` will be the provided tag).
```
jmartinesp, line 76: "the `./release_ios.sh` script" maybe?

RELEASE.md
```diff
@@ -73,10 +73,15 @@ cd platforms/android && ./gradlew publish closeAndReleaseRepository`
...
+This will open a PR against [the swift package repo](https://github.com/matrix-org/matrix-wysiwyg-composer-swift) with the latest changes.
+Remember to manually check and merge the PR that will get opened on the [SWIFTPM Repo PR list](https://github.com/matrix-org/matrix-wysiwyg-composer-swift/pulls), and provide as commit for the merge the same name of the branch/PR.
```
jmartinesp, line 78: This is a bit confusing. Does this mean that the commit name should be the name of the branch? Or that we should check that the right branches/tags are used in the PR?

jmartinesp (approved):
LGTM, just a couple of suggestions/questions for the README.

RELEASE.md
Velin92: just that the commit name should be the name of the branch (so that the tag will automatically be applied on main when the merge is done).
Any suggestions on how to make this less confusing?

jmartinesp: Maybe:

"Remember to manually check that the opened PR on the [SWIFTPM Repo](https://github.com/matrix-org/matrix-wysiwyg-composer-swift/pulls) exists and it contains the source branch name as the commit message, then merge it."

Velin92: But I feel in this way doesn't feel clear that we need to manually provide the commit message when merging to match the name of the branch.
What about...
"Remember to manually check the PR that will get opened on the [SWIFTPM Repo PR list](https://github.com/matrix-org/matrix-wysiwyg-composer-swift/pulls), and before merging it provide as commit message the name of the branch/PR."

jmartinesp: Sounds good to me. I guess it seems a bit confusing because you need to override the commit message.

Velin92: Yeah I know, would be way easier if the workflow could access not the commit message but the commit comment (since by the default is already the branch name), but seems that there is no way to do so

PR #433, opened 2022-12-23
Clippy fixes

alunturner:
Fixes all remaining clippy errors. Majority of the fixes are deriving Eq in various places, with a couple of default implementations and one error fixed with an explicit type definition.

I've made no changes besides what clippy suggested, and for the ones that may not be obvious what clippy was taking issue with, I've added explanatory comments to the PR.

All tests still passing.

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -13,6 +13,12 @@ pub struct ComposerModel {
     inner: Mutex<wysiwyg::ComposerModel<Utf16String>>,
 }
 
+impl Default for ComposerModel {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl ComposerModel {
     pub fn new() -> Self {
         Self {
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -9,14 +9,9 @@ use crate::ffi_link_actions::LinkAction;
 use crate::into_ffi::IntoFfi;
 use crate::{ActionState, ComposerAction};
 
+#[derive(Default)]
 pub struct ComposerModel {
     inner: Mutex<wysiwyg::ComposerModel<Utf16String>>,
-}
-
-impl Default for ComposerModel {
-    fn default() -> Self {
-        Self::new()
-    }
 }
 
 impl ComposerModel {
--- a/bindings/wysiwyg-wasm/src/lib.rs
+++ b/bindings/wysiwyg-wasm/src/lib.rs
@@ -73,14 +73,9 @@ impl IntoFfi for &HashMap<wysiwyg::ComposerAction, wysiwyg::ActionState> {
 }
 
 #[wasm_bindgen]
+#[derive(Default)]
 pub struct ComposerModel {
     inner: wysiwyg::ComposerModel<Utf16String>,
-}
-
-impl Default for ComposerModel {
-    fn default() -> Self {
-        Self::new()
-    }
 }
 
 #[wasm_bindgen]
--- a/bindings/wysiwyg-wasm/src/lib.rs
+++ b/bindings/wysiwyg-wasm/src/lib.rs
@@ -77,6 +77,12 @@ pub struct ComposerModel {
     inner: wysiwyg::ComposerModel<Utf16String>,
 }
 
+impl Default for ComposerModel {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 #[wasm_bindgen]
 impl ComposerModel {
     pub fn new() -> Self {
@@ -457,7 +463,7 @@ impl DomChildren {
     }
 
     // Clippy suggests that this name is ambiguous
-    pub fn next(&mut self) -> Option<DomHandle> {
+    pub fn next_child(&mut self) -> Option<DomHandle> {
         self.inner.pop_front()
     }
 }
--- a/bindings/wysiwyg-wasm/src/lib.rs
+++ b/bindings/wysiwyg-wasm/src/lib.rs
@@ -462,7 +462,6 @@ impl DomChildren {
         }
     }
 
-    // Clippy suggests that this name is ambiguous
     pub fn next_child(&mut self) -> Option<DomHandle> {
         self.inner.pop_front()
     }
--- a/crates/wysiwyg/src/dom/action_list.rs
+++ b/crates/wysiwyg/src/dom/action_list.rs
@@ -195,31 +195,31 @@ mod test {
     use super::*;
     use widestring::Utf16String;
 
-    type Dal = DomActionList<Utf16String>;
+    type ActionList = DomActionList<Utf16String>;
 
     #[test]
     fn default_creates_empty_action_list() {
-        let list = Dal::default();
+        let list = ActionList::default();
         assert!(list.actions.is_empty());
     }
 
     #[test]
     fn new_creates_action_list_with_passed_actions() {
         let actions = default_actions();
-        let list = Dal::new(actions.clone());
+        let list = ActionList::new(actions.clone());
         assert_eq!(actions, list.actions);
     }
 
     #[test]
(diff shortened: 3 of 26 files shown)


bindings/wysiwyg-wasm/src/lib.rs
```diff
@@ -457,7 +463,7 @@ impl DomChildren {
     }
 
     // Clippy suggests that this name is ambiguous
-    pub fn next(&mut self) -> Option<DomHandle> {
+    pub fn next_child(&mut self) -> Option<DomHandle> {
```
alunturner, line 465: Probably the only change with the potential for knock on effects. Could only find a single usage which is renamed in platforms/web/lib/dom.ts

crates/wysiwyg/src/dom/action_list.rs
```diff
@@ -195,31 +195,31 @@ mod test {
     use super::*;
     use widestring::Utf16String;
 
-    type DAL = DomActionList<Utf16String>;
+    type Dal = DomActionList<Utf16String>;
```
alunturner, line 198: Clippy doesn't like capitalized acronyms for some reason, even if you do it snake case like D_A_L

crates/wysiwyg/src/dom/dom_block_nodes.rs
```diff
@@ -86,7 +86,8 @@ where
                 last_ancestor_to_split.depth(),
             );
             let mut min_depth = 0;
-            for i in min_depth..max_depth {
+            let start_depth = min_depth;
+            for i in start_depth..max_depth {
```
alunturner, line 90: Here clippy didn't like the fact that the name of the variable changed on 91 is the same as that which is the start of this range. Just separated them into separate variables to get around that.

crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -759,12 +759,16 @@ where
...
+                let children_are_empty = container.children().is_empty();
+
+                let is_list_item_and_keep_empty_list_items =
+                    keep_empty_list_items && container.is_list_item();
+
+                let is_block_with_only_zwsp =
+                    container.is_block_node() && container.only_contains_zwsp();
+
+                if children_are_empty && !is_list_item_and_keep_empty_list_items
+                    || is_block_with_only_zwsp
```
alunturner, line 771: Here the issue was two branches both resolving in the same way. I tried to tidy this up using variables for the booleans because when I put all of the logic checks together in the if check (as they were) I didn't think it looked that good.

platforms/web/lib/dom.ts
```diff
@@ -65,7 +65,7 @@ export function refreshComposerView(
         list.className = `group_${idCounter % 10}`;
         const children = node.children(composerModel);
         let child: DomHandle | undefined;
-        while ((child = children.next())) {
+        while ((child = children.next_child())) {
```
alunturner, line 68: This is the renamed method in the web bindings

crates/wysiwyg/src/dom/dom_struct.rs
```diff
@@ -445,7 +445,7 @@ where
             let sub_handle = handle.sub_handle_up_to(i);
             if sub_handle.is_root() {
                 continue;
-            } else if let Some(node) =
+            } else if let Some::<&DomNode<S>>(node) =
```
alunturner, line 448: Not sure why but this was't coming back with the generic applied, it was being inferred as <DomNode<Unknown>>, which was causing an error due to a type mismatch when trying to assign to current.

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -13,6 +13,12 @@ pub struct ComposerModel {
     inner: Mutex<wysiwyg::ComposerModel<Utf16String>>,
 }
 
+impl Default for ComposerModel {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
```
Velin92, line 21: Actually to fix these I have a better/more clean suggestion... but it may take some time (not that much).
You can actually give as `derive` the `Default` trait. 
If you give the default trait as derive to all of the internal types of a type, then the type itself can also auto implement and derive default.

The idea is to give default to all of our types, so that we don't need to write any `impl` code to make Default work for that type

Velin92: Also while default is very cool, we should also use it, and change any instance of `Type::new()` with `Default::default` which is the generic function that will return the default impl

crates/wysiwyg/src/dom/action_list.rs
jonnyandrew: I agree `Dal` is a bit weird. We could also change the alias to something meaningful `Actions` or `ActionList`? Or get rid of the alias and create a local function that calls `DomActionList<Utf16String>::new`?

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
alunturner: I'll have a go, agree it would be cleaner, it does seem redundant having these blocks as I've added them

jonnyandrew (approved):

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
alunturner: @Velin92 What do you think to this? I've switched to trying to derive defaults where possible, the only place where I think it isn't clear what to do is in the DomNode, where I have had to implement it manually. I opted for a text node, felt it was a coin toss between that and a container node.

Velin92: i think it's good I am wondering if we really need then DomNode to implement default since maybe it should not have a default constructor?

alunturner: If DomNode doesn't implement a default, then it goes up the tree, so DomStruct can't derive a default, so ComposerState can't derive a default.

Maybe for ComposerState we stick to the implementation of default and just calling new()?

Velin92: Yeah I understand.
By checking DomStruct I see that default DomNode it may include is of Generic type, so maybe DomNode should be a generic container too as default value.
I guess that would make sense since it seems the most generic and defaulty of the options.

Velin92 (approved):
LGTM! Up to you how you want to solve the Default for DomNode in the end we are not really needing that one specifically.
Also all the `::new()` should be replaced with `::default` to take advantage of the Default impls. But I think is something that can be done in a separate PR

alunturner:
> LGTM! Up to you how you want to solve the Default for DomNode in the end we are not really needing that one specifically. Also all the `::new()` should be replaced with `::default` to take advantage of the Default impls. But I think is something that can be done in a separate PR, since it's not really a clippy fix but more an extra.

I think the simplest way to do this is to manually implement the Defaults as we had before the later changes, and put comments on them explaining why they're necessary. I think that proper implementation of default vs new should then be a separate PR as you mentioned. I may have a crack at that between Christmas and New Year

crates/wysiwyg/src/dom/action_list.rs
alunturner: Just noting for oversight over the Christmas-New year period:
This was updated and merged

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
alunturner: Just noting for oversight over the Christmas-New year period:
This was switched over to use the container node as the default type, and the implementation was copied from dom_struct.rs to make sure the default was set up correctly

alunturner:
Just noting for oversight over the Christmas-New year period:
I looked at the implementation of defaults and thought it was neater to amend the default dom node type to be the default (ie top level) container node type. The details of how to do this were lifted from dom_struct.rs so they match the existing behaviour. All tests still passing (plus we're not relying on this default behaviour at the moment) so code was merged based on the previous acceptance.

Will start work on another PR to move over to using this default behaviour and hopefully get that raised today.

PR #434, opened 2022-12-23
iOS: Snapshot testing

Velin92:
Added snapshot testing, since it will make testing for the appearence of complex attributed strings way easier when we will start implement complex blocks.
Also added some simple tests for starting.


aringenbach (approved):
LGTM

PR #436, opened 2022-12-28
Use defaults where appropriate

alunturner:
Recent work to fix clippy warnings and errors introduced manual defaults in the DomNode and ContainerNode files, as well as deriving defaults where possible.

This work uses the newly derived/specified defaults in the places I thought appropriate. To reduce repetition, it also adds an `add_children` method to tidy up creating a new container node (as it was effectively just creating a default but with a few children appended).

### DIFF
--- a/crates/wysiwyg/src/composer_state.rs
+++ b/crates/wysiwyg/src/composer_state.rs
@@ -32,9 +32,9 @@ where
 {
     pub fn new() -> Self {
         Self {
-            dom: Dom::new(Vec::new()),
-            start: Location::from(0),
-            end: Location::from(0),
+            dom: Dom::default(),
+            start: Location::default(),
+            end: Location::default(),
             toggled_format_types: Vec::new(),
         }
     }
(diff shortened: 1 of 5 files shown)


crates/wysiwyg/src/composer_state.rs
```diff
@@ -32,9 +32,9 @@ where
 {
     pub fn new() -> Self {
         Self {
-            dom: Dom::new(Vec::new()),
-            start: Location::from(0),
-            end: Location::from(0),
+            dom: Dom::default(),
+            start: Location::default(),
+            end: Location::default(),
```
alunturner, line 37: A `usize` defaults to 0 so I thought it made sense to use the default here.

I didn't replace any calls like `Location::from(0)` in tests because I thought that would detract from the clarity of the tests, whereas here `ComposerState::new()` is only ever called once, from base.rs, so thought it was (marginally) clearer that we're setting up the default state and default locations by writing it like this.

aringenbach (approved):
LGTM

PR #438, opened 2023-01-03
Fix special/extra character handling on Android

jonnyandrew:
## What?
- **Fix editor/model index mapping**
   Fixes the edge case where the cursor is at an extra new line character. In this case, the new calculation reflects that the extra character is neither preceding nor included in the selection.
- **Remove custom zero width space insertion**
   Fixes the handling of zero width space characters. Previously the Android code inserted these characters at the start of lists but this is now done by the internal Rust model.
   
## Why?
Zero width spaces are used by the editor to help display lists and new line characters are used by the internal model to distinguish whether the cursor selection is at the end of a list item or the start of the next.

This change updates the Android editor implementation to match the internal representation.

[`PSU-1012`](https://element-io.atlassian.net/browse/PSU-1012)

### DIFF
--- a/platforms/android/.idea/.name
+++ b/platforms/android/.idea/.name
@@ -1,1 +1,1 @@
-WYSIWYG
+Rich Text Editor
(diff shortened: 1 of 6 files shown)


aringenbach (approved):
LGTM

PR #439, opened 2023-01-03
Replace create empty list usage

aringenbach:
Remove dedicated empty list creation and use range-based route instead.

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -386,6 +386,9 @@ where
     /// Returns false if no updates was done.
     /// e.g. first text-like node is already a ZWSP.
     pub fn add_leading_zwsp(&mut self) -> bool {
+        // Note: handle might not be set in cases where we are transforming a node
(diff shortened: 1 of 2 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -384,17 +384,34 @@ where
 
     /// Add a leading ZWSP char to this container.
     /// Returns false if no updates was done.
-    /// e.g. first text-like node is a line break
-    /// or a text node that already starts with a ZWSP.
+    /// e.g. first text-like node is already a ZWSP.
     pub fn add_leading_zwsp(&mut self) -> bool {
+        let handle_is_set = self.handle().is_set();
```
jonnyandrew, line 389: What situations might the handle not already be set in?

jonnyandrew (approved):

crates/wysiwyg/src/dom/nodes/container_node.rs
aringenbach: https://github.com/matrix-org/matrix-rich-text-editor/pull/412#discussion_r1050751773

PR #441, opened 2023-01-04
Add dom method to extract list items

aringenbach:
Add a dom method to allow extracting list items out of lists.

Update: 
* Added multiple list items extractions variant + tests
* Replace the dedicated `remove_list` and the single list item variant
* Added a bit of documentation.
* Replaced one of the legacy list extraction code in `composer_model`

### DIFF
--- a/crates/wysiwyg/src/dom/dom_list_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_list_methods.rs
@@ -94,39 +94,6 @@ where
         self.join_nodes_in_container(&handle.parent_handle());
     }
 
-    pub fn extract_list_item(&mut self, handle: &DomHandle) {
-        let list_handle = handle.parent_handle();
-        let index_in_parent = handle.index_in_parent();
-        let list = self.parent_mut(handle);
-        let list_item = list.remove_child(index_in_parent);
-
-        let DomNode::Container(mut list_item) = list_item else {
-            panic!("List item is not a container")
-        };
-
-        if index_in_parent == 0 {
-            list_item.remove_leading_zwsp();
-            if list.children().is_empty() {
-                self.replace(&list_handle, list_item.take_children());
-            } else {
-                self.insert(&list_handle, list_item.take_children());
-            }
-        } else {
-            let mut to_insert = list_item.take_children();
-            let new_list_children = list.take_children_after(index_in_parent);
-
-            if !new_list_children.is_empty() {
-                let new_list = DomNode::new_list(
-                    list.get_list_type().expect("Node is not a list").clone(),
-                    new_list_children,
-                );
-                to_insert.push(new_list);
-            }
-
-            self.insert(&list_handle.next_sibling(), to_insert);
-        }
-    }
-
     pub fn extract_list_items(
         &mut self,
         handle: &DomHandle,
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/dom/dom_list_methods.rs
```diff
@@ -148,4 +261,11 @@ mod test {
         model.state.dom.remove_list(&first_handle);
         assert_eq!(tx(&model), text);
     }
+
+    // TODO: move this to a more globally usable location if needed
+    fn ds(dom: &Dom<Utf16String>) -> String {
```
jmartinesp, line 266: There is already a `restore_whitespace` fn in `testutils_composer_model.rs` that does a similar conversion (sorry, it's not the most intuitive name). We could replace it with this or viceversa?

jmartinesp (commented):
A small note and a question: should we just extract everything from the list items without adding extra line breaks? If we had:

```
<ul><li>List item</li></ul>Some text
```

I think extracting it get us:

```
List ItemSome Text
```

Unless I missed something?

crates/wysiwyg/src/dom/dom_list_methods.rs
aringenbach: Actually the equivalent of this is just the last step of `to_example_format` on the composer model (a.k.a `tx`). I set this temporarily but I think `dom` should have it's own `to_example_format` and it's own shortcut for the tests, as it's not really correct to use composer models on the DOM layer tests.

aringenbach:
> A small note and a question: should we just extract everything from the list items without adding extra line breaks? If we had:
> 
> ```
> <ul><li>List item</li></ul>Some text
> ```
> 
> I think extracting it get us:
> 
> ```
> List ItemSome Text
> ```
> 
> Unless I missed something?
> 
> EDIT: actually, given the current implementation with line breaks I guess we'd have this as a starting point:
> 
> ```
> <ul><li>List item</li></ul><br />Some text
> ```
> 
> Some maybe there's no need to add line breaks at all.

Hum actually, not all cases work yet but what's expected I guess is to have a ZWSP after the list, so if we extract the last item of a list I would expect to lookup for that ZWSP node and replace it with a line break.

jmartinesp (approved):
LGTM!

crates/wysiwyg/src/dom/dom_list_methods.rs
```diff
@@ -94,6 +104,13 @@ where
         self.join_nodes_in_container(&handle.parent_handle());
     }
 
+    /// Extract items from the list at given handle and positions
+    /// and move them appropriately into the DOM. Extracted list
+    /// items are separated by line breaks.
+    ///
+    /// * `handle` - the list handle.
+    /// * `start` - child index at which the extraction should start.
```
jmartinesp, line 94: I personally find `start` a bit confusing, since I automatically associate it with the `start, end` parameters we usually use for selections. I'd change it to something like `from_child_index`, but that's probably subjective, it's ok to leave it as is.

jmartinesp (approved):
LGTM, just a minor remark about naming, but it's just my personal opinion, so feel free to ignore it. 

Great job unifying all those different methods!

crates/wysiwyg/src/dom/dom_list_methods.rs
aringenbach: Yeah I wanted to change that too actually, I was thinking `start_index` before but it didn't felt like it was making things more clear. `from_child_index` makes some lines a bit long 😅 so I went with `child_index`, I think it's quite clear it's a position from which we are starting the operation (and if it's not, the documentation says so 🙂 ).

PR #442, opened 2023-01-04
Bindings & fix corner cases in block nodes (code block, quotes) with new lines

jmartinesp:
This PR:

* Adds FFI bindings for quotes and code blocks so they can be integrated into mobile clients.
* Fixes an issue with adding text between line breaks/zwsp nodes that appeared while testing block nodes.
* Fixes several corner cases with block nodes and line breaks.

### DIFF
--- a/crates/wysiwyg/src/composer_model/quotes.rs
+++ b/crates/wysiwyg/src/composer_model/quotes.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::dom::nodes::dom_node::DomNodeKind::{Generic, Quote};
+use crate::dom::DomLocation;
 use crate::{
     ComposerAction, ComposerModel, ComposerUpdate, DomHandle, DomNode,
     UnicodeString,
@@ -35,7 +36,19 @@ where
         let Some(wrap_result) = self.state.dom.find_nodes_to_wrap_in_block(s, e) else {
             // No nodes to be wrapped found.
             // Adding an empty Quote block with an a single ZWSP
-            self.state.dom.append_at_end_of_document(DomNode::new_quote(vec![DomNode::new_text(S::zwsp())]));
+            let range = self.state.dom.find_range(s, e);
+            let leaves: Vec<&DomLocation> = range.leaves().collect();
+            if leaves.is_empty() {
+                self.state.dom.append_at_end_of_document(DomNode::new_quote(vec![DomNode::new_zwsp()]));
+            } else {
+                let first_leaf_loc = leaves.first().unwrap();
+                let insert_at = if first_leaf_loc.is_start() {
+                    first_leaf_loc.node_handle.next_sibling()
+                } else {
+                    first_leaf_loc.node_handle.clone()
+                };
+                self.state.dom.insert_at(&insert_at, DomNode::new_quote(vec![DomNode::new_zwsp()]));
+            }
             self.state.start += 1;
             self.state.end += 1;
             return self.create_update_replace_all();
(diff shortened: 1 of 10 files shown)


crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -440,6 +448,33 @@ where
         action_list
     }
 
+    // FIXME: this is a workaround to avoid inserting the same text twice when selection is between 2 line breaks/ZWSP nodes.
```
jmartinesp, line 451: Honestly, this is a very ugly workaround, but I couldn't think of any other solution that wouldn't require a whole re-write of this function. All suggestions for a better replacement are welcome.

crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -726,20 +761,24 @@ where
         if (cur_handle == *from_handle
             || (from_handle.is_ancestor_of(&cur_handle)
                 && cur_handle.index_in_parent() == 0))
-            && (1..text_node.data().chars().count()).contains(&start_offset)
+            && (1..text_node.data().chars().count() + 1).contains(&start_offset)
```
alunturner, line 764: Could this be written as `..=`? I think this would achieve the aim and having struggled with off by one errors in the past, having fewer +/- 1s in the codebase is an improvement in my eyes

crates/wysiwyg/src/composer_model/quotes.rs
```diff
@@ -35,7 +36,19 @@ where
         let Some(wrap_result) = self.state.dom.find_nodes_to_wrap_in_block(s, e) else {
             // No nodes to be wrapped found.
             // Adding an empty Quote block with an a single ZWSP
-            self.state.dom.append_at_end_of_document(DomNode::new_quote(vec![DomNode::new_text(S::zwsp())]));
+            let range = self.state.dom.find_range(s, e);
+            let leaves: Vec<&DomLocation> = range.leaves().collect();
+            if leaves.is_empty() {
+                self.state.dom.append_at_end_of_document(DomNode::new_quote(vec![DomNode::new_zwsp()]));
```
alunturner, line 42: Perhaps not a change for this work, and appreciate that this function wasn't added in this work, but do you think that the `DomNode::new_quote()` function should do a check to see whether the first child is a zwsp and then add it if it's missing?

It seems to me like this could potentially be very easy to miss (or might require a check of the argument before you call the function if you weren't sure whether or not there was a zwsp present).

jmartinesp: Actually, that's such a simple solution that I feel quite dumb for not coming up with it 😅 . It's definitely worth implementing it.

jmartinesp: Hmm... it might not be possible, since we actually need to also modify the current selection (which is in the `ComposerState`) and this method has no access to it.

jmartinesp: I also tried asserting that the 1st child is always a ZWSP, but there are some points where we're creating/merging block nodes and it still doesn't have any children. Since we're probably moving away from using ZWSP sooner than later, I think it's probably not worth it to re-write all those calls.

alunturner: Sounds reasonable, agree that it would make more sense if we knew we were going to be sticking with the approach and it's only a marginal gain anyway, not having to write it manually at the moment

alunturner:
Code looks good to me, approved based on that, but one of the added tests isn't happy. Could we get that passing prior to merge?

 double_enter_in_code_block_when_empty_removes_it_and_adds_new_line is the bad one (according to the last test run)

alunturner (approved):

jmartinesp:
> Code looks good to me, approved based on that, but one of the added tests isn't happy. Could we get that passing prior to merge?
> 
> `double_enter_in_code_block_when_empty_removes_it_and_adds_new_line` is the bad one (according to the last test run)

Yep, it's fixed now. I tried to be too clever while simplifying the simple "add a '\n' char" behaviour and I ended up having infinite recursive calls to the same function, and for some reason my IDE marked it as ignored instead of failing so I just saw "all tests passed, some ignored".

PR #443, opened 2023-01-04
Include Espresso only in Android test configurations

jonnyandrew:
## What?
Include Espresso only in test configuration.

## Why?

Espresso shouldn't be needed outside of test builds since it is a test framework.

It creates dependency resolution issues in consumers of the library e.g.
- https://github.com/vector-im/element-android/pull/7883
- https://github.com/vector-im/element-android/actions/runs/3833639242/jobs/6525294449#step:9:1915


jmartinesp (approved):
LGTM!

PR #446, opened 2023-01-04
Add remove link and get link methods

florianduros:
Add bindings to the removeLinks method and an access to the current selected link


alunturner (approved):
Looks good to me

PR #447, opened 2023-01-04
Replace some legacy methods with dom methods

aringenbach:
* Removed another legacy function
* Add the capability for removing container that only contains a ZWSP node, instead of leaving them empty.
* Move slicing list item to dom methods

### DIFF
--- a/crates/wysiwyg/src/dom/dom_list_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_list_methods.rs
@@ -152,6 +152,31 @@ where
         }
         self.join_nodes_in_container(&handle.parent_handle());
     }
+
+    /// Slice list item at given handle and offset. A ZWSP node
+    /// is added at the beginning of the created list item.
+    ///
+    /// * `handle` - the list item handle.
+    /// * `offset` - offset at which the list item should be sliced
+    pub(crate) fn slice_list_item(
+        &mut self,
+        handle: &DomHandle,
+        offset: usize,
+    ) {
+        let list_item = self.lookup_node_mut(handle);
+        let mut slice = list_item.slice_after(offset);
+        slice.as_container_mut().unwrap().add_leading_zwsp();
+        let list = self.lookup_node_mut(&handle.parent_handle());
+        let DomNode::Container(list) = list else { panic!("List node is not a container") };
+        if slice.text_len() == 0 {
+            list.insert_child(
+                handle.index_in_parent() + 1,
+                DomNode::new_list_item(vec![DomNode::new_zwsp()]),
+            );
+        } else {
+            list.insert_child(handle.index_in_parent() + 1, slice);
+        }
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


crates/wysiwyg/src/dom/dom_list_methods.rs
```diff
@@ -152,6 +152,31 @@ where
...
+        let list = self.lookup_node_mut(&handle.parent_handle());
+        let DomNode::Container(list) = list else { panic!("List node is not a container") };
+        if slice.text_len() == 0 {
+            list.insert_child(
+                handle.index_in_parent() + 1,
+                DomNode::new_list_item(vec![DomNode::new_zwsp()]),
+            );
+        } else {
+            list.insert_child(handle.index_in_parent() + 1, slice);
+        }
```
alunturner, line 178: Are the `+1`s here because of the presence of ZWSP? 

Also as we're calling the same method on list (but with different second arguments) do you think it might be easier to read if we do something like (may be syntax errors here)
```
let child = if slice.text_len() != 0 { slice } else { DomNode::new_list_item(vec![DomNode::new_zwsp()]) }

...

list.insert_child(..., child);
``` 

Maybe you think that's less clear, could just be a personal preference

alunturner:
Just the one comment which isn't a required change. Always like PRs that reduce the amount of code!

alunturner (approved):

crates/wysiwyg/src/dom/dom_list_methods.rs
aringenbach: Thanks, I actually just removed this condition by making the `container_node` method for adding ZWSP work a bit better. 
The +1 is because we want to insert the new list item after the current one.

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -405,39 +405,42 @@ where
...
                 }
                 true
             }
             DomNode::LineBreak(_) => {
-                if handle_is_set {
-                    self.insert_child(0, DomNode::new_zwsp());
-                } else {
-                    self.children.insert(0, DomNode::new_zwsp());
-                }
+                insert_zwsp(self, handle_is_set);
```
alunturner, line 443: Again not a necessary change, but I _think_ we could use a match guard here to help clarify what branches result in the same control flow.

Disclaimer, this may not work. Could we do something like:

```
DomNode::Text(t) if t.data.is_empty => ...,
DomNode::Text(_) | DomNode::Linebreak(_) => {
    insert_zwsp(...);
    true
}

```

Think it would help cut down on a level of indentation and make it clearer that in non empty text cases it's treated the same as a linebreak.

alunturner (approved):

crates/wysiwyg/src/dom/nodes/container_node.rs
aringenbach: Yes, definitely. I guess I'm not used yet to work with conditional cases on Rust matches 😅

PR #449, opened 2023-01-05
Make Android bullet size configurable

jonnyandrew:
## What?

Make the bullet size configurable via XML layout attributes.

Note that I've replaced the default `BulletSpan` with a custom `UnorderedListSpan` because the `BulletSpan` did not allow customising the bullet radius without setting a color (which might be different to the text color).

## Why?

This will allow the Element app to match the style of bullet points as they appear in the timeline.

Following design feedback on
- https://github.com/vector-im/element-android/pull/7887

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
@@ -11,6 +11,8 @@ import android.text.Selection
 import android.text.Spannable
 import android.text.SpannableStringBuilder
 import android.util.AttributeSet
+import android.util.TypedValue.COMPLEX_UNIT_DIP
+import android.util.TypedValue.applyDimension
 import android.view.KeyEvent
 import android.view.MotionEvent
 import android.view.inputmethod.EditorInfo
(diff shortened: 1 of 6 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -55,7 +65,31 @@ class EditorEditText : TextInputEditText {
...
+                    getDimension(R.styleable.EditorEditText_bulletGap, bulletGapWidth)
+                bulletRadius = getDimension(R.styleable.EditorEditText_bulletRadius, bulletRadius)
+            } finally {
+                styleConfig = StyleConfig(
+                    bulletGapWidth = bulletGapWidth,
+                    bulletRadius = bulletRadius,
+                )
+                recycle()
+            }
+        }
```
jmartinesp, line 91: As this might easily grow, it might be a good idea to turn it into a separate helper class when we have to re-visit it to add further customisation?

jmartinesp (approved):
LGTM! I'm not a fan of also exposing those through XML, but I really like the idea of having a `StyleConfig` that can be passed to the spans parser, we should try to make more spans configurable using that in the near future 👍 .

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
jonnyandrew: Good point - it is a bit lengthy already. I'd agree that it's worth refactoring on the next pass.

PR #453, opened 2023-01-05
Fix creating lists with a leading line break

### LINKED ISSUE #452: [Lists] Unexpected behaviour starting list after linebreak
Found when implementing lists in web.

Writing something, inserting a linebreak, then clicking on list puts a linebreak in the first list item, making it appear that the cursor is not in the list item.

Recording demonstrating this can be seen here:

https://user-images.githubusercontent.com/56027671/210814607-aa752815-66b8-42d4-95a5-a746a631d739.mov

Example test to recreate this:

```
#[test]
fn new_list_after_linebreak() {
    let mut model = cm("start|");
    model.enter();
    model.unordered_list();

    assert_eq!(tx(&model), "start<ul><li>~|</li></ul>");
    // this will fail, result is "start<ul><li>~<br />|</li></ul>"
}
```

aringenbach:
Fixes an issue with creating lists with a leading line break. 
Line break is now replaced by the ZWSP node and doesn't end up inside the list item.
Closes #452 

Note: there is still an issue when removing the list afterwards where in some cases a line break should technically be re-inserted (e.g.: when the previous content is not a structure node - but this case won't exist with paragraphs though)

### DIFF
--- a/crates/wysiwyg/src/composer_model/lists.rs
+++ b/crates/wysiwyg/src/composer_model/lists.rs
@@ -248,7 +248,10 @@ where
         // strictly after the first char of the range (or if we are creating a list from an
         // empty text node, in that case the first node text_len is 0). If a leading ZWSP
         // is already present, e.g. the node following another list both corrections are 0.
-        if first_node.has_leading_zwsp() {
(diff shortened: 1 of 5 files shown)


alunturner (approved):
Looks good

PR #454, opened 2023-01-05
Add inline code formatting on Android

jonnyandrew:
## What?
Proof-of-concept for inline code formatting based on the official Google sample.
- [Drawing a rounded corner background on text](https://medium.com/androiddevelopers/drawing-a-rounded-corner-background-on-text-5a610a95af5)
- [Google sample on Github](https://github.com/googlearchive/android-text/tree/996fdb65bbfbb786c3ca4e4e40b30509067201fc/RoundedBackground-Kotlin)

The previous implementation used a simple background color (with sharp corners). This implementation achieves the rounded corners by drawing on the background as the standard Android spans are a bit limiting in what they can draw. 

Note that the original author recommends an additional performance optimisation:
> If you’re planning on integrating this implementation in your app, we strongly suggest to modify it and cache the calculations done in [TextRoundedBgHelper](https://github.com/googlesamples/android-text/blob/master/RoundedBackground-Kotlin/lib/src/main/java/com/android/example/text/styling/roundedbg/RoundedBgTextView.kt#L52). Then, make sure you invalidate the cache whenever anything related to text, like text color, size or other properties, has changed.
 
## Why?
[`PSU-1062`](https://element-io.atlassian.net/browse/PSU-1062)

Inline code should be displayed using a colored background with rounded corners.

## Screenshots
 
[inline-code.webm](https://user-images.githubusercontent.com/4940864/210825047-62e6a470-8680-4402-81ef-345b45c875ea.webm)


platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/fakes/FakeStyleConfig.kt
```diff
@@ -1,8 +1,16 @@
 package io.element.android.wysiwyg.fakes
 
+import android.content.Context
+import android.graphics.drawable.ColorDrawable
 import io.element.android.wysiwyg.utils.StyleConfig
 
-internal val fakeStyleConfig = StyleConfig(
+internal fun createFakeStyleConfig(context: Context) = StyleConfig(
```
jmartinesp, line 7: Is `Context` really needed here?

platforms/android/library/src/main/java/io/element/android/wysiwyg/inlinebg/InlineBgHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    private fun calculateDrawPosition(layout: Layout, spanPosition: SpanPosition): DrawPosition {
+        val (spanStart, spanEnd) = spanPosition
+        val startLine = layout.getLineForOffset(spanStart)
+        val endLine = layout.getLineForOffset(spanEnd)
+
+        // start can be on the left or on the right depending on the language direction.
+        val startOffset = (layout.getPrimaryHorizontal(spanStart)
+                + -1 * layout.getParagraphDirection(startLine) * horizontalPadding).toInt()
```
jmartinesp, line 121: `-layout.getParagraphDirection(startLine) * horizontalPadding).toInt()`?

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/StyleConfig.kt
```diff
@@ -1,10 +1,17 @@
...
+import android.graphics.drawable.Drawable
 import androidx.annotation.Px
 
 internal data class StyleConfig(
     @Px
     val bulletGapWidth: Float,
     @Px
     val bulletRadius: Float,
+    val inlineCodeHorizontalPadding: Int,
+    val inlineCodeVerticalPadding: Int,
```
jmartinesp, line 12: Maybe we could annotate these with `@Dimension(unit=Dimension.DP)` to differentiate them from the ones with `@Px` above?

jmartinesp (approved):
It's quite a complex solution, but given that we have full control of both the EditText where these spans will be written, and the usage from the developer's perspective is exactly the same as with any other span, LGTM!

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/StyleConfig.kt
jonnyandrew: Actually these are also pixel values (just rounded to the nearest integer): [`b0388b0`](https://github.com/matrix-org/matrix-rich-text-editor/pull/454/commits/b0388b0544249de8dd8f1841340062392aeb3b1a)

platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/fakes/FakeStyleConfig.kt
jonnyandrew: Thanks, I forgot to remove this! [`f5db458`](https://github.com/matrix-org/matrix-rich-text-editor/pull/454/commits/f5db458f1cc775f3b9b7ccb025459fdc7552f571)

platforms/android/library/src/main/java/io/element/android/wysiwyg/inlinebg/InlineBgHelper.kt
jonnyandrew: [`b0388b0`](https://github.com/matrix-org/matrix-rich-text-editor/pull/454/commits/b0388b0544249de8dd8f1841340062392aeb3b1a)

PR #461, opened 2023-01-09
Add parsing of code blocks and quotes to the WASM HTML parser

jmartinesp:
## What?

Allows the WASM HTML parser to parse also `pre` and `blockquote` tags.

## Why?

The WASM target uses a different parser, and this code wasn't included when we created the PR for code blocks and quotes.


aringenbach (approved):
LGTM

PR #462, opened 2023-01-09
Remove Swift Package .github and README from rsync

aringenbach:
* Remove files that are specific to the Swift Package repository
* Stop syncing/mirroring them on release

These will be maintain directly in the Swift Package repository if needed.


jmartinesp (approved):
LGTM!

PR #468, opened 2023-01-11
Call inputEventProcessor on keyDown

florianduros:
The purpose is to call the inputEventProcessor on keyDown. I need in element-web to stop the propagation of the `Enter` event in order to avoid the editor to grow (new line).

Enter should send the message and not add a new line.

### DIFF
--- a/platforms/web/lib/composer.ts
+++ b/platforms/web/lib/composer.ts
@@ -20,17 +20,18 @@ import {
     InputEventProcessor,
     Wysiwyg,
     FormattingFunctions,
+    WysiwygEvent,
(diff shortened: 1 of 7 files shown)


alunturner:
The code looks good to me, do we need to add tests here around the enter behaviour? Enter may send the message, or may insert a new line, depending on the user settings.

florianduros:
@alunturner The behaviour around enter and send message is in the `react-sdk`, here we just add a kind of hook in the keyDown behaviour (we have the same behaviour on input event)

alunturner (approved):

PR #470, opened 2023-01-11
Fix link structure

jonnyandrew:
## What?

Previously, when creating a link, the resulting structure might look like so:
```
├>a "https://element.io"
│ └>"test_"
└>a "https://matrix.org"
  ├>                               // <<< Note the redundant container here
  │ └>"link"
  └>" test"
```

After this change, the structure should no longer contain the redundant container:

```
├>a "https://element.io"
│ └>"test_"
└>a "https://matrix.org"
  └> "link test"
```

So to fix this,
- Stop inserting redundant container nodes during the creation of links
- Add a check for the invariant that generic container nodes can only exist at the root
- Refactor `split_new_sub_trees` to return a `Dom` (rather than a `DomNode`), so that the existence of a root node is implicit

[`PSU-1067`](https://element-io.atlassian.net/browse/PSU-1067)

## Why?

To keep the document representation minimal.

### DIFF
--- a/crates/wysiwyg/src/dom/dom_invariants.rs
+++ b/crates/wysiwyg/src/dom/dom_invariants.rs
@@ -55,6 +55,7 @@ where
     pub(crate) fn assert_invariants(&self) {
         self.assert_no_empty_text_nodes();
         self.assert_no_adjacent_text_nodes();
+        self.assert_exactly_one_generic_container();
 
         // We probably want some more asserts like these:
         // self.assert_document_node_is_a_container();
@@ -99,13 +100,33 @@ where
             }
         }
     }
+
+    /// Check there is only one generic container and that it is the root node
+    #[cfg(any(test, feature = "assert-invariants"))]
+    fn assert_exactly_one_generic_container(&self) {
+        use super::nodes::ContainerNodeKind;
+
+        let generic_nodes = self
+            .iter_containers()
+            .filter(|n| matches!(n.kind(), ContainerNodeKind::Generic));
+        let handles = generic_nodes.map(|n| n.handle()).collect::<Vec<_>>();
+
+        if handles.len() > 1 {
+            let first = handles.into_iter().find(|h| !h.is_root());
+            panic!(
+                "More than one generic container node found. Handle: {:?}\n{}",
+                first.unwrap().raw(),
+                self.to_tree()
+            );
+        }
+    }
 }
 
 #[cfg(test)]
 mod test {
     use widestring::Utf16String;
 
-    use crate::dom::nodes::TextNode;
+    use crate::dom::nodes::{ContainerNode, TextNode};
     use crate::dom::Dom;
     use crate::{DomNode, InlineFormatType};
 
@@ -138,6 +159,19 @@ mod test {
                 vec![DomNode::Text(TextNode::from(Utf16String::from("a")))],
             ),
             DomNode::Text(TextNode::from(Utf16String::from("b"))),
+        ]);
+
+        dom.assert_invariants();
+    }
+
+    #[test]
+    #[should_panic(
+        expected = "More than one generic container node found. Handle: [1]"
+    )]
+    fn multiple_generic_containers_fails_invariants() {
+        let dom = Dom::new(vec![
+            DomNode::Text(TextNode::from(Utf16String::from("a"))),
+            DomNode::Container(ContainerNode::default()),
         ]);
 
         dom.assert_invariants();
--- a/crates/wysiwyg/src/dom/dom_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_methods.rs
@@ -472,21 +472,16 @@ where
         let right = clone.split_sub_tree_from(from_handle, offset, depth);
 
         // Remove unmodified children of the right split
-        // TODO: create a utility for this
-        let right = right.as_container().unwrap();
-        let mut right =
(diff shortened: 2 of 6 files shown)


jmartinesp (commented):
The code looks good overall, although it would be great to have some more comments around the tricky parts of `insert_parent`.

I was also wondering, since we now have `split_sub_tree_between(from_handle, start_offset, to_handle, to_offset, depth)`, couldn't we simplify this code quite a bit? That code actually extracts everything between those 2 handles into a new subtree and returns it, so maybe we could use that to extract what needs to be wrapped inside a link node?

crates/wysiwyg/src/dom/dom_invariants.rs
```diff
@@ -99,13 +100,33 @@ where
             }
         }
     }
+
+    /// Check there is only one generic container and that it is the root node
+    #[cfg(any(test, feature = "assert-invariants"))]
+    fn assert_exactly_one_generic_container(&self) {
```
jmartinesp, line 106: 👌

jonnyandrew:
> I was also wondering, since we now have `split_sub_tree_between(from_handle, start_offset, to_handle, to_offset, depth)`, couldn't we simplify this code quite a bit? That code actually extracts everything between those 2 handles into a new subtree and returns it, so maybe we could use that to extract what needs to be wrapped inside a link node?

@jmartinesp I think that's a good idea as that sounds like exactly what `insert_parent` is doing internally.

I tried this suggestion but found a few differences in the behaviour that made it incompatible in it's current form (examples below). So although I think it'd be a worthwhile improvement, can/should we align the following separately before we attempt a refactor?

### Example behaviour differences

#### A range that starts and ends in the same node is split once

```rust
    #[test]
    fn split_sub_tree_between_flat() {
        let mut model = cm("ABC|");
        let result = model.state.dom.split_sub_tree_between(
            &handle(vec![0]), // start handle ("ABC")
            1,                // start offset
            &handle(vec![0]), // end handle ("ABC")
            2,                // end offset
            0,
        );
        let result = Dom::new(result.into_container().unwrap().take_children());

        assert_eq!(result.to_html(), "B"); // actual: "BC"
        assert_eq!(model.state.dom.to_html(), "AC"); // actual: "A"
    }
```

#### End node is split in a different direction

```rust
    #[test]
    fn split_sub_tree_between_simple() {
        let mut model = cm("<b>ABCDE</b>XXX<i>VWXYZ</i>|");
        let result = model.state.dom.split_sub_tree_between(
            &handle(vec![0, 0]), // start handle ("ABCDE")
            2,                   // start offset
            &handle(vec![2, 0]), // end handle ("VWXYZ")
            3,                   // end offset
            0,
        );
        let result = Dom::new(result.into_container().unwrap().take_children());

        assert_eq!(result.to_html(), "<b>CDE</b>XXX<i>VWX</i>"); // actual: "<b>CDE</b>XXX<i>YZ</i>"
        assert_eq!(model.state.dom.to_html(), "<b>AB</b><i>YZ</i>"); // actual: "<b>AB</b><i>VWX</i>"
    }
```

#### Adjacent nodes could become empty

```rust
    #[test]
    fn split_sub_tree_between_ignores_adjacent_nodes() {
        let mut model = cm("X{<b>AB</b><i><u>C</u></i>}|<u>D</u>");

        let result = model.state.dom.split_sub_tree_between(
            &handle(vec![0]),    // start handle ("X")
            1,                   // start offset at the end of the node
            &handle(vec![3, 0]), // end handle ("D")
            0,                   // end offset at the start of the node
            0,
        );

        assert_eq!(result.to_html(), "<b>AB</b><i><u>C</u></i>"); // actual: "<b>AB</b><i><u>C</u></i><u>D</u>"
        assert_eq!(model.state.dom.to_html(), "X<u>D</u>"); // actual: "X<i></i>"
    }

    #[test]
    fn split_sub_tree_between_ignores_adjacent_nodes2() {
        let mut model = cm("X{<b>AB</b><i><u>C</u></i>}|<u>D</u>");

        let result = model.state.dom.split_sub_tree_between(
            &handle(vec![1, 0]),    // start handle ("AB")
            0,                      // start offset at the start of the node
            &handle(vec![2, 0, 0]), // end handle ("C")
            1,                      // end offset at the end of the node
            0,
        );

        assert_eq!(result.to_html(), "<b>AB</b><i><u>C</u></i>"); // actual: "<b>AB</b><i><u></u></i>"
        assert_eq!(model.state.dom.to_html(), "X<u>D</u>"); // actual: "X<i><u>C</u></i><u>D</u>"
    }
```

jmartinesp:
I see, I didn't realise these 2 differed. I agree then, we should probably align the 2 behaviours and then just pick one implementation.

jonnyandrew:
> it would be great to have some more comments around the tricky parts of insert_parent.

@jmartinesp I assume you mean the part where we're splitting the partially covered leaves? I've added a couple more inline comments there ([`ff1c497`](https://github.com/matrix-org/matrix-rich-text-editor/pull/470/commits/ff1c49792775218fc867a8687ca37151b32e4f2a)) but let me know if there's anything else I could cover.

jmartinesp (approved):
LGTM!

jonnyandrew:
Thanks @jmartinesp! I've also created a ticket for the refactoring work: [`PSU-1073`](https://element-io.atlassian.net/browse/PSU-1073).

PR #472, opened 2023-01-11
Paragraphs re-work

jmartinesp:
## What?

The progression of the decisions around line breaks are as follow:

1. In the beginning we used `<br />` tags to represent new lines as most of our nodes were inline ones and it seemed like the simplest solution at the time.
2. Since the introduction of lists, we realised that we’d have to add some extra offsets to represent intermediate positions between list items and before/after the list: to do that we started using zero-width spaces *(ZWSP)* characters to try to delimit that position and add an extra offset.
3. This however, led to some issues with mapping indexes on Android/iOS, and also increased complexity quite a bit since we now had to check for special characters inside the different text nodes.
4. Since using ZWSP characters was really complex, we tried using a ‘ZWSP node’ instead, which simplified some checks. This however, didn’t fix all issues with index mappings and we realised that we had add ZWSP to both the block nodes/lists and another one after it, which had its own issues (parsing, finding the right adjacent nodes when a ZWSP was present…).

None of these solutions actually worked, so we proposed a new approach: use paragraphs along with other block nodes instead of line breaks, remove the ZWSPs altogether from the codebase and add extra offsets where needed.

### Steps:

1. Replace line breaks with actual `<p>` tags as much as possible.
2. Try to remove all traces of ZWSP nodes/chars in our codebase.
3. Add an extra offset to the cursor position at the end of block nodes when they have a next sibling:

```
"Foo <pre>bar</pre>" // Doesn't need to add an extra offset at the end since there is no next sibling
"<pre>Foo</pre> bar" // Needs to add an extra offset there, as visually there will be a new line after the <pre> tag

// So, this cursor position is 3
"<pre>Foo|</pre> bar"
// And this one is 4
"<pre>Foo</pre>| bar"
```

This also has the extra benefit that we’ll always use inline nodes (formatting, links, text, etc.) with inline nodes and block nodes with block nodes (paragraphs, blockquotes, pre, lists), making it easier to make them work together.

## Why?

```
// Say we have a list like this:
<ul><li>A</li><li>B</li></ul>

// This cursor position
<ul><li>A|</li><li>B</li></ul>
// And this one
<ul><li>A</li><li>|B</li></ul>
 
// Are technically the same (1)

// The same happens to
<pre>Code|</pre>Text
// And
<pre>Code</pre>|Text
```
So we need a way to differentiate them.

## Work to do:

- [x] Add paragraph block nodes.
- [x] Use paragraphs instead of line break nodes as much as possible.
- [x] Update the way selection is read and written in the tests and `find_pos`.
- [x] Fix code blocks. 
- [x] Fix quotes.
- [x] Fix lists.
- [x] Fix indentation.
- [x] Make sure a container node only has either inline nodes or block nodes inside.
- [x] Fix selection mis-matches in some corner cases.
- [x] Fix backspacing.
- [x] Remove ZWSP nodes.


jmartinesp:
Status update:
* Indentation mostly works again now, but we might need to add some post-processing to either the parser or to `DomNode::new_list_item` to make sure that, if a block node is added to the list item, all the nodes in the ListItem will be converted into block nodes (wrapping those that are inline inside paragraphs).
* Backspacing is partly working, bu the code needs quite some love. I also simplified a bit how `replace_text` works when deleting nodes.

jmartinesp:
Status update:
* 15 failing tests, most of them need to be re-written since they still use the old way of adding new lines with `<br />` instead of paragraphs.
* Fixed almost all issues with backspacing that were found while testing.
* Still need to add the assertion/post-process to make sure inline nodes and block nodes don't get mixed inside a container node.
* Changed a bit how we calculate the `text_len` of block nodes to always apply the extra offset but only take into account for selections when it's actually needed.

To do:
* The code quality needs to be improved, as some code that was added for quick testing purposes might not be needed or placed in the right files at the moment.
* We still need to remove ZWSP nodes, they're no longer used anywhere as far as I can tell.
* Line break nodes won't be removed, but won't be able to add them in the near future since they'll be replaced by paragraphs. We might want to re-introduce them in the future, but they carry extra complexity so it's probably better to discourage its use at the moment.
* We'll need some extra work to re-integrate this with the different clients:
  * We'll need to re-write part of how HTML is rendered in Android, translating `<p>` into `\n`.
  * Possibly the same on iOS (@aringenbach?).
  * On Web, we'll have to change how we calculate the current cursor position inside the composer model by adding extra offsets for block nodes/elements (so, when we iterate past a paragraph, list item, code block, etc., we'll need to 'move' the cursor forward).

aringenbach (approved):
Great job! Overall code looks pretty good, I noted a couple minor issues I saw along the way. 
I suggest we try to integrate this as soon as possible (i.e. the moment every platform code is ready) so we can go for a test session and narrow down corner cases. 

I think I saw some conditions that are probably repeated somewhere else, we should probably try to improve the code base/write helpers later to avoid noise on this piece of work though.

crates/wysiwyg/src/composer_model/new_lines.rs
```diff
@@ -0,0 +1,423 @@
...
+                            );
+                        }
+                    } else {
+                        self.do_new_line_in_paragraph(
+                            first_leaf,
+                            block_location,
+                        );
+                    }
+                } else {
+                    self.do_new_line_in_paragraph(first_leaf, block_location);
```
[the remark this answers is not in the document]
jmartinesp: I see how we can merge the 2 nested `do_new_line_in_paragraph` inside the `if let` check, but I don't see how we can merge them also with the external `else` clause.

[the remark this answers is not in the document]
aringenbach: Oh yeah I guess the only way to do something like this would be to extract to a function and return when the action is done. We could do it in a match arm too I guess, but I'm sure everyone will hate something like this (😆 ): 

```
if let ... {
    if ... {
        do_something();
        break;
    }
}

do_default();
```

I guess repeating this only twice is decent enough

crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -38,9 +38,18 @@ where
         // push_state_to_history is called if we can indent
         let (s, e) = self.safe_selection();
         let range = self.state.dom.find_range(s, e);
-        if !range.locations.is_empty() && self.can_indent(&range.locations) {
+        let locations: Vec<DomLocation> = range
+            .locations
+            .into_iter()
+            .filter(|l| l.relative_position() != Before)
```
aringenbach, line 44: Is this filter correct ? Removing it doesn't break any test, but it fixes this: https://github.com/matrix-org/matrix-rich-text-editor/issues/506

crates/wysiwyg/src/composer_model/lists.rs
```diff
@@ -51,9 +60,16 @@ where
         // push_state_to_history is called if we can unindent
         let (s, e) = self.safe_selection();
         let range = self.state.dom.find_range(s, e);
-        if self.can_unindent(&range.locations) {
+        let locations: Vec<DomLocation> = range
+            .locations
+            .into_iter()
+            .filter(|l| l.relative_position() != Before)
```
aringenbach, line 66: Same comment as above (haven't checked if it breaks tests though)

aringenbach (commented):
Found an issue with indent/unindent

crates/wysiwyg/src/composer_model/lists.rs
jmartinesp: It was probably needed at the time, but the code for matching selections was changed at least 3 times, so it might no longer apply.

PR #475, opened 2023-01-12
Code blocks for web

alunturner:
This PR:

- adds a button for adding a code block
- adds a new action type for code blocks
- handles this new action type
- exports the rust function required to add code blocks

In addition it:

- adds two svgs, one to update the inline code button, one for the new code block button
- renames two existing svgs for consistency


florianduros (approved):
LGTM

PR #477, opened 2023-01-12
Enable live document tree and display it by default

aringenbach:
Display document tree by default and update it live

Simulator Screen Recording - iPhone 13 - 2023-01-12 at 17 37 47


Velin92 (approved):
nice, live updates for the tree!
LGTM!

PR #479, opened 2023-01-13
Removed vitest coverage in ci

florianduros:
I removed the coverage ci which were slowing done make sometimes the ci fails. I tried the `watch` argument at `false` without success.

The coverage is still exported to sonar, so available.

Add rust & cargo cache
Remove Build step which is already done in make web (Building WASM bindings)


jmartinesp (approved):
LGTM!

PR #481, opened 2023-01-13
Rust/http scheme link

florianduros:
Add `http scheme` for link without url base.

Add `mailto` to email address.

### DIFF
--- a/crates/wysiwyg/src/composer_model/hyperlinks.rs
+++ b/crates/wysiwyg/src/composer_model/hyperlinks.rs
@@ -46,7 +46,7 @@ where
 
     pub fn set_link_with_text(
         &mut self,
-        link: S,
+        mut link: S,
         text: S,
     ) -> ComposerUpdate<S> {
         let (s, mut e) = self.safe_selection();
@@ -58,17 +58,17 @@ where
         self.do_replace_text(text.clone());
         e += text.len();
         let range = self.state.dom.find_range(s, e);
-        let link = self.add_http_scheme(&link);
+        self.add_http_scheme(&mut link);
         self.set_link_range(range, link)
     }
 
-    pub fn set_link(&mut self, link: S) -> ComposerUpdate<S> {
+    pub fn set_link(&mut self, mut link: S) -> ComposerUpdate<S> {
         self.push_state_to_history();
         let (mut s, mut e) = self.safe_selection();
 
         let mut range = self.state.dom.find_range(s, e);
 
-        let link = self.add_http_scheme(&link);
+        self.add_http_scheme(&mut link);
         // Find container link that completely covers the range
         if let Some(link) = self.find_closest_ancestor_link(&range) {
             // If found, update the range to the container link bounds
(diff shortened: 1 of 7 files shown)


Velin92:
Wouldn't be better to default always to https instead of http? SonarCloud doesn't like that and I kinda agree, even if in the end an https website should always redirect to the secure protocol, it's also true that the reverse.
Maybe would be better to always redirect to a secure connection by default, and let then the website do its redirect.

florianduros:
@Velin92 make sense, I have made the changes

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -87,6 +91,32 @@ where
...
+
+                if is_email {
+                    new_link.insert(0, &S::from("mailto:"));
+                } else {
+                    new_link.insert(0, &S::from("https://"));
+                }
+                return new_link;
+            }
+            Err(_) => {
+                return new_link;
```
aringenbach, line 112: I don't think we really need to return something here, we will return the new_link in any cases at the end of the method anyway. 
It's fine to write something like `Ok(_) => {}` for the standard working case and have the match return nothing.

aringenbach (approved):
LGTM just a couple suggestions

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -87,6 +91,32 @@ where
...
+            }
```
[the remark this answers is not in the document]
florianduros: This is the list of all the errors:

```
    EmptyHost => "empty host",
    IdnaError => "invalid international domain name",
    InvalidPort => "invalid port number",
    InvalidIpv4Address => "invalid IPv4 address",
    InvalidIpv6Address => "invalid IPv6 address",
    InvalidDomainCharacter => "invalid domain character",
    RelativeUrlWithoutBase => "relative URL without a base",
    RelativeUrlWithCannotBeABaseBase => "relative URL with a cannot-be-a-base base",
    SetHostOnCannotBeABaseUrl => "a cannot-be-a-base URL doesn’t have a host to set",
    Overflow => "URLs more than 4 GB are not supported",
```

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -87,6 +91,32 @@ where
         self.create_update_replace_all()
     }
 
+    fn add_http_scheme<'a>(&'a mut self, link: &'a S) -> S {
+        let mut new_link = link.clone();
```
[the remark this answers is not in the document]
florianduros: @aringenbach I have pushed changes where I'm using a mutable reference instead of doing a cloning

crates/wysiwyg/src/composer_model/hyperlinks.rs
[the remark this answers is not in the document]
aringenbach: I wonder if some very specific emails could trigger something else than RelativeUrlWithoutBase 🤔

aringenbach (approved):
LGTM 

We can still easily add code later if we ever want to have some kind of user notification system when we detect an invalid URL

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -87,6 +91,26 @@ where
...
+
+        match Url::parse(str) {
+            Ok(_) => {}
+            Err(ParseError::RelativeUrlWithoutBase) => {
+                let is_email = EmailAddress::is_valid(str);
+
+                if is_email {
+                    link.insert(0, &S::from("mailto:"));
+                } else {
+                    link.insert(0, &S::from("https://"));
```
jonnyandrew, line 106: Should we add `http://` (rather than `https://`) and let sites do their own redirection to `https`? From testing it, this is the way browser URL bars handle input without a scheme.

Velin92: https://github.com/matrix-org/matrix-rich-text-editor/pull/481#issuecomment-1383878338

jonnyandrew: Sorry I hadn't seen this, but makes sense. Users can still manually specify `http://` if needed.

jonnyandrew (approved):

PR #486, opened 2023-01-16
Add inline code style helpers for Android

jonnyandrew:
### Allow apps to display the inline code style
Allow users of the library to display inline code in their apps using `InlineCodeSpan`. There are now two ways to do this:

- **`EditorStyledTextView`**
This convenience view allows apps to use and display inline code using `InlineCodeSpan` without overriding `View.onDraw`.
- **`SpanBackgroundHelperFactory`**
This can be used by apps to create the background renderer needed to display inline code. Apps can then implement their own custom text view (rather than being restricted to using or subclassing `EditorStyledTextView`).

### Fix inline code style at edges
Fix inline code background that was being clipped by the start or end of the text view. Note that the padding at the start or end of inline code is now limited by the bounds of the text view.

image

## Why?
[`PSU-1069`](https://element-io.atlassian.net/browse/PSU-1069)

After a message is sent or received, the Element app needs to display the new inline code style in the timeline.

### DIFF
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/fakes/FakeStyleConfig.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/fakes/FakeStyleConfig.kt
@@ -1,20 +1,29 @@
 package io.element.android.wysiwyg.fakes
 
 import android.graphics.drawable.ColorDrawable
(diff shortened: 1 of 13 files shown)


jmartinesp (approved):
LGTM! You implemented a few ideas I had in my mind but had no time to do: splitting StyleConfig, making the helpers lazy, and creating a base TextView component. Thanks!

PR #488, opened 2023-01-16
`get_link_action()` returns `CreateWithText` for blank selections

Velin92:
When the user selects a portion of text that only contains whitespace characters, line breaks or zwsp nodes, since a link would actually not be visible (and on some platforms not even tappable), we want to actually replace that selection with some new text, and replicate the exact same behaviour that happens when there is not selection.

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/text_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/text_node.rs
@@ -78,6 +78,12 @@ where
             .all(|c| matches!(c, ' ' | '\x09'..='\x0d'))
     }
 
+    pub fn is_blank_in_range(&self, range: Range<usize>) -> bool {
+        self.data[range]
+            .chars()
+            .all(|c| matches!(c, ' ' | '\x09'..='\x0d'))
+    }
+
     pub fn remove_trailing_line_break(&mut self) -> bool {
         if self.data.chars().last() == Some('\n') {
             self.data.pop_last();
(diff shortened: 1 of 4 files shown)


crates/wysiwyg/src/dom/nodes/text_node.rs
```diff
@@ -133,6 +139,8 @@ where
         self.data().len() != 0
     }
 
+    /// Check if the textNode contains only whitespaces
+
```
aringenbach, line 143: Seems misplaced

crates/wysiwyg/src/tests/test_get_link_action.rs
```diff
@@ -119,3 +119,67 @@ fn get_link_action_from_selection_that_contains_multiple_links_partially_in_diff
...
+#[test]
+fn get_link_action_on_blank_selection_with_different_types_of_whitespaces() {
+    let model = cm("test { \t \n \r }| test");
+    assert_eq!(model.get_link_action(), LinkAction::CreateWithText)
+}
+
+#[test]
+fn get_link_action_on_blank_selection_after_a_link() {
+    let model = cm("<a href=\"https://element.io\">test</a>{  }|");
+    // This is the correct behaviour because the end of a link should be considered part of the link itself
```
aringenbach, line 180: Actually I wonder if it should, if it's a zero-length selection then sure but when selecting more than that it might feel weird. 
Anyway I don't think making changes about how selections behave in the composer should be the purpose of this PR.

aringenbach (approved):
LGTM

crates/wysiwyg/src/dom/nodes/text_node.rs
Velin92: moved it

PR #496, opened 2023-01-18
[iOS] Integrate DTCoreText

aringenbach:
* Integrates DTCoreText as a replacer for standard Apple HTML attributed string builder
* Improve quote/code blocks display using dedicated layer drawing
* Extract HTMLParser + Tests to dedicated package
* Expand SnapshotTests and move them to a dedicated test suite


platforms/ios/lib/WysiwygComposer/Sources/DTCoreTextExtended/UIFont+AttributedStringBuilder.m
```diff
@@ -0,0 +1,81 @@
...
+}
+
+@end
+
+#pragma mark - Implementation
+
+@implementation UIFont(DTCoreTextFix)
+
+// DTCoreText iOS 13 fix. See issue and comment here: https://github.com/Cocoanetics/DTCoreText/issues/1168#issuecomment-583541514
+// Also see https://github.com/Cocoanetics/DTCoreText/pull/1245 for a possible future solution
```
Velin92, line 61: actually we are not using iOS 13 as min target anymore on element-ios, so this may not be required. And we may update the library too to support iOS 14.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
```diff
@@ -17,13 +17,13 @@
 import Foundation
 
 /// Describe an eror occuring during HTML to attributed ranges manipulation.
-enum AttributedRangeError: LocalizedError, Equatable {
+public enum AttributedRangeError: LocalizedError, Equatable {
```
Velin92, line 20: why public?

aringenbach: We do use these for logging purposes on the lib which is now in a separated package target

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/BackgroundStyle.swift
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import UIKit
+
+internal enum BackgroundStyle {
```
Velin92, line 19: guess you can omit internal.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/BackgroundStyle.swift
```diff
@@ -0,0 +1,56 @@
...
+    var tempColor: UIColor {
+        switch self {
+        case .inlineCode:
+            return UIColor.blue
+        case .codeBlock:
+            return UIColor.green
+        case .quote:
+            return UIColor.red
+        }
+    }
```
Velin92, line 33: what is the purpose of the tempColor?

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParser.swift
```diff
@@ -0,0 +1,127 @@
...
+            in: NSRange(location: 0, length: mutableAttributedString.length)
+        ) { value, range, _ in
+            if value != nil {
+                mutableAttributedString.removeAttribute(.underlineStyle, range: range)
+                mutableAttributedString.removeAttribute(.underlineColor, range: range)
+                mutableAttributedString.addAttributes([.foregroundColor: style.linkColor], range: range)
+            }
+        }
+
+        mutableAttributedString.applyBackgroundStyles()
```
Velin92, line 117: not a fan of this because... it uses the style of the HTMLParser behind curtains in its own implementation, I feel this function should have the style passed directly, and become `applyBackgroundStyles(style: HTMLParserStyle)`.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/BackgroundStyle.swift
```diff
@@ -0,0 +1,56 @@
...
+
+    var borderColor: UIColor {
+        switch self {
+        case .inlineCode, .codeBlock:
+            return HTMLParser.style.codeBorderColor
+        case .quote:
+            return HTMLParser.style.quoteBorderColor
+        }
+    }
```
Velin92, line 55: not a fan of this because it creates a strong dependency between this enum and the HTMLParser which is a static function... maybe... these could be converted to function, and have the style as a parameter.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSMutableAttributedString.swift
```diff
@@ -0,0 +1,50 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import DTCoreText
+import UIKit
+
+extension NSMutableAttributedString {
+    /// Sets the background style for detected quote & code blocks within the attributed string.
+    func applyBackgroundStyles() {
```
Velin92, line 22: this function could take the style as a parameter

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/UITextView.swift
```diff
@@ -33,8 +33,15 @@ extension UITextView {
...
             self.selectedRange = content.selection
+
+            self.drawBackgroundStyleLayers()
         }
     }
+
+    override open func draw(_ rect: CGRect) {
+        super.draw(rect)
+        drawBackgroundStyleLayers()
+    }
```
Velin92, line 44: I would move this draw override to our custom `PlaceholdableTextView` and maybe change its name to WysiwygTextView or something like that. Extending the textview openly here in a library could create undesired or unneeded effects on the apps that are going to use it.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/BackgroundStyle.swift
Velin92: Or maybe instead of being an enum the BackgroundStyle could become a struct that is initialised by passing the HTMLParserStyle.
So the applyBackgroundStyle function can initialise the struct.
And then the layer can just use the BackgroundStyle struct directly.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/UITextView.swift
Velin92: I would still make `drawBackgroundStyleLayers()` public, so that we can still use this (and we may want it for the timeline on element-ios maybe?) to use the custom background style inside only the specific text views that we want, by overriding only those.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/BackgroundStyle.swift
aringenbach: Added comment

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/BackgroundStyle.swift
aringenbach: Fixed

Velin92 (approved):
LGTM

PR #497, opened 2023-01-18
Codeblock button state change

alunturner:
Small change to make firing `code_block` disable a few more options.

I've removed a test as I thought that we shouldn't be able to put code blocks inside quotes from my reading of the comments, but this might have been a mistake.

Adds a test to verify the new behaviour

### DIFF
--- a/crates/wysiwyg/src/composer_model/quotes.rs
+++ b/crates/wysiwyg/src/composer_model/quotes.rs
@@ -312,13 +312,6 @@ mod test {
     }
 
     #[test]
-    fn remove_quote_containing_code_block() {
-        let mut model = cm("<blockquote>~<pre>~Some| code</pre></blockquote>");
-        model.quote();
-        assert_eq!(tx(&model), "<pre>~Some| code</pre>");
-    }
-
-    #[test]
     fn create_and_remove_quote() {
         let mut model = cm("|");
         model.quote();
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/composer_model/quotes.rs
```diff
@@ -311,13 +311,6 @@ mod test {
         );
     }
 
-    #[test]
-    fn remove_quote_containing_code_block() {
-        let mut model = cm("<blockquote>~<pre>~Some| code</pre></blockquote>");
-        model.quote();
-        assert_eq!(tx(&model), "<pre>~Some| code</pre>");
-    }
-
```
alunturner, line 320: Removed this as I didn't think this would be allowed from our conversation today @jmartinesp . Let me know if I'm wrong here.

jmartinesp: A quote can contain a code block, but not the other way around:

```
<blockquote><p>This will become code|</p></blockquote>
// Becomes
<blockquote><pre>This will become code|</pre></blockquote>
```

jmartinesp: Oh, never mind 🤦 . Given this change, it's impossible to do this without removing the code block first or creating a new paragraph outside of it. I think that there should be a way to do this, but we haven't really explored these cases so this is fine.

jmartinesp (approved):
LGTM. 

Ideally, we'd also handle the case where we're inside a quote to enable the quote button again, but this might lead to issues in the future if done wrong, so it's better to think about all the implications before we try to implement it.

PR #500, opened 2023-01-19
Disable link in code block and add test

alunturner:
As per the title. Disables links inside codeblocks and adds a test for this.

### DIFF
--- a/crates/wysiwyg/src/composer_model/menu_state.rs
+++ b/crates/wysiwyg/src/composer_model/menu_state.rs
@@ -213,6 +213,7 @@ where
                 ComposerAction::OrderedList,
                 ComposerAction::UnorderedList,
                 ComposerAction::Quote,
+                ComposerAction::Link,
             ])
         }
         disabled_actions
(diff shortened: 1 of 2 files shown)


Velin92 (approved):

Velin92:
One quick question...

What if the selection spand from non codeblock, to include also partially or fully a codeblock? Is the link button still tappable?

Example
```
te{st <blockcode> test_block_code </blockcode> t}|est
te{st <blockcode> test_block}|_code </blockcode> test
```

@alunturner

Velin92 (approved):
LGTM

alunturner:
> One quick question...
> 
> What if the selection spans from a non codeblock node, to include partially or fully a codeblock? Is the link button still tappable?
> 
> Example
> 
> ```
> te{st <blockcode> test_block_code </blockcode> t}|est
> te{st <blockcode> test_block}|_code </blockcode> test
> ```
> 
> @alunturner
The link button is disabled in this case - tests have been expanded to cover this now

PR #501, opened 2023-01-19
Links will use the innermost approach when the selection contains block or structure nodes

Velin92:
This is a fix that uses a non optimal solution to deal with links breaking block nodes.
Essentially if the selection contains a non block node, we will use the old approach, the innermost one, where we add the link node directly to each single text node leaf.

This is not optimal, and I am already working with @jonnyandrew on a more optimal solution, but will at least fix the bug, make links usable alongside block nodes.


jmartinesp (approved):
LGTM!

PR #504, opened 2023-01-20
iOS: Paragraph support

Velin92:
- Fixed some weird interactions between DTCoreText and paragraphs
- Removed the old lists mapping used with ZWSP, it's not required anymore
- Updated most of the tests to account for the paragraph changes

However this PR does not include a fix for the issue of quote and code blocks not being rendered if a newline is not placed afterwards... looking into it right now, however this might need some changes in the Rust codebase if there is no way to fix this on iOS


Velin92:
@jmartinesp rebasing feature/paragraph with main should make this PR less big, the fact is that it include DTCoreText to fix the weird interactions it had with paragraphs

jmartinesp:
> @jmartinesp rebasing feature/paragraph with main should make this PR less big, the fact is that it include DTCoreText to fix the weird interactions it had with paragraphs

I merged `main` into `feature/paragraphs`, if that helps. It also fixed the weird CI issue with tests.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParser.swift
```diff
@@ -107,15 +104,29 @@ public final class HTMLParser {
...
+    
+    private static func removeTrailingNewlineIfNeeded(from mutableAttributedString: NSMutableAttributedString, given html: String) {
+        // DTCoreText always adds a \n at the end of the document, which we need to remove
+        // however it does not add it if </code> </a> are the last nodes.
+        // Also we don't want to remove if a blockquote or a codeblock have more than one paragraph
+        if mutableAttributedString.string.last == "\n",
+           !html.hasSuffix("</code>"),
+           !html.hasSuffix("</a>"),
+           !html.hasSuffix("</p><p></p></blockquote>"),
+           !html.hasSuffix("\n</pre>") {
```
jmartinesp, line 128: That's... oddly specific 😅 . What happens if the last paragraph in either the blockquote or codeblock is not empty, or if the previous item in a blockquote is a list?

Velin92: if the paragraph is not empty nothing will happen, this is just to prevent iOS from essentially not entering a newline when is actually needed.
Without this check is impossible to enter a newline after writing some stuff in a blockquote or a codeblock, because we are removing that same newline which is actually a needed one.

For the list case you're right let me check it!

Velin92: Yeah the list case had some issues! added it, also improved the comment to explain better the reasoning for the case in which we do not want to remove that \n

Velin92:
```
// DTCoreText always adds a \n at the end of the document, which we need to remove
// however it does not add it if </code> </a> are the last nodes.
 // Also we don't want to remove it if blockquote and codeblock contain that newline
// and are not empty, because DTCoreText does not add a newline if these blocks
// contain one at the end.
```

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
```diff
@@ -148,7 +132,7 @@ extension NSAttributedString {
                 }
             }
         }
-
+
```
aringenbach, line 135: Seems like you have one of the two `automatically trim whitespaces` options disabled ?

aringenbach (approved):
One small comment inline, otherwise LGTM.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
Velin92: yeah had the` including whitespace only lines` one disabled

PR #508, opened 2023-01-24
Fix selection for indent and unindent actions.

### LINKED ISSUE #506: [Indent/UnIndent] Can't indent on last char of a list item
Failing test (despite indent button/action being marked as available): 

```rust
#[test]
fn can_indent_on_last_char_of_list_item() {
    let mut model = cm("<ol><li>abc</li><li>def|</li></ol>");
    model.indent();
    assert_eq!(
        tx(&model),
        "<ol><li><p>abc</p><ol><li>def|</li></ol></li></ol>",
    )
}
```

If moving the cursor one character to the left it does work.

jmartinesp:
## What?

Removes the `location.relative_position() != Before` check for selection results when using indent & unindent. It also improves the way `can_indent` and `can_unindent` work.

## Why?

Fixes #506 .

### DIFF
--- a/crates/wysiwyg/src/composer_model/lists.rs
+++ b/crates/wysiwyg/src/composer_model/lists.rs
@@ -38,13 +38,8 @@ where
         // push_state_to_history is called if we can indent
         let (s, e) = self.safe_selection();
         let range = self.state.dom.find_range(s, e);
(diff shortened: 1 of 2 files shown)


aringenbach (approved):
Thanks, LGTM

PR #509, opened 2023-01-24
Remove FIXME and restore working UI tests

aringenbach:
* Removes FIXME from old broken cases
* Updates UI tests to use paragraph rework


Velin92 (approved):
LGTM

PR #511, opened 2023-01-24
Empty paragraphs will return NBSP inside of them

Velin92:
This should allow DTCoreText and Web to easily parse empty paragraphs as new lines, however this character should then be removed on the client.


aringenbach (commented):
Code LGTM. 

I wonder though if we should do this like that if it's not needed on all platforms (especially depending on if it's really needed on Web or not)
I suppose we could just have some pre-processing on the HTML content on the iOS side as well (i.e. replace `<p></p>` with `<p>&nbsp;</p>` only right before the DTCoreText parsing)
A bit scared we might end up with some TextNode containing `&nbsp;` at some point while using e.g. edit.

PR #514, opened 2023-01-24
Improve quote/code blocks display upon applying content or on character input

aringenbach:
* Treat quote & code block on zero-length selection as pending formats
* Force redraw when applying content to text view


Velin92:
I found a bug on blockquotes when adding newlines, basically newlines cannot be added anymore.
I partially fixed it here: https://github.com/matrix-org/matrix-rich-text-editor/pull/515

However now consecutive newlines inside the blockquote will not be accepted (which was the whole point of using originally &nbsp;)

PR #515, opened 2023-01-24
Rendering improvement for codeblock and quoteblocks

Velin92:
- The pending format is also added when a newline is entered and either blockquote or codeblock are reversed (now the rendering is triggered as soon as a character is typed also on lines that are not the first)
- Fixed a bug, that deleted blockquotes last newline
- nbsp is now as static variable in String and Character

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParser.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParser.swift
@@ -133,9 +133,9 @@ public final class HTMLParser {
         if mutableAttributedString.string.last == "\n",
            !html.hasSuffix("</code>"),
            !html.hasSuffix("</a>"),
-           !html.hasSuffix("</p><p></p></blockquote>"),
-           !html.hasSuffix("</ul><p></p></blockquote>"),
-           !html.hasSuffix("</ol><p></p></blockquote>"),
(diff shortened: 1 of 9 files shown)


platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParser.swift
```diff
@@ -133,9 +133,9 @@ public final class HTMLParser {
         if mutableAttributedString.string.last == "\n",
            !html.hasSuffix("</code>"),
            !html.hasSuffix("</a>"),
-           !html.hasSuffix("</p><p></p></blockquote>"),
-           !html.hasSuffix("</ul><p></p></blockquote>"),
-           !html.hasSuffix("</ol><p></p></blockquote>"),
+           !html.hasSuffix("</p><p>\(Character.nbsp)</p></blockquote>"),
+           !html.hasSuffix("</ul><p>\(Character.nbsp)</p></blockquote>"),
+           !html.hasSuffix("</ol><p>\(Character.nbsp)</p></blockquote>"),
```
Velin92, line 138: @aringenbach this fixed a bug where newlines where not added anymore to blockquotes however... I think that actually the nbsp character should be removed from here, because we are now not able anymore to add more than one consecutive newline inside a blockquote

aringenbach (approved):
LGTM

PR #516, opened 2023-01-25
[iOS] Fix nbsp parsing with discardable text

aringenbach:
Should fix issue with nbsp parsing by marking it for removal instead of removing immediately.


Velin92 (approved):
LGTM!
Had to remove the check for removing the newline in case of blockquote containing an empty paragraph.
I guess with this implementation DTCoreText adds now in blockquotes the newline regardless and we can safely remove it all the time for those nodes.

PR #518, opened 2023-01-25
Paragraphs for web

alunturner:
This touches a bit more than I was expecting.

- rust changes allow us to handle the case where the user starts typing into a formatting node (on opening the editor) and then presses enter to move us into paragraphs, adds tests for these cases
- TS changes add some util functions to allow us to map the selection properly with the paragraph changes, and adds a chunk of new test cases
- new utils (post refactor) are currently not unit tested, which I'll be working on now but I think the PR is good to be looked at in the meantime (and tests will be incoming)

https://user-images.githubusercontent.com/56027671/214888239-4f6b1fb7-4f46-463b-a4f0-f7543902ad3c.mov

nb issue with the cursor inside the codeblocks has already been raised


florianduros:
To iterate of the node, the [createNodeIterator](https://developer.mozilla.org/fr/docs/Web/API/Document/createNodeIterator) can also be a good idea.

platforms/web/index.html
```diff
@@ -3,7 +3,7 @@
     <head>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-        <title>Wysisyg</title>
+        <title>Wysiwyg</title>
```
alunturner, line 6: This had been winding me up for a while and needed addressing before it became critical for me.

platforms/web/lib/dom.ts
```diff
@@ -16,18 +16,6 @@ limitations under the License.
...
-        return offset ?? node.textContent?.length ?? 0;
-    } else if (node.hasChildNodes()) {
-        return Array.from(node.childNodes)
-            .map((childNode) => computeSelectionOffset(childNode))
-            .reduce((prev, curr) => prev + curr, 0);
-    } else {
-        return 0;
-    }
-}
-
```
alunturner, line 30: Not sure about the history of this but this isn't used anywhere, so I deleted it. My guess is it was superseded by `computeNodeAndOffset` but never got removed.

platforms/web/lib/useWysiwyg.delete.test.tsx
```diff
@@ -77,8 +80,8 @@ describe('delete content', () => {
 
         // Then
         await waitFor(() => {
-            expect(textbox).toContainHTML('foar');
-            expect(textbox).toHaveAttribute('data-content', 'foar');
+            expect(textbox).toContainHTML('<p>foar</p>');
+            expect(textbox).toHaveAttribute('data-content', '<p>foar</p>');
         });
```
alunturner, line 85: No changes to these tests or behaviour, simply needed to update the fact they use `<br />` tags before and we've moved to paragraphs now

platforms/web/lib/useWysiwyg.formatting.test.tsx
```diff
@@ -62,7 +63,7 @@ describe.each([
         'reversed',
         '<del>foo</del>',
         'fo<del>o&nbsp;</del>bar',
-        '<del>fo</del>o<br/>b<del>ar</del>',
+        '<p><del>fo</del>o</p><p>b<del>ar</del></p>',
         '<del>fo</del>o <del>bar</del>',
```
alunturner, line 67: Updating to move away from line breaks and over to paragraphs

platforms/web/index.html
langleyd:
```suggestion
        <title>Rich Text Editor</title>
```
Wysiwyg is also not on fleek.
Fixed it for you. 😂

platforms/web/lib/dom.ts
```diff
@@ -389,3 +431,74 @@ export function countCodeunit(
...
+ * require an additional offset to match up with the rust model. We also need to
+ * handle the case where we have consecutive formatting nodes, as only the last
+ * of the formatting nodes needs the extra offset (if applicable).
+ *
+ * This is called in two places and only with TextNodes
+ *
+ * Returns a boolean, true if the node needs an extra offset
+ */
+
+function nodeNeedsExtraOffset(node: Node | null) {
```
alunturner, line 447: This could do with some unit tests

platforms/web/lib/dom.ts
```diff
@@ -316,9 +345,22 @@ function findCharacter(
...
             // this node.
+
+            // The extra check for an offset here depends on the ancestor of the
+            // text node and can be seen as the opposite to the equivalent call
+            // in computeNodeAndOffset
+            const shouldAddOffset = nodeNeedsExtraOffset(currentNode);
+            const extraOffset = shouldAddOffset ? 1 : 0;
+
+            // ...but also have a special case where we don't count a textnode
+            // if it is an nbsp, as this is what we use to mark out empty paras
```
jmartinesp, line 356: 'paras' -> 'paragraphs'?

platforms/web/lib/dom.ts
```diff
@@ -389,3 +431,74 @@ export function countCodeunit(
...
+        } else {
+            checkNode = checkNode.parentNode;
+        }
+    }
+    return false;
+}
+
+function isFormattingNode(node: Node | ParentNode | null) {
+    if (node === null) return false;
+    const formattingNodeNames = ['EM', 'U', 'STRONG', 'DEL'];
```
jmartinesp, line 484: I wonder if link nodes or inline code should also be included here, as they behave almost like formatting nodes.

jmartinesp (commented):
I mentioned a possible typo, and more importantly, a couple of cases (links and inline code) that we might have missed, I couldn't find any new tests for them in the TS code.

platforms/web/lib/dom.ts
alunturner: Interesting, I'll take a look at that.

If I expand this list of node names to include `a` and `code`, am I just at the point where I'm saying `!isCountainerNode(...)`. I'll have a look at the node types in the rust model and see if I can figure that out. If that's the case that could slim this PR down a bit.

florianduros (commented):
LGFM except the missing node (inline code..) raised by @jmartinesp

platforms/web/lib/dom.ts
alunturner: Tests added

platforms/web/lib/dom.ts
alunturner: This seems to be working fine without as it stands, although I can't test links in the example app at the moment. I'll bear this in mind when I'm integrating into element web when I play with links

jmartinesp (approved):
LGTM!

platforms/web/lib/dom.ts
alunturner: ... I take that back. Figured out how to break it. Have added the tags!

PR #520, opened 2023-01-25
iOS: Replace NBSP with ZWSP to allow block rendering

Velin92:


aringenbach (approved):
LGTM

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
```diff
@@ -95,9 +109,17 @@ extension NSAttributedString {
...
         }
 
+        let discardableTextRanges = discardableTextRanges()
+        var actualIndex = attributedIndex
+
+        for discardableTextRange in discardableTextRanges {
+            if discardableTextRange.upperBound <= attributedIndex {
+                actualIndex -= discardableTextRange.length
+            }
+        }
```
aringenbach, line 119: Think this raises a Swiftlint warning, I'm guessing we can e.g. write something like:

```swift
actualIndex -= discardableTextRanges
    .filter { $0.upperBound <= attributedIndex }
    .reduce(0) { $1 + $2.length }
```

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
```diff
@@ -117,8 +139,16 @@ extension NSAttributedString {
...
     /// - Returns: the index inside the attributed representation
     func attributedPosition(at htmlIndex: Int) throws -> Int {
+        let discardableTextRanges = discardableTextRanges()
+        var actualIndex = htmlIndex
+
+        for discardableTextRange in discardableTextRanges {
+            if discardableTextRange.location < actualIndex {
+                actualIndex += discardableTextRange.length
+            }
+        }
```
aringenbach, line 149: same here

but more generally I'm sure we can globally improve the code here, and perhaps merge list prefixes and discardable text into the same Array so we don't have multiple arrays that achieve same kind of computation

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSMutableAttributedString.swift
```diff
@@ -53,12 +53,18 @@ extension NSMutableAttributedString {
...
-    func removeDiscardableText() {
+    /// Finds any text that has been marked as discardable
+    /// and either replaces it with ZWSP if contained overlaps with text marked with a background style
+    /// or removes it otherwise
+    func replaceOrDeleteDiscardableText() {
         enumerateTypedAttribute(.discardableText) { (discardable: Bool, range: NSRange, _) in
             guard discardable == true else { return }
-
-            self.deleteCharacters(in: range)
+            let attributes = self.attributes(at: range.location, effectiveRange: nil)
```
aringenbach, line 62: we can directly get the backgroundStyle alone here and avoid looking up the entire dictionary ourselves

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/DTCoreText/DTHTMLElement.swift
```diff
@@ -35,4 +35,70 @@ extension DTHTMLElement {
...
+                leadingDiscardableElement = createDiscardableElement()
+            }
+            
+            if text.hasSuffix("\(Character.nbsp)") {
+                shouldReplaceNodes = true
+                text.removeLast()
+                trailingDiscardableElement = createDiscardableElement()
+            }
+
+            if shouldReplaceNodes {
```
aringenbach, line 65: I'm guessing this could be replaced by 

```swift
let hasLeadingNbsp = text.hasPrefix(...)
let hasTrailingNbsp = text.hasSuffix(...)

if hasLeadingNbsp || hasTrailingNbsp {
    removeAllChildNodes() 
    
    // Prepare nodes and push them, we only need to create the new nodes at this point
    // Don't need to work with these optionals.
    [...] 
}
```

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/DTCoreText/DTHTMLElement.swift
```diff
@@ -35,4 +35,70 @@ extension DTHTMLElement {
...
+                if let leadingDiscardableElement = leadingDiscardableElement {
+                    addChildNode(leadingDiscardableElement)
+                    addChildNode(createLineBreak())
+                }
+
+                let newTextNode = DTTextHTMLElement()
+                newTextNode.inheritAttributes(from: self)
+                newTextNode.interpretAttributes()
+                newTextNode.setText(text)
+                addChildNode(newTextNode)
```
aringenbach, line 77: I think you might be able to re-use the same text node after detaching it instead of creating a new one (no need to re-inherit/interpret attributes then)

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
aringenbach: Yeah not sure what happens in that case, technically on backspace we do replace the HTML so we should be re-parsing and adding that ZWSP.

PR #521, opened 2023-01-25
Optimal fix for links in block nodes

Velin92:
The idea behind this optimisation is to split the selected range into multiple subranges for each block node and apply the insert_parent with a link on each of the subrange, in this way we can avoid links to contain block nodes, since each one of them, (and each element in between) will be treated as its own selection


jmartinesp (approved):
LGTM!

PR #524, opened 2023-01-26
NBSP Char in empty paragraphs in `<pre>`

Velin92:
Empty paragraphs in `<pre>` will now be added as an nbsp character in the `fmt_html` function.
At first I just wanted to have this character present for empty `<pre>` and on trailing newlines, but it seems we might actually need it on iOS for each newline (since DTCoreText seems to ignore newlines in cases like this: `<pre>\nTest</pre>
`)


jmartinesp (approved):
LGTM!

PR #525, opened 2023-01-26
Fix reversed actions in nested/indented lists

### LINKED ISSUE #510: Nested list items of different type highlight both buttons
## Steps to reproduce
1. Create an ordered list with 2 list items.
2. Indent the 2nd one.
3. Toggle its list type by toggling the unordered list button.

This can also be done by reversing the ordered -> unordered list types.

#### What did you expect?

I'd expect only the unordered list button to be highlighted, as it's the active one at this level of indentation.

#### What happened instead?

Both list buttons are highlighted. This is a bit confusing, as I'd expect the **highlighted** ordered list button to remove the external list type, 'reversing' the previous action, as most highlighted action buttons work, but instead it toggles the list type for the internal list items.

## Screenshot

This screenshot shows what it looks like at the moment, with the 'Now ordered' list item being selected.

image

aringenbach:
* Fixes issue with unmatching(type) indented lists marking both list types as reversed
* Renaming `active_button` to `reversed_action` to match rest of the naming

Fixes #510

### DIFF
--- a/crates/wysiwyg/src/tests/test_menu_state.rs
+++ b/crates/wysiwyg/src/tests/test_menu_state.rs
@@ -190,6 +190,28 @@ fn formatting_is_disabled_when_selection_covers_inline_code_node_and_others() {
 }
 
 #[test]
+fn selecting_indented_list_only_marks_the_deepest_list_type_as_reversed() {
+    let model = cm("<ol><li><p>Item 1</p><ul><li>Item 1|A</li></ul></li></ol>");
+    assert!(!model.action_is_reversed(ComposerAction::OrderedList));
+    assert!(model.action_is_reversed(ComposerAction::UnorderedList));
+}
+
+#[test]
+fn selecting_cross_unmatching_indented_list_exclude_both_list_types() {
+    let model =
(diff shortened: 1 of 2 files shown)


crates/wysiwyg/src/tests/test_menu_state.rs
```diff
@@ -189,6 +189,28 @@ fn formatting_is_disabled_when_selection_covers_inline_code_node_and_others() {
...
+    assert!(!model.action_is_reversed(ComposerAction::UnorderedList));
+}
+
+#[test]
+fn selecting_cross_matching_indented_list_include_list_type() {
+    let model =
+        cm("<ol><li><p>It{em 1</p><ol><li>Item 1}|A</li></ol></li></ol>");
+    assert!(model.action_is_reversed(ComposerAction::OrderedList));
+}
+
```
jmartinesp, line 213: Could it be worth it to add tests for 'deep' nested lists (`ul > ol > ul`)?

jmartinesp (approved):
LGTM, I just have a suggestion.

PR #526, opened 2023-01-26
Fix menu state when adding inline_code to `toggled_format_types`

### LINKED ISSUE #495: [Inline code] Disabled buttons states lagging when toggling inline code
Seen in the [example app](https://matrix-org.github.io/matrix-rich-text-editor/) on 18 Jan.

**Steps to reproduce:**

- click the inline code button
- note the state of the buttons, inline code shows as active, but disabled buttons do not appear disabled
- type a single letter
- now the disabled buttons appear disabled
- click on inline code again, note some buttons still show as disabled
- type a single letter, the buttons display the expected states

https://user-images.githubusercontent.com/56027671/213184583-4164e403-1b1d-4404-a343-804b936c935a.mov

**Expected behaviour:**

- on clicking the inline code button to start, the disabled formatting buttons should immediately be disabled
- on clicking the inline code button to finish, the disabled formatting buttons should immediately be enabled

**Notes:**

- As we will just lookup if we are inside a `<code>`, when we hit the button with no selection we only create the node when we start typing. We need to lookup both the current parent nodes & the pending formats to apply.

aringenbach:
Menu state is now properly and immediately updated when hitting the inline code button while on a cursor (zero-length selection). 

Fixes #495

### DIFF
--- a/crates/wysiwyg/src/composer_model/menu_state.rs
+++ b/crates/wysiwyg/src/composer_model/menu_state.rs
@@ -210,7 +210,15 @@ where
         if !self.can_unindent(&top_most_list_locations) {
             disabled_actions.insert(UnIndent);
         }
-        if contains_inline_code(locations) {
+        // XOR on inline code in selection & toggled format types.
+        // If selection is not a cursor, toggled format types is always
+        // empty, which makes `contains_inline_code` the only condition.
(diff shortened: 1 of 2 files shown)


Velin92 (approved):
LGTM

PR #528, opened 2023-01-26
NBSP only as leading and trailing in `<pre>`

Velin92:
This small improvement reduces the number of nbsp nodes so that they are present only in trailing and leading empty paragraphs, which ate the places where parser have a hard time tell if the newline is needed or not


aringenbach (approved):
LGTM

Issue #491, opened 2023-01-17
[Code blocks] [Quotes] Enter does not appear to move cursor to new line


alunturner:
Seen in the [example app](https://matrix-org.github.io/matrix-rich-text-editor/ ) on 17 Jan.

**Steps to reproduce:**

- Start a code block 
- enter some text
- press enter
- note cursor appears to still be on same line
- start typing
- new text appears on a new line
This can be seen in the video below:

https://user-images.githubusercontent.com/56027671/212918729-9770bb13-2e9c-4208-87be-9df8c9fe10c3.mov

**Expected behaviour:**

- pressing enter moves the cursor to a new line and this can be seen by the user

**Notes:**

- I'm not sure if this is purely a web issue, but I think it could be so have labelled it as such
- This bug is a bit annoying combined with our decision to make two consecutive presses of `Enter` exit the code block, as there's no visual indication of the first press
- Same behaviour can be seen in Quotes

alunturner:
This will be fixed by paragraph branch integration

langleyd:
This is fixed now

PR #530, opened 2023-01-27
Fix: insert paragraph at the start of a block node with enter

jmartinesp:
## What?

We had an issue where adding a new line at the start of the 1st paragraph inside a block node with multiple children wouldn't actually create a new paragraph at the start and would move the selected paragraph around instead.

We now check for this particular case and fix this behaviour. It also fixes a bug related to this issue in Android.

## Why?

Fixes [`PSU-1092`](https://element-io.atlassian.net/browse/PSU-1092) <del>and  #491</del> (my bad, it's not the same issue).


Velin92 (approved):
LGTM

PR #531, opened 2023-01-27
Fix: inserting paragraph at middle of block with enter

jmartinesp:
Refinement of #530 .

### DIFF
--- a/crates/wysiwyg/src/tests/test_paragraphs.rs
+++ b/crates/wysiwyg/src/tests/test_paragraphs.rs
@@ -432,7 +432,7 @@ fn pressing_enter_at_the_start_of_a_multiline_block_quote() {
 
 #[test]
 fn pressing_enter_in_the_middle_of_a_multiline_code_block() {
-    let mut model = cm("<pre>\n\n|line_1\nline_2</pre>");
+    let mut model = cm("<pre>&nbsp;\n|line_1\nline_2</pre>");
     model.enter();
     assert_eq!(tx(&model), "<pre>&nbsp;\n\n|line_1\nline_2</pre>")
 }
--- a/crates/wysiwyg/src/tests/test_paragraphs.rs
+++ b/crates/wysiwyg/src/tests/test_paragraphs.rs
@@ -429,3 +429,18 @@ fn pressing_enter_at_the_start_of_a_multiline_block_quote() {
         "<blockquote><p>&nbsp;</p><p>|line_1</p><p>line_2</p></blockquote>"
     )
 }
+
+#[test]
+fn pressing_enter_in_the_middle_of_a_multiline_code_block() {
+    let mut model = cm("<pre>\n\n|line_1\nline_2</pre>");
+    model.enter();
+    assert_eq!(tx(&model), "<pre>&nbsp;\n\n|line_1\nline_2</pre>")
+}
+
+#[test]
+fn pressing_enter_in_the_middle_of_a_multiline_block_quote() {
+    let mut model =
+        cm("<blockquote><p></p><p>|line_1</p><p>line_2</p></blockquote>");
+    model.enter();
+    assert_eq!(tx(&model), "<blockquote><p>&nbsp;</p><p>&nbsp;</p><p>|line_1</p><p>line_2</p></blockquote>")
+}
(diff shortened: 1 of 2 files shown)


crates/wysiwyg/src/tests/test_paragraphs.rs
```diff
@@ -429,3 +429,18 @@ fn pressing_enter_at_the_start_of_a_multiline_block_quote() {
         "<blockquote><p>&nbsp;</p><p>|line_1</p><p>line_2</p></blockquote>"
     )
 }
+
+#[test]
+fn pressing_enter_in_the_middle_of_a_multiline_code_block() {
+    let mut model = cm("<pre>\n\n|line_1\nline_2</pre>");
+    model.enter();
+    assert_eq!(tx(&model), "<pre>&nbsp;\n\n|line_1\nline_2</pre>")
+}
```
Velin92, line 438: not sure about this one? shouldn't it be `<pre>\n|line_1\nline_2</pre>` the initial condition?

jmartinesp: The first new line char will be removed when parsing, so it actually only has a single line break.

jmartinesp: I could add a `&nbsp;` at the start of the code block instead of the double line break though.

crates/wysiwyg/src/tests/test_paragraphs.rs
```diff
@@ -429,3 +429,18 @@ fn pressing_enter_at_the_start_of_a_multiline_block_quote() {
...
+
+#[test]
+fn pressing_enter_in_the_middle_of_a_multiline_block_quote() {
+    let mut model =
+        cm("<blockquote><p></p><p>|line_1</p><p>line_2</p></blockquote>");
+    model.enter();
+    assert_eq!(tx(&model), "<blockquote><p>&nbsp;</p><p>&nbsp;</p><p>|line_1</p><p>line_2</p></blockquote>")
+}
```
Velin92, line 446: I would actually also write the test case for `<blockquote><p>line_0</p><p>|line_1</p><p>line_2</p></blockquote>` and `<pre>line_0\n|line_1\nline_2</pre>`

crates/wysiwyg/src/tests/test_paragraphs.rs
Velin92: oh yeah right, actually maybe the other test is way more clear, the one I wrote below

jmartinesp: Done in [`86d17d4`](https://github.com/matrix-org/matrix-rich-text-editor/pull/531/commits/86d17d4de9cf92c810eef071d5bc03157e70225b).

crates/wysiwyg/src/tests/test_paragraphs.rs
jmartinesp: I added some tests in [`f0bc508`](https://github.com/matrix-org/matrix-rich-text-editor/pull/531/commits/f0bc50822e09e4d02f7140d7c90c181f61623eb0). Let me know if those are what you meant.

crates/wysiwyg/src/tests/test_paragraphs.rs
```diff
@@ -429,3 +429,35 @@ fn pressing_enter_at_the_start_of_a_multiline_block_quote() {
         "<blockquote><p>&nbsp;</p><p>|line_1</p><p>line_2</p></blockquote>"
     )
 }
+
+#[test]
+fn pressing_enter_in_the_middle_of_a_multiline_code_block_with_empty_starting_paragraphs(
+) {
+    let mut model = cm("<pre>|line_1\nline_2</pre>");
+    model.enter();
+    assert_eq!(tx(&model), "<pre>&nbsp;\n\n|line_1\nline_2</pre>")
```
Velin92, line 438: this one feels a bit weird... look like two lines have been added?
The thing is that, what should be a test case of pressing enter in a situation like this?
simulator_screenshot_C4F662C8-6EC4-47DC-8333-F495A1C3E62E

jmartinesp: Actually, that one was broken, too much copy&paste 😓 . It should be fixed in [`63969e9`](https://github.com/matrix-org/matrix-rich-text-editor/pull/531/commits/63969e9ebd3eb3ad38f7cee19f990f001b388fe6), adding a single paragraph before the starting one.

Velin92 (approved):
LGTM

PR #532, opened 2023-01-27
Paragraphs edge cases

alunturner:
The web integration had an issue with bold tags appearing part way into list items. The cursor would appear in the wrong place. This fixes that issue and adds a test to cover it.

Behaviour before:

https://user-images.githubusercontent.com/56027671/215111116-be174ed3-0001-4f0d-9f07-1077b031664e.mov

Behaviour after:

https://user-images.githubusercontent.com/56027671/215111136-8231dba9-f12c-496a-b252-b06067962218.mov


jmartinesp (approved):
I can't say much about the TS code, but the new behaviour LGTM.

PR #533, opened 2023-01-27
[Android] Add configuration options for code size

jonnyandrew:
## What?

- Allow the size of inline code and code blocks to be configured.
- Add default size for inline code and code blocks.

## Why?

Design feedback on the Element Android integration:
- https://github.com/vector-im/element-android/pull/8011

## Screenshots
| Before | After (with new defaults) |
|-|-|
|before|after|

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/spans/CodeSpanConstants.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/spans/CodeSpanConstants.kt
@@ -1,0 +1,5 @@
+package io.element.android.wysiwyg.spans
+
+internal object CodeSpanConstants {
+    const val DEFAULT_RELATIVE_SIZE_PROPORTION = 0.87f
+}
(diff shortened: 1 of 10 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/spans/CodeSpanConstants.kt
```diff
@@ -0,0 +1,5 @@
+package io.element.android.wysiwyg.spans
+
+internal object CodeSpanConstants {
+    const val DEFAULT_RELATIVE_SIZE_PROPORTION = 0.87f
```
jonnyandrew, line 4: 0.87 is the default size for code in Element Android (the value comes [from Markwon](https://github.com/noties/Markwon/blob/2ea148c30a07f91ffa37c0aa36af1cf2670441af/markwon-core/src/main/java/io/noties/markwon/core/MarkwonTheme.java#L106))

jmartinesp (approved):
LGTM!

PR #537, opened 2023-01-30
Add utility to build the array of ancestor handles from a handle

aringenbach:
* Add a util that allows to build the array of ancestors from an handle (including current)
* Update code in a couple locations to use this tool


jmartinesp (approved):
LGTM!

PR #538, opened 2023-01-30
iOS: Adding some padding to blockquotes and codeblocks, and adding a side border to blockquotes

Velin92:


platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/UITextView.swift
```diff
@@ -25,14 +25,33 @@ public extension UITextView {
...
+                guard let range = self.textRange(from: start, to: end) else { return }
+                let rects = selectionRects(for: range)
+                var textRect = CGRect.null
+                for rect in rects {
+                    textRect = textRect.union(rect.rect)
+                }
+                if !textRect.isNull {
+                    let rect = CGRect(x: 5, y: textRect.origin.y, width: 4, height: textRect.size.height)
+                    styleLayer = BackgroundStyleLayer(style: style, frame: rect)
+                }
```
aringenbach, line 52:
```suggestion
                let glyphRange = layoutManager.glyphRange(forCharacterRange: range, actualCharacterRange: nil)
                let textRect = layoutManager
                    .boundingRect(forGlyphRange: glyphRange, in: self.textContainer)
                let rect = CGRect(x: 5, y: textRect.origin.y, width: 4, height: textRect.size.height)
                styleLayer = BackgroundStyleLayer(style: style, frame: rect)
```

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParserStyle.swift
```diff
@@ -51,18 +51,22 @@ public struct HTMLParserStyle {
...
-    var quoteBackgroundStyle: BackgroundStyle {
-        BackgroundStyle(backgroundColor: quoteBackgroundColor,
-                        borderColor: quoteBorderColor,
-                        borderWidth: borderWidth,
-                        cornerRadius: cornerRadius)
+    var quoteBlockStyle: BlockStyle {
+        BlockStyle(backgroundColor: quoteBackgroundColor,
+                   borderColor: quoteBorderColor,
+                   borderWidth: borderWidth,
+                   cornerRadius: cornerRadius,
```
aringenbach, line 67: I guess both these should be 0.0 to draw properly

aringenbach (commented):
I see this slight issue: upon double-entering, the indent is not removed immediately after exiting the code block / quote

I guess we should have a ZWSP or something to old the actual indent of the new line (or don't have the padding to the last new line char of the code block/quote, if that works)

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/UITextView.swift
Velin92: I think that the indent issue is not related on this but  actually on the paragraph adding that we do in `applyBackgroundStyles(style: HTMLParserStyle)`

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/HTMLParserStyle.swift
Velin92: Yeah, I will set them to 0.0 as default value, but maybe we could keep them to give more control to the user of the library on how to render them

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/UITextView.swift
aringenbach: Oh yeah it's definitely not related to this piece of code. Basically we'll have either `\n(ZWSP)\n` or `(Some text)\n` at the end of code block/quotes. 
The very last `\n` shouldn't have the code block / quote paragraph style, but the standard paragraph style instead (as this represents the newline between the block and the next line)

Velin92: Tried to fix the issue by doing this when adding the paragraph style but... doesn't seem to work?
```
 if self.character(at: range.location + range.length - 1) == "\n" {
                    addAttribute(.paragraphStyle, value: NSParagraphStyle.default, range: NSRange(location: range.location + range.length - 1, length: 1))
                }
```

Velin92:
> I see this slight issue: upon double-entering, the indent is not removed immediately after exiting the code block / quote
> 
> I guess we should have a ZWSP or something to old the actual indent of the new line (or don't have the padding on the last new line char of the code block/quote, if that works)

Yeah this one is probably an issue of the reconciliate, maybe is better to fix it in a separate PR

aringenbach (approved):
LGTM

PR #539, opened 2023-01-30
Add indent and unindent

florianduros:


platforms/web/lib/useListeners/utils.ts
```diff
@@ -42,5 +42,6 @@ export function mapToAllActionStates(
                 key.substring(1)) as ActionTypes
         ] = value.toLowerCase() as ActionState;
     }
+    console.log('actionStatesMap', actionStatesMap);
```
alunturner, line 45: This is probably just leftover from integration work

platforms/web/lib/constants.ts
```diff
@@ -28,4 +28,6 @@ export const ACTION_TYPES = [
     'link',
     'codeBlock',
     'quote',
+    'indent',
+    'unIndent',
```
alunturner, line 32: Does this have to be `unIndent` for any reason? To match something from the rust model? I find it jarring, but if it has to be that way, it has to be that way

florianduros: Yes, it's needed to match the state returned by the rust model

platforms/web/lib/useListeners/utils.ts
florianduros: Good catch !

alunturner (approved):
Not quite sure why SonarCloud thinks all the test cases are repetition in that file

florianduros:
Also, it seems only work in the case of list. You can't indent outside of list

PR #541, opened 2023-01-30
[Android] Fix crashes in `EditorStyledTextView`

jonnyandrew:
- Fix crash caused by unsafe access of lateinit vars
- Fix crash caused by missing default styles
- Add a simple test for the `EditorStyledTextView`


jmartinesp (approved):
LGTM! Sorry, I should have checked this more thoroughly.

PR #543, opened 2023-01-31
Fix line indent after quotes & code blocks

aringenbach:
Fixes line indent issues by replacing all `discardable texts` by ZWSP within the text view. ZWSP located right after an indented block will hold the expected attributes to fix the indent on the very next new line.


Velin92 (approved):
LGTM

PR #545, opened 2023-01-31
[Android] Fix rendering of code with trailing new line

jonnyandrew:
## What?

Ignore a single trailing newline when rendering code.

## Why?

[`PSU-1099`](https://element-io.atlassian.net/browse/PSU-1099)

Code blocks created using the legacy composers contain a trailing newline which is not intended to be displayed as part of the code block.


aringenbach (approved):
LGTM

PR #547, opened 2023-01-31
Fix parent `li` paragraph hierarchy after backspacing an indented `li`

### LINKED ISSUE #544: Backspacing an indented list item leads to inconsistent `enter` behaviour afterwards
```rust
let mut model = cm("item|");
model.unordered_list();
model.enter();
model.indent();
model.replace_text("it");
model.enter();
model.indent();
model.backspace();
assert_eq!(tx(&model), "<ul><li><p>item</p><ul><li><p>it|</p></li></ul></li></ul>");
// This ^ should probably be "<ul><li><p>item</p><ul><li>it|</li></ul></li></ul>"
model.enter();
assert_eq!(tx(&model), "<ul><li><p>item</p><ul><li><p>it|</p><p>&nbsp;</p></li></ul></li></ul>");
// So ^ this would be: "<ul><li><p>item</p><ul><li>it</li><li>|</li></ul></li></ul>"
```

Upon hitting `enter` right after backspacing an empty indented list item, a new `p` is created inside the list item, instead of creating a new list item. This is due to the text deletion not handling this special case where removing the nested list should rework the parent list item which shouldn't contain a `p` anymore.

aringenbach:
Fixes #544

### DIFF
--- a/crates/wysiwyg/src/dom/dom_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_methods.rs
@@ -138,8 +138,39 @@ where
                 }
             }
         }
+
+        if let Some(leaf_handle) = self.first_leaf_handle_at_location(start) {
+            self.remove_list_item_child_paragraph_if_needed(&leaf_handle);
+        }
+
         #[cfg(any(test, feature = "assert-invariants"))]
         self.assert_invariants();
+    }
+
+    /// Removes paragraph from the closest list item ancestor, if
+    /// it exists and if it is an only child.
+    fn remove_list_item_child_paragraph_if_needed(
+        &mut self,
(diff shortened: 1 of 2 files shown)


jonnyandrew (approved):

jonnyandrew (commented):
Is it worth making this one of the invariants we check for? I.e. list items cannot contain a single `<p>` tag

aringenbach:
> Is it worth making this one of the invariants we check for? I.e. list items cannot contain a single `<p>` tag

Could be interesting indeed, the current issue right now is that we have some functions that assert invariants which are used internally on states were this is not true (e.g. `remove_and_keep_children`)

Issue #540, opened 2023-01-30
Select all (ctrl/cmd a) not working correctly on web


alunturner:
Seen in the example app 30 Jan 23. 

**Steps to reproduce:**

- [Go to the rich text editor demo](https://matrix-org.github.io/matrix-rich-text-editor/)
- enter a few lines of text
- use cmd+a (ctrl+a on windows) to select all of the text
- observe the output below the textbox - the selection there does not match what you see in the composer
- press delete or backspace
- observe that the text deleted matches what was shown in the model, not what was highlighted in the composer

https://user-images.githubusercontent.com/56027671/215539603-f5e06fd4-8477-47fd-9dad-ccd78d47fcbb.mov

**Notes**

Looks to be something wrong in the handleSelectionChange function, likely due to the switch to paragraphs

PR #548, opened 2023-02-01
Ctrl/cmd a behaviour web

alunturner:
This will close [issue 540](https://github.com/matrix-org/matrix-rich-text-editor/issues/540).

The fix for the cmd a behaviour is small, we just needed to count the extra offsets that are introduced by certain tags after the paragraph refactor.

That work, however, made me realise that a lot of the dom tests for `getCurrentSelection` needed rewriting as they no longer represented what a user would be writing in the composer. This is why the diff is so large.

Behaviour:

https://user-images.githubusercontent.com/56027671/216019885-bb68cc26-156d-4aa8-95a5-2980e0492352.mov

### DIFF
--- a/platforms/web/lib/dom.test.ts
+++ b/platforms/web/lib/dom.test.ts
@@ -612,7 +612,7 @@ describe('getCurrentSelection', () => {
         expect(sel.focusNode).toBe(secondParagrahStrongTextNode);
         expect(sel.focusOffset).toBe(secondNodeOffset);
 
-        // We should see ourselves as on code unit 8, because a paragraph tag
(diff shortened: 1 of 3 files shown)


alunturner:
Just putting back into draft whilst I appease SonarCloud

alunturner:
If I check the coverage locally, it doesn't match the issues that SonarCloud is showing, I'm not sure what is going on there

florianduros:
> If I check the coverage locally, it doesn't match the issues that SonarCloud is showing, I'm not sure what is going on there

The CI doesn't run the coverage however the report is sent to sonar.

florianduros (approved):

PR #549, opened 2023-02-01
[API breaking change] Rename UnIndent as Unindent

aringenbach:
* Rename all occurrences of UnIndent/un_indent to Unindent/unindent
* This affects some public/external APIs and will probably require a minor update on clients upon next release


jonnyandrew (approved):
Should we also add a note to the changelog (or the commit message on `main`) in advance so that we don't forget this is a breaking change?

aringenbach:
> Should we also add a note to the changelog (or the commit message on `main`) in advance so that we don't forget this is a breaking change?

Sounds good, I added a noticeable reminder on the commit message so we don't forget about it upon next release.

PR #550, opened 2023-02-01
Fix `enter` behaviour on empty list item with formatting

### LINKED ISSUE #546: Entering twice in a list with formatting doesn't move out of list...
...and indefinitely repeats an extra list item creation followed by it's transformation into a `p`.

```rust
let mut model = cm("a|");
model.unordered_list();
model.enter();
model.replace_text("b");
model.bold();
model.select(Location::from(3), Location::from(3));
model.enter();
assert_eq!(tx(&model), "<ul><li>a</li><li><strong>b</strong></li><li><strong>|</strong></li></ul>");
model.enter();
assert_eq!(tx(&model), "<ul><li>a</li><li><strong>b</strong></li><li>|</li><li><strong></strong></li></ul>");
model.enter();
assert_eq!(tx(&model), "<ul><li>a</li><li><strong>b</strong></li><li><strong>|</strong></li></ul><p>&nbsp;</p>");
```

aringenbach:
Fixes #546 

* A list item is now considered empty as long as its text_len is 0 (since it can contain some empty formatting nodes)
* Contained formatting nodes are transferred to the new paragraph or list item

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -376,7 +376,7 @@ where
 
     pub fn is_empty_list_item(&self) -> bool {
         match self.kind {
-            ContainerNodeKind::ListItem => self.is_empty(),
+            ContainerNodeKind::ListItem => self.text_len() == 0,
             _ => false,
         }
     }
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -374,13 +374,6 @@ where
         }
     }
 
-    pub fn is_empty_list_item(&self) -> bool {
-        match self.kind {
-            ContainerNodeKind::ListItem => self.text_len() == 0,
-            _ => false,
-        }
-    }
-
     pub(crate) fn get_list_type(&self) -> Option<&ListType> {
         match &self.kind {
             ContainerNodeKind::List(t) => Some(t),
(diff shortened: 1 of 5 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -376,7 +376,7 @@ where
 
     pub fn is_empty_list_item(&self) -> bool {
         match self.kind {
-            ContainerNodeKind::ListItem => self.is_empty(),
+            ContainerNodeKind::ListItem => self.text_len() == 0,
```
alunturner, line 379: I wonder if this would be better as a change to the implementation of `is_empty()`? Perhaps that function becomes a check of if it has no children or if it has children but a text length of zero, return true.

It could be that other places check the `ListItem` using `.is_empty()` and I presume we'd want them to be using the same check

aringenbach: That's a fair point, I'll check that, and try to do something a bit smart, because `text_len()` will be slower (hence why I didn't really wanted to apply these everywhere), we probably want to return as soon as we find a non-empty text node.

aringenbach: I actually removed this method since it was unnecessarily checking ContainerNode type on context where it is already known it is a list item.
I added a method to check if a Container or any node actually contains text, this allows to make a difference between `really empty` and `contains no text` which are not the same.

alunturner (approved):
Looks good to me

PR #552, opened 2023-02-01
Fix menu state in empty formatted paragraphs/list items

### LINKED ISSUE #551: Empty paragraph/list item with formatting don't return expected menu state
Test cases that don't work:

```rust
#[test]
fn empty_paragraph_with_formatting_computes_expected_menu_state() {
    let model = cm("<p><em>abc</em></p><p><em>|</em></p>");
    assert!(model.action_is_reversed(ComposerAction::Italic));
}

#[test]
fn empty_list_item_with_formatting_computes_expected_menu_state() {
    let model = cm("<ol><li><em>abc</em></li><li><em>|</em></li></ol>");
    assert!(model.action_is_reversed(ComposerAction::Italic));
}
```

aringenbach:
Fixes #551

### DIFF
--- a/crates/wysiwyg/src/dom/range.rs
+++ b/crates/wysiwyg/src/dom/range.rs
@@ -399,6 +399,22 @@ impl Range {
             .max()
     }
 
+    pub(crate) fn deepest_container_node(
+        &self,
+        ancestor_of: Option<&DomHandle>,
+    ) -> Option<&DomLocation> {
+        self.locations
+            .iter()
+            .filter(|l| {
+                let mut found = true;
+                if let Some(ancestor_of) = ancestor_of {
+                    found = l.node_handle.is_ancestor_of(ancestor_of);
+                }
+                found && !l.is_leaf()
+            })
+            .max()
+    }
+
     pub(crate) fn deepest_node_of_kind(
         &self,
         kind: DomNodeKind,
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/dom/range.rs
```diff
@@ -399,6 +399,22 @@ impl Range {
...
+            .filter(|l| {
+                let mut found = true;
+                if let Some(ancestor_of) = ancestor_of {
+                    found = l.node_handle.is_ancestor_of(ancestor_of);
+                }
+                found && !l.is_leaf()
+            })
+            .max()
+    }
+
```
alunturner, line 417: Could this be a call to `deepest_node_of_kind` where it adds the container kind to the function call for you?

alunturner (commented):
Just the one comment, but looks good to me other than that

crates/wysiwyg/src/dom/range.rs
aringenbach: If it was only for formatting we'd still have to rework the `deepest_node_of_kind` func because it just compares `kind` and things like `kind() == DomNodeKind::Formatting(_)` is not allowed.
Moreover we probably want to include any type of container that might be present in between since it could be links, lists, etc and these should update the menu state as well.

alunturner (approved):

Issue #456, opened 2023-01-06
Web does not handle `deleteSoftLineBackward` event


alunturner:
The `deleteSoftLineBackward` event can be fired on mac by doing cmd + backspace (expect it would be ctrl + backspace on windows).

Expected
Pressing cmd/ctrl + backspace should remove a line

Actual
Pressing cmd/ctrl + backspace appears to remove the line, but then type anything and the line reappears

This can be seen in the sample app, screen grab below:

https://user-images.githubusercontent.com/56027671/210985122-377ccad6-0521-4cc6-bfd1-1e20aebf0ce2.mov

alunturner:
Fixed in v1.0.0

PR #553, opened 2023-02-02
Soft delete line backwards implementation

alunturner:
This work is done to address [issue 456](https://github.com/matrix-org/matrix-rich-text-editor/issues/456).

Implementation has to be done on the web side, not the rust side, as we need to know the width of what is being rendered.

The testing is limited because jsdom does not implement `Selection.modify`. I'm not surprised it doesn't, because the functionality is surprisingly clever and would involve a whole heap of modelling that (I'm guessing) jsdom probably doesn't do at the moment.

### DIFF
--- a/platforms/web/lib/composer.ts
+++ b/platforms/web/lib/composer.ts
@@ -71,6 +71,13 @@ export function processInput(
             return action(composerModel.backspace(), 'backspace');
         case 'deleteWordBackward':
             return action(composerModel.backspace_word(), 'backspace_word');
+        case 'deleteSoftLineBackward': {
+            document
+                .getSelection()
+                ?.modify('extend', 'backward', 'lineboundary');
+            document.dispatchEvent(new CustomEvent('selectionchange'));
+            return action(composerModel.delete(), 'backspace_line');
+        }
         case 'deleteContentForward':
             return action(composerModel.delete(), 'delete');
         case 'deleteWordForward':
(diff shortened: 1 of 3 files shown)


platforms/web/lib/composer.ts
```diff
@@ -71,6 +71,13 @@ export function processInput(
             return action(composerModel.backspace(), 'backspace');
         case 'deleteWordBackward':
             return action(composerModel.backspace_word(), 'backspace_word');
+        case 'deleteSoftLineBackward': {
+            document
+                .getSelection()
+                ?.modify('extend', 'backward', 'lineboundary');
+            document.dispatchEvent(new CustomEvent('selectionchange'));
```
florianduros, line 78: If we can't call modify (ie no selection), I think we shouldn't dispatch the `selectionchange` event

jonnyandrew (approved):
First I heard about this command! Looks good to me

florianduros (changes requested):
Looks good, thanks!

A small improvement can be made

platforms/web/lib/composer.ts
alunturner: That's a fair point, I'll do that to appease TS, but the case that returns a null selection is pretty niche

alunturner: Changes applied

florianduros (approved):

PR #555, opened 2023-02-02
Add transaction functionality to Dom

jonnyandrew:
## What?
Adds the ability to perform a set of operations in a transactional manner on the Dom class. The transaction is indicated by the new member variable `is_transaction_in_progess`, which is set to `true` when the operations start and `false` when they complete. The class now includes methods for starting and ending a transaction, which enforce class invariants and ensure that only one transaction can be active at a time.

## Why?

Sometimes we need to carry out a number of operations without maintaining consistency of the DOM between them. However, some DOM methods assert invariants each time they are run. Transactions allow us to run these methods until we reach a good state, at which point we can safely assert the invariants.

### DIFF
--- a/crates/wysiwyg/src/dom/dom_invariants.rs
+++ b/crates/wysiwyg/src/dom/dom_invariants.rs
@@ -53,6 +53,11 @@ where
 
     #[cfg(any(test, feature = "assert-invariants"))]
     pub(crate) fn assert_invariants(&self) {
+        if self.is_transaction_in_progress() {
+            // Invariants should not be asserted during a transaction
+            // as the DOM is known to be in an inconsistent state
+            return;
+        }
         self.assert_no_empty_text_nodes();
         self.assert_no_adjacent_text_nodes();
         self.assert_exactly_one_generic_container();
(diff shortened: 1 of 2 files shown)


aringenbach (commented):
Interesting I think there are two ways we can tackle this issue, the other way could be to enforce invariants in `pub` methods of the dom only. (and then a complex operation would have a dedicated `pub` method that calls internal versions of the methods that don't check invariants, only `pub` methods need to do that)

If we are going for the transaction approach, we need at least to have a way to enforce we don't create the composer update in the middle of a transaction, not sure how annoying this would be (hopefully not too much - but we need to make sure we are always checking that).

Also all the transaction code should only be available when `#[cfg(any(test, feature = "assert-invariants"))]`, not sure if we can do that to the extra property though.

jonnyandrew:
> the other way could be to enforce invariants in pub methods of the dom only. (and then a complex operation would have a dedicated pub method that calls internal versions of the methods that don't check invariants, only pub methods need to do that)

I thought about doing this but it felt a bit awkward for my use case:

```
dom.remove_and_keep_children(handle_a); // creates an inconsistent state
dom.remove_and_keep_children(handle_b); // creates a consistent state
```

In order to make a public method that maintains the invariant, I found myself making the public interface more complex (and a bit confusing):

```
dom.remove_and_keep_children(vec![handle_a, handle_b]);
```

So the reason I thought of creating the transaction was to allow these simpler methods to be composed. What do you think?

> we need at least to have a way to enforce we don't create the composer update in the middle of a transaction
> Also all the transaction code should only be available when #[cfg(any(test, feature = "assert-invariants"))]

I agree with these points - will take a look.

aringenbach:
@jonnyandrew Yeah I guess on very complex operations there is a strong argument upon trying not to create some functions that will end up with a million parameters.

As long as we can have the `no transaction in progress` check and hide as much as this code as possible behind cfg conditions, I'm fine with it.

aringenbach (approved):
LGTM

jonnyandrew:
@aringenbach I didn't end up needing transactions after changing my approach to code block HTML rendering.

Do you think it's still worth merging even though it's currently unused? I think it could come in handy for other invariant checks that would not otherwise be possible (for example https://github.com/matrix-org/matrix-rich-text-editor/pull/547#issuecomment-1410756188).

PR #559, opened 2023-02-02
Set iOS BlockStyle parameters as public

aringenbach:
Set required API for iOS quote/code blocks style as public


langleyd (approved):
LGTM

PR #561, opened 2023-02-06
Fix code block HTML rendering

jonnyandrew:
## What?

Currently, code blocks are rendered to HTML as a `<pre>` element. After this change, code blocks are rendered using `<code>` wrapped in `<pre>`.

Note that code blocks are still represented internally as a single block node and only converted to this new format during HTML rendering so that the DOM AST is not coupled to the HTML representation.

### Before

```
<pre>
...code content...
</pre>
```

### After
```
<pre>
<code>
...code content...
</code>
</pre>
```

## Why?

The HTML `<pre>` tag represents 'preformatted text' but doesn't imply the text is code. Some clients may choose to format `<pre>` as code but we cannot assume they will. In order to describe the contents of a code block explicitly, we can use the `<code>` tag.

From [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code):

> To represent multiple lines of code, wrap the [`<code>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code) element within a [`<pre>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre) element.

[`PSU-1099`](https://element-io.atlassian.net/browse/PSU-1099)

### DIFF
--- a/crates/wysiwyg/src/composer_model/base.rs
+++ b/crates/wysiwyg/src/composer_model/base.rs
@@ -233,7 +233,7 @@ mod test {
                 </blockquote>\
                 <p>&nbsp;</p>\
                 <p>Some text</p>\
-                <pre>A\n\tcode\nblock</pre>\
+                <pre><code>A\n\tcode\nblock</code></pre>\
                 <p>Some <code>inline</code> code</p>",
         ));
         assert_eq!(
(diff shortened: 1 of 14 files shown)


aringenbach (approved):
LGTM

PR #564, opened 2023-02-06
Handle insertReplacementText input event

florianduros:


aringenbach (approved):
LGTM

IIRC, `set_content_from_html` by default sets the selection to the last char in the Rust model. Are we sure we always recover properly from this ?

florianduros:
I hadn't any trouble in my local test. When the `insertReplacementText` input event is fired, the editor already had the new content, we only need to update the new content from the rust model. 

Sadly, I can't use the `replace_text` method, the browser handles the selection and the new string differently in the spellchecking case.

PR #565, opened 2023-02-06
Add an invariant assert for `li` never contains a single `p`

aringenbach:
Add an invariant assert for list items to never contain a single paragraph. 

Note: added temporary FIXMEs we can try to make work since it might be better that transactions are performed on dedicated dom methods if possible.

### DIFF
--- a/crates/wysiwyg/src/dom/dom_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_methods.rs
@@ -63,7 +63,7 @@ where
 
     pub fn replace_text_in(&mut self, new_text: S, start: usize, end: usize) {
         #[cfg(any(test, feature = "assert-invariants"))]
-        self.assert_invariants();
+        self.start_transaction();
 
         let extra_len = new_text.len() + 1;
         let range = self.find_range(start, end);
@@ -144,7 +144,7 @@ where
         }
 
         #[cfg(any(test, feature = "assert-invariants"))]
-        self.assert_invariants();
+        self.end_transaction();
     }
 
     /// Removes paragraph from the closest list item ancestor, if
--- a/crates/wysiwyg/src/composer_model/lists.rs
+++ b/crates/wysiwyg/src/composer_model/lists.rs
@@ -430,6 +430,10 @@ where
             list_items_after_removed_ones.insert(0, child);
         }
 
+        // FIXME: extract transaction to a dom method if possible
+        #[cfg(any(test, feature = "assert-invariants"))]
+        self.state.dom.start_transaction();
+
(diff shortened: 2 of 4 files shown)


crates/wysiwyg/src/dom/dom_methods.rs
```diff
@@ -63,7 +63,7 @@ where
 
     pub fn replace_text_in(&mut self, new_text: S, start: usize, end: usize) {
         #[cfg(any(test, feature = "assert-invariants"))]
-        self.assert_invariants();
+        self.start_transaction();
```
jonnyandrew, line 66: By starting the transaction here, we're saying that `replace_text_in()` cannot itself be composed within a transaction. Was that the intention or should the transaction be moved into the composer layer?

jonnyandrew (approved):
Just one question I had, otherwise looks good!

crates/wysiwyg/src/dom/dom_methods.rs
aringenbach: Oh yeah I haven't thought about that directly. I was mostly thinking about "let's do most of the transactions code inside the Dom layer", but composition might be an issue for this kind of functions. 
Maybe we need to think a bit about what we want to achieve. I think I'd like a solution were a dom method can perform a transaction with multiple other methods, but be available as well for composition 😅 

Maybe we shouldn't fail if a transaction is already in progress, and make it possible to compose transaction with other transaction, but I guess we would have to modify the mechanism a bit (like, attach an id or something to a transaction, and make it so the transaction can only be ended by its initiator) 🤔

jonnyandrew: The thinking behind disallowing nested transactions was that it would simplify things and encourage us to design Dom methods that should usually preserve the invariants. But equally I can see how it might be an unnecessary constraint.

> we would have to modify the mechanism a bit (like, attach an id or something to a transaction, and make it so the transaction can only be ended by its initiator

Yes I think so - I wonder if a macro would help us here (maybe automatically assign the caller name as an ID and ensure the transaction is ended at the end of a method)?

aringenbach: BTW I've been checking on this and actually there is no direct way to get the caller function name / id: https://github.com/rust-lang/rfcs/pull/2818

This is pretty annoying because I guess we'd have to explicitly pass some information to the caller so it can end the transaction, which makes the API less nice to use.

aringenbach:
Closed this for now, as the solution is not a 100% satisfying, might come back to it later.

PR #566, opened 2023-02-06
[Android] Recover from Rust model crashes

jonnyandrew:
## What?

Recover from Rust panics on Android.

When an error is seen,
- replace the composer model with a new instance
- attempt to revert to the last known good state with all formatting removed

This does not affect the behaviour of debug builds, which should continue to crash.

## Why?

Currently, when the Rust model panics, the editor simply hangs and creates a confusing UX.

[`PSU-1112`](https://element-io.atlassian.net/browse/PSU-1112)

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -252,8 +252,14 @@ impl ComposerModel {
         self.inner.lock().unwrap().get_link_action().into()
     }
 
+    #[cfg(not(debug_assertions))]
+    pub fn debug_panic(self: &Arc<Self>) {
+        // No-op
+    }
+
     /// Force a panic for test purposes
-    pub fn force_panic(self: &Arc<Self>) {
+    #[cfg(debug_assertions)]
+    pub fn debug_panic(self: &Arc<Self>) {
         panic!("This should only happen in tests.");
     }
 }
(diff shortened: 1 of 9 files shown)


bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -251,4 +251,9 @@ impl ComposerModel {
     pub fn get_link_action(self: &Arc<Self>) -> LinkAction {
         self.inner.lock().unwrap().get_link_action().into()
     }
+
+    /// Force a panic for test purposes
+    pub fn force_panic(self: &Arc<Self>) {
+        panic!("This should only happen in tests.");
+    }
```
aringenbach, line 258: I think I'd rather see that not committed to the repo. We can force crash easily with some very invalid `set_content_from_html` usage. e.g. I use "<//strong>" but I guess a lot of other things can work.

aringenbach (approved):
One small suggestion, otherwise LGTM

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
jonnyandrew: The reason I added this was so that we could make it explicit and guarantee the crash would happen in tests (whereas something like `set_content_from_html("<//strong>")` is a bug that we could expect to be fixed later).

Would it help if I put it behind a conditional compilation so that there's no risk of this getting into production? E.g.


    /// Force a panic for test purposes (no-op in release builds)
    #[cfg(not(debug_assertions))]
    pub fn debug_panic(self: &Arc<Self>) {
        // No-op
    }

    /// Force a panic for test purposes
    #[cfg(debug_assertions)]
    pub fn debug_panic(self: &Arc<Self>) {
        panic!("This should only happen in tests.");
    }


Ideally we could also exclude it from the binding interface too but I couldn't work a way to do this.

aringenbach: Yeah I don't think you could translate Rust cfg easily to have this magically disappear on production build. 
I guess this is fine then.

PR #568, opened 2023-02-07
Fix crashes from HTML parsing

aringenbach:
WIP on avoiding crashes during HTML parsing
Creates a new type of ComposerUpdate that is a `PanicRecovery` which returns the previous state for the client to restore content.


aringenbach:
Use `Result` for function that should explicitly throw: https://github.com/matrix-org/matrix-rich-text-editor/pull/581

PR #569, opened 2023-02-07
Hide indent/unindent buttons outside of lists

aringenbach:
* Adds a new `ActionState` to specify some action that should be hidden rather than just disabled.
* Indent/Unindent are now hidden outside of lists
* When lists are selected or partially selected, the buttons are visible but still rely on `can_indent` / `can_unindent` to be enabled

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_action_state.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_action_state.rs
@@ -3,6 +3,7 @@ pub enum ActionState {
     Enabled,
     Reversed,
     Disabled,
+    Hidden,
 }
 
 impl From<&wysiwyg::ActionState> for ActionState {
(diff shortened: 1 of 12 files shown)


jonnyandrew:
@aringenbach should the visibility of the button be a concern of this library? I'd think it would be up to the host app to decide whether a disabled button is visible (as is the rest of the action button UI currently).

aringenbach:
@jonnyandrew With this change we do have a difference between `disabled` and `hidden`. e.g. a button can be disabled but not hidden, like bold when we're inside a code block. Or even indent/unindent when we can't process the action but are inside a list anyway. 

I guess we could make all the code for this in the platform specific implementation. It will repeat the same code in all platforms though. (`if indent is disabled && (ol is reversed || ul is reversed)`)

jonnyandrew:
> a button can be disabled but not hidden, like bold when we're inside a code block like bold when we're inside a code block

That's a good example because the application might choose to hide the bold button while inside a code block but the library doesn't have an opinion on whether it should.

I think the indentation buttons are similar in that they are not applicable in a certain context. The only difference here is that they are disabled when _not_ inside other formatting.

IMO it's right to duplicate this logic across the clients here, like we have done with the rest of the button UI

jonnyandrew:
If we wanted to centralise the logic in this library perhaps we could add an 'visibility hint' for the action that is separate from the enabled/disabled state? This way the app can see what we'd recommend but choose whether it follows the design choice or not.

aringenbach:
@florianduros Yeah I agree, it's fine, as long as the computation of these status is not too hard to handle from a client perspective. I'm closing this.

PR #571, opened 2023-02-08
[iOS] Remove force tries in binding code and adapt the overall code to throwing functions

aringenbach:
* Allow to catch Rust panics within the iOS code
* Add tests for it
* Replace content of the composer with the latest plain text in case of failure

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,10 @@ ios: setup
 	  -headers ${IOS_GENERATION_DIR}/headers \
 	  -library ../../target/ios-simulator/libuniffi_wysiwyg_composer.a \
 	  -headers ${IOS_GENERATION_DIR}/headers \
-	  -output ${IOS_PACKAGE_DIR}/WysiwygComposerFFI.xcframework
+	  -output ${IOS_PACKAGE_DIR}/WysiwygComposerFFI.xcframework && \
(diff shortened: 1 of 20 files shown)


Makefile
```diff
@@ -53,7 +53,10 @@ ios: setup
 	  -headers ${IOS_GENERATION_DIR}/headers \
 	  -library ../../target/ios-simulator/libuniffi_wysiwyg_composer.a \
 	  -headers ${IOS_GENERATION_DIR}/headers \
-	  -output ${IOS_PACKAGE_DIR}/WysiwygComposerFFI.xcframework
+	  -output ${IOS_PACKAGE_DIR}/WysiwygComposerFFI.xcframework && \
+	sed -i "" -e '1h;2,$$H;$$!d;g' -e 's/) -> ComposerUpdate {\n        return try! FfiConverterTypeComposerUpdate.lift(\n            try!/) throws -> ComposerUpdate {\n        return try FfiConverterTypeComposerUpdate.lift(\n            try/g' ${IOS_PACKAGE_DIR}/Sources/WysiwygComposer/WysiwygComposer.swift && \
+	sed -i "" -e '1h;2,$$H;$$!d;g' -e 's/) -> ComposerUpdate/) throws -> ComposerUpdate/g' ${IOS_PACKAGE_DIR}/Sources/WysiwygComposer/WysiwygComposer.swift
+
```
jonnyandrew, line 59: I guess this could break in future if the behaviour of `uniffi-bindgen` changes? Would it be worth adding a comment in `ComposerModelWrapper.swift` to help future maintainters track down any potential issues?

And maybe a comment here to explain what it is and why it's here?

jonnyandrew (approved):
It's a shame that there was no other way than to hack the Swift bindings in order to catch the Rust panics :sweat_smile: But otherwise it looks good. I think a comment to explain could help us out in future.

Makefile
aringenbach: Done

jonnyandrew (approved):
Thanks!

PR #573, opened 2023-02-08
Add utility to get content as plain text

jonnyandrew:
## What?

Add a (basic) utility to get content as plain text.

## Why?

If the internal Rust model panics due to a bug, we'd like to be able to reset and continue editing without losing any data. However, recovering to the last known good state is risky because this state is 'nearby' the bad state that caused the panic.

So instead, this method allows the editor to recover the contents of the editor from the Rust model in plain text, stripped of all formatting, and allow the user to continue editing from a safer state.

[`PSU-1117`](https://element-io.atlassian.net/browse/PSU-1117)

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -20,9 +20,10 @@ use crate::dom::dom_handle::DomHandle;
 use crate::dom::nodes::dom_node::{DomNode, DomNodeKind};
 use crate::dom::to_html::{ToHtml, ToHtmlState};
 use crate::dom::to_markdown::{MarkdownError, MarkdownOptions, ToMarkdown};
+use crate::dom::to_plain_text::ToPlainText;
 use crate::dom::to_raw_text::ToRawText;
 use crate::dom::to_tree::ToTree;
-use crate::dom::unicode_string::{UnicodeStrExt, UnicodeStringExt};
+use crate::dom::unicode_string::{UnicodeStr, UnicodeStrExt, UnicodeStringExt};
 use crate::dom::{self, UnicodeString};
 use crate::{InlineFormatType, ListType};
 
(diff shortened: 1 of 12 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -765,6 +766,61 @@ where
     }
 }
 
+impl<S> ToPlainText<S> for ContainerNode<S>
```
jonnyandrew, line 769: This implementation is not meant to be perfect, but good enough to preserve the flow of content by ensuring there are line breaks between block nodes, list items etc.

crates/wysiwyg/src/tests/test_to_plain_text.rs
```diff
@@ -0,0 +1,212 @@
...
+#[test]
+fn text_with_italic() {
+    assert_to_plain("<em>abc</em>", "abc");
+    assert_to_plain("abc <em>def</em> ghi", "abc def ghi");
+    assert_to_plain(
+        "abc <em>line1<br />line2<br /><br />line3</em> def",
+        r#"abc line1
+line2
+
+line3 def"#,
```
aringenbach, line 53: We can use indoc macro to make these nicer to read

e.g.

```rust
        indoc! {
            r#"
            abc line1
            line2
    
            line3 def"#
        },
```

crates/wysiwyg/src/tests/test_to_plain_text.rs
```diff
@@ -0,0 +1,212 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::{dom::to_plain_text::ToPlainText, ComposerModel};
+use widestring::Utf16String;
+
+#[test]
+fn text() {
```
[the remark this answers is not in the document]
aringenbach: Maybe add a test for indented lists as well ? Since it is one of our latest enabled feature, and I guess if it crashes we want to be sure we recover nicely.

aringenbach (approved):
Looks good, just some optional suggestions

crates/wysiwyg/src/tests/test_to_plain_text.rs
jonnyandrew: I agree, that way is much more readable! [`d62d26e`](https://github.com/matrix-org/matrix-rich-text-editor/pull/573/commits/d62d26e2be63692fe432da1da517b58cad8f7e9c)

crates/wysiwyg/src/tests/test_to_plain_text.rs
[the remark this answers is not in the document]
jonnyandrew: [`cb7d3db`](https://github.com/matrix-org/matrix-rich-text-editor/pull/573/commits/cb7d3db0d962740a81f781a514a19045930ae160)

aringenbach (approved):
LGTM

PR #575, opened 2023-02-08
Handle rust model crash

florianduros:


aringenbach (approved):
Is it possible to fix the code smells reported by sonarcloud easily ? Otherwise LGTM

PR #576, opened 2023-02-09
[Android] Update NDK to r25c

jonnyandrew:
## What?

Update the Android NDK to LTS [r25c](https://developer.android.com/ndk/downloads#lts-downloads).

## Why?

- This is the latest long term support version
- It's now the default installed version on the MacOS 12 Github Actions runner so no need to install it manually

### DIFF
--- a/platforms/android/library/build.gradle
+++ b/platforms/android/library/build.gradle
@@ -61,7 +61,7 @@ android {
         jacocoVersion = "0.8.8"
     }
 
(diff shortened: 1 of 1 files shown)


aringenbach (approved):
Looks like it worked, cool ! :) 

I'm a bit scared that we get inconsistent behaviour depending on which runners come up, we'll see (I know I had some issues before with versions of Xcode that just weren't there as described 😅 )

PR #578, opened 2023-02-10
[iOS] Reapply pending formats after enter in list

### LINKED ISSUE #577: [iOS] Formatted is not systematically applied in new list item
* Create a first formatted list item
* Enter
* Type

Formatting is not applied until next HTML complete replacement.

This is due to the fact that the text view will use latest character traits upon character input, which in case of placeholder NBSP/ZWSP and list prefixes doesn't contain the formatting traits.

aringenbach:
Fixes #577

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/ComposerModel.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/ComposerModel.swift
@@ -55,4 +55,11 @@ extension ComposerModel {
 
         return update
     }
+
+    /// Returns currently reversed (active) actions on the composer model.
+    var reversedActions: Set<ComposerAction> {
+        Set(actionStates().compactMap { (key: ComposerAction, value: ActionState) in
+            value == .reversed ? key : nil
+        })
+    }
 }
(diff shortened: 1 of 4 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/ComposerModel.swift
```diff
@@ -55,4 +55,11 @@ extension ComposerModel {
 
         return update
     }
+
+    /// Returns currently reversed (active) actions on the composer model.
```
jonnyandrew, line 59: > currently reversed (active) actions

Outside the scope of this change, but I find it a bit confusing that 'reversed' is equivalent to 'active'. Should we consider separating the concepts of toggling ('reversing') the state from the state itself (i.e. active/inactive)?

jonnyandrew (approved):

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Extensions/ComposerModel.swift
aringenbach: It can be a bit confusing, but IIRC the reasoning behind this is that the ComposerAction refers to calling the associated function as well. e.g. if bold is reversed, calling `bold` will actually `unbold`. If we flagged this as `active` instead, one could assume that calling `bold` would do nothing since it's already active.

PR #581, opened 2023-02-13
Update HTML/markdown setting functions to return `Result` type

jonnyandrew:
## What?

Update two Rust model functions (`set_content_from_markdown()`,`set_content_from_html()`) to return `Result`.

Note that nothing is fixed by this change and further work is still needed to return errors where appropriate.

### To do
- [x] Update platform code

## Why?

The interface change reflects the fact that these methods may fail under normal conditions and given malformed inputs for example, the model should return an error, not crash.


jonnyandrew:
@aringenbach thanks for your review!

- I've added a test for some known bad HTML [`606dd40`](https://github.com/matrix-org/matrix-rich-text-editor/pull/581/commits/606dd40952fb34fefb99cb89da127b45722542dc). I don't think we have a way yet to trigger failure for Markdown parsing without panicking.
- I've also stopped returning `Result` from the tests [`1853f7d`](https://github.com/matrix-org/matrix-rich-text-editor/pull/581/commits/1853f7d3c4a081059cd02712394e48e682c27522). I'm not 100% sure which is more idiomatic in this case but I can see `unwrap()` is a bit easier to read and maintain.

aringenbach (approved):
LGTM

PR #585, opened 2023-02-14
Improve selection range computation

aringenbach:
* Merge all text that are not part of the HTML under a single attribute (discardable text)
* Improve selection range computation
* Use typed variant of enumerateAttribute when possible
* Fix a Swiftlint issue


jonnyandrew (approved):
Nice, it looks like you've managed to simplify this quite a bit

PR #586, opened 2023-02-15
Fix creating list with selection containing quote / code blocks

aringenbach:
* Fixes issue with list created inside a quote
* Fixes creating a list from a complex selection such as a paragraph + a quote
  * For now quote/code blocks are removed and their content is reused to create a valid list item
  * This behaviour might change in the future if we handle quote/code blocks inside lists ^
* Add Rust tests for this change

### DIFF
--- a/crates/wysiwyg/src/dom/range.rs
+++ b/crates/wysiwyg/src/dom/range.rs
@@ -290,6 +290,8 @@ impl Range {
         self.locations
             .iter()
             .filter(move |l| l.node_handle.depth() == depth)
+            // FIXME: filtering positions that are before start, these shouldn't be returned.
+            .filter(|l| !(l.relative_position() == DomLocationPosition::Before))
     }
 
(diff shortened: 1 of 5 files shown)


crates/wysiwyg/src/dom/range.rs
```diff
@@ -305,6 +305,29 @@ impl Range {
         self.locations_at_depth(self.top_level_depth())
     }
 
+    /// Return true if the current range contains a single top
+    /// level node of optional given kind.
+    pub fn has_single_top_level_node(&self, kind: Option<DomNodeKind>) -> bool {
+        let mut top_level_locations = self
+            .top_level_locations()
+            // FIXME: filtering positions that are before start, these shouldn't be returned from a > 0 range
```
jonnyandrew, line 313: Was it not simple to update this behaviour in `top_level_locations()`?

jonnyandrew (approved):

PR #591, opened 2023-02-21
Detect at/hash/slash pattern and add them to a MenuAction suggestion

aringenbach:
* Provides at/hash/slash patterns detection from Rust
* Uniffi & wasm bindings are available
* When a pattern is detected, it is returned in a new MenuAction item

To fix: 
- [x] Handle complex characters (multiple code units)
- [x] Add more tests to verify the states are forwarded properly
- [x] Improve & optimise current code


crates/wysiwyg/src/tests/test_suggestions.rs
```diff
@@ -0,0 +1,65 @@
...
+    };
+    suggestion.trailing_strategy = TrailingStrategy::ColonSpace;
+    model.set_link_suggestion(
+        "https://matrix.to/#/@alice:matrix.org".into(),
+        "Alice".into(),
+        suggestion,
+    );
+    assert_eq!(
+        tx(&model),
+        "<a href=\"https://matrix.to/#/@alice:matrix.org\">Alice</a>:&nbsp;|",
```
jonnyandrew, line 63: One question for design: is the trailing colon something we're committed to for the new composer?

I always found it a bit awkward when I wanted to mention multiple users before writing my message. For example:

```
@user1, @user2 what are your thoughts?
```

crates/wysiwyg/src/composer_model/menu_action.rs
```diff
@@ -0,0 +1,91 @@
...
+    fn pattern_for_text(mut text: S) -> Option<(PatternKey, String)> {
+        let Some(first_char) = text.pop_first() else {
+            return None;
+        };
+        let Some(key) = PatternKey::from_char(first_char) else {
+            return None;
+        };
+
+        // Exclude if there is inner whitespaces.
+        if text.chars().any(|c| matches!(&c, ' ' | '\x09'..='\x0d')) {
```
jonnyandrew, line 85: Could it be safer to use a built-in Rust utility for detecting whitespace? I saw [`is_whitespace`](https://doc.rust-lang.org/std/primitive.char.html#method.is_whitespace) and the source [list of characters](https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt) that mention a few more characters than the range of `'\x09'..='\x0d'`.

If not, perhaps we can extract this to a named utility that explains why we need to do it in a custom way?

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -48,6 +49,19 @@ where
...
+    pub fn set_link_suggestion(
+        &mut self,
+        link: S,
+        text: S,
+        suggestion: SuggestionPattern,
+    ) -> ComposerUpdate<S> {
+        self.do_replace_text_in(S::default(), suggestion.start, suggestion.end);
+        self.state.start = Location::from(suggestion.start);
+        self.state.end = self.state.start;
+        self.set_link_with_text(link, text);
```
jonnyandrew, line 61: Do you think representing pills internally as a link node is enough, or should we consider something more flexible?

I'm asking because pills might need some custom behavior like removing the whole pill when backspacing from the end. So, would it be better to represent pills internally as an abstract 'pill' and then convert it to a link during HTML rendering?

jonnyandrew (approved):
Looks good to me, just a few comments

crates/wysiwyg/src/composer_model/menu_action.rs
aringenbach: Yeah actually I thought about the same and did not commit the change. Right now we are using this matches macro for both this feature and the backspace/delete word feature, and in both cases I assume it is fair to stop at any character that has the unicode `whitespace` property.

crates/wysiwyg/src/composer_model/hyperlinks.rs
aringenbach: Yeah, we talked about this a bit. It is done that way to ease up parsing. If we e.g. parse HTML to edit an old message including pills, pills in that HTML are just links, and if we want the Rust library to be aware this is represented as a pill, we'll have to add some kind of custom HTML attributes manually before parsing for the library to understand these are not standard links. 
This would mean we'd have to both run some permalink detection before sending the HTML to Rust, and run permalink detection again when Rust sends back the current HTML. And AFAIK it seems that deleting the entire pill with backspace is almost the only custom behaviour required. We are for now trying to implement this behaviour in the platform specific code (on iOS it just works and remove the entire pill already, because we do some selection if needed before backspace in case of multiple code points emojis - in the case of pills, they are a 1-length item in the attributed string, and we will just calculate the real length in HTML and remove that, without any change to the current backspace implementation)

The main idea is to try out without Rust having that knowledge, because Rust shouldn't know if some platform wants to represent these links in some other way.
We might move to Rust having that knowledge if it causes too much problems on a specific platform.

crates/wysiwyg/src/tests/test_suggestions.rs
aringenbach: Yeah that's a good question, I'll leave it at that for now, and we can discuss it with the team.

crates/wysiwyg/src/composer_model/hyperlinks.rs
jonnyandrew: Ok, I see why we'd represent the pill as a link internally :+1: 

About the backspace handling though, is it right that this is the app's responsibility? Let's say an app does not implement it and uses the default behaviour. Am I right that backspacing will result in a misleading link? For example

**Before**
```
<a href="matrix.to/alice">@alice</a>
```
**After two backspaces**
```
<a href="matrix.to/alice">@ali</a> // misleading!
```

aringenbach: Fair point. Well, hopefully most clients will implement it.

In the case where these are represented as standard links, isn't it weird from a composing perspective that a standard link text component can be edited freely and what was supposed to be a mention but is represented in the exact same fashion can't ? And what would happen if I move the cursor inside the text of that mention link ? Do we just don't allow it from the Rust side ? 

Also, same thing could actually happen if the permalink detection doesn't work for some reason (and we can't prevent it in that case).

The other important thing is that, in Matrix, mentions are links, it's not just an internal representation of a pill. I would rather say that Pills are an external/UI representation of a link.
I think we can still see the same kind of issue if we have a Rust internal type for Pills. If a client doesn't care about pill representation, it's fair to assume that it won't provide Rust with permalink detection for parsing either, so anytime we parse the HTML (e.g. editing a message), the Rust internal type for Pills will not be here anyway, and we'll end up with just a standard link.

crates/wysiwyg/src/tests/test_suggestions.rs
aringenbach: Actually done here: https://github.com/matrix-org/matrix-rich-text-editor/pull/591/commits/f8afc7d5b04e282fa7d6f0f637fb40664d6f1229

crates/wysiwyg/src/composer_model/menu_action.rs
aringenbach: I replaced usages of the macro here: https://github.com/matrix-org/matrix-rich-text-editor/pull/591/commits/248077dffa77f09d6ccd176c39c71b16c6c64220

Looking at Alun's helper, I wondered if we cared about Punctuation signs, but at the moment it would break things to try to care about them, because `@` is considered one. And I guess it's fair that a display name could contain some punctuation signs for some reason.

crates/wysiwyg/src/composer_model/hyperlinks.rs
jonnyandrew: I can see the argument that the pill representation is a UI/app concern given that mentions are just links according to the spec.

It could still be useful to have the expected behaviour be the default IMO. But given the edge cases you mentioned, I'd agree we can keep it simple to start and add it later.

jonnyandrew (approved):
Thanks for the changes!

PR #594, opened 2023-02-28
Test maintenance

alunturner:
Some housekeeping work to improve testing, mainly for web:

- Turns the coverage back on - having tried loads of stuff, the answer seems to be just update vitest, which then lead to updating vite and the `@vitest` dependencies. Most of the diff is in the lock file.
- Fixes rust warning that was being shown when running `make web`.
- Suppresses wasm errors to tidy up log output (web tests).


platforms/web/yarn.lock
```diff
@@ -4450,14 +4616,26 @@ vite-plugin-dts@^1.5.0:
...
+
 vitest-sonar-reporter@^0.3.3:
   version "0.3.3"
   resolved "https://registry.yarnpkg.com/vitest-sonar-reporter/-/vitest-sonar-reporter-0.3.3.tgz#10e3761144be2cd8efb4fee88265ff919de67c1d"
   integrity sha512-OvH+AYMgDFk3SK7sohn5KUHUOhB42tWgSTW7YzB/v/dN3x+7bgXD2L0FZpDYjC1hvBMJJ4HFqOsiO9LsETmgxg==
   dependencies:
     pathe "^0.3.9"
 
-vitest@0.23.4, vitest@^0.23.4:
+vitest@0.23.4:
```
florianduros, line 4638: The old version shouldn't be necessary with the upgrade

florianduros (changes requested):
It seems than some old dependencies are still listed in the lock file (esbuild/*, vitest)

platforms/web/yarn.lock
alunturner: To sort this out I think it's necessary to upgrade vite and the react plugin, so I've done that in the most recent commits.

florianduros (approved):
Thanks ! Coverage is baaaack 🥳

PR #596, opened 2023-02-28
Test coverage improvements

alunturner:
Following reactivation of the coverage reports, it showed that some of the recent work was lacking in coverage. This work:

- rectifies some of the bits that were missed by testing in recent work
- makes sonarcloud look at the correct files for source and test files to avoid spurious warnings (implemented due to warnings being raised for duplication in test files)

### DIFF
--- a/platforms/web/lib/useWysiwyg.formatting.test.tsx
+++ b/platforms/web/lib/useWysiwyg.formatting.test.tsx
@@ -312,3 +312,97 @@ describe('link', () => {
         expect(actionsRef.current?.getLink()).toBe('https://mylink.com');
     });
 });
+
+describe('indentation', () => {
+    let textbox: HTMLDivElement;
+
+    beforeEach(async () => {
+        render(<Editor />);
+        textbox = screen.getByRole('textbox');
+        await waitFor(() =>
+            expect(textbox).toHaveAttribute('contentEditable', 'true'),
+        );
+    });
+
+    it('Should not show the indent/unindent buttons when empty', async () => {
+        const indentButton = screen.queryByRole('button', { name: 'indent' });
+        const unindentButton = screen.queryByRole('button', {
+            name: 'unindent',
+        });
+
+        [indentButton, unindentButton].forEach((button) => {
+            expect(button).not.toBeInTheDocument();
+        });
+    });
+
+    // eslint-disable-next-line max-len
+    it('Should show the indent/unindent buttons when a list is reversed', async () => {
+        await userEvent.click(
+            screen.getByRole('button', { name: 'orderedList' }),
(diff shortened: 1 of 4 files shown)


platforms/web/lib/useWysiwyg.formatting.test.tsx
```diff
@@ -66,6 +60,51 @@ describe.each([
...
+    ],
+    [
+        'quote',
+        'enabled',
+        'reversed',
+        '<blockquote><p>foo</p></blockquote>',
+        '<blockquote><p>foo bar</p></blockquote>',
+        '<blockquote><p><del>foo</del></p><p><del>bar</del></p></blockquote>',
+        'foo bar',
+    ],
```
alunturner, line 107: Additional test cases for the newer functions

platforms/web/lib/useWysiwyg.formatting.test.tsx
```diff
@@ -94,25 +133,21 @@ describe.each([
...
+        it(`Should be ${expectedActivationState} after single activation`, async () => {
+            // When
+            await userEvent.click(button);
+
+            // Then
+            expect(button).toHaveAttribute(
+                'data-state',
+                expectedActivationState,
+            );
+        });
```
alunturner, line 146: There were quite a few unnecessary acts and await/waitFor's so I removed as many as I could

platforms/web/lib/useWysiwyg.inputEventProcessor.test.tsx
```diff
@@ -48,41 +48,25 @@ describe('inputEventProcessor', () => {
...
-     * There must be something we should be waiting for or similar but I really
-     * can't figure it out. I also can't see why this test is problematic and
-     * others are not.
-     *
-     * If the test is run on its own, it tends to pass, but if others are
-     * running, it is less reliable. This leads me to suspect there is a timing
-     * issue.
-     */
-    it.skip('Should call inputEventProcessor', async () => {
+    it('Should call inputEventProcessor', async () => {
```
alunturner, line 51: I tried to make this work in what I thought was the spirit of the original test in order to remove the only use of `.skip`.

My guess is that the issues were coming from the fact we had multiple assertions inside the waitFor block (where the callback can be run multiple times).

alunturner:
It seems fitting that, after managing to get the coverage working again, I'm being punished for it by SonarCloud.

There is repetition in the test files in this project, but it always highlights test case array type variables as doing this. @florianduros do we have a way of changing this rule for `.test.` files?

florianduros:
@alunturner To be honest, I don't know

florianduros (approved):

PR #597, opened 2023-03-01
Mentions for web

alunturner:
Integrate a (hardcoded) mention button for the example app and the hook. Add tests.

### DIFF
--- a/platforms/web/src/App.tsx
+++ b/platforms/web/src/App.tsx
@@ -189,7 +189,7 @@ function App() {
                         <button
                             type="button"
                             onClick={(_e) =>
-                                wysiwyg.link('www.link.com', 'some text')
+                                wysiwyg.mention('www.link.com', 'some text')
                             }
                         >
                             Add @mention
--- a/platforms/web/src/App.tsx
+++ b/platforms/web/src/App.tsx
@@ -190,8 +190,8 @@ function App() {
                             type="button"
                             onClick={(_e) =>
                                 wysiwyg.mention(
-                                    'https://matrix.to/#/@alunt:element.io',
-                                    'Alun Turner',
+                                    'https://matrix.to/#/@a_user:element.io',
+                                    'a user',
                                 )
                             }
                         >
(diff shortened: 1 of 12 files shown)


platforms/web/src/App.tsx
```diff
@@ -186,6 +187,19 @@ function App() {
                         <button type="button" onClick={(_e) => wysiwyg.clear()}>
                             clear
                         </button>
+                        {suggestion && (
+                            <button
+                                type="button"
+                                onClick={(_e) =>
+                                    wysiwyg.mention(
+                                        'https://matrix.to/#/@a_user:element.io',
+                                        'a user',
```
jonnyandrew, line 196: nitpick: `a user` looks a bit jarring IMO - maybe we could we go with a 'fake' username e.g. `@alice`?

jonnyandrew (approved):
Is it right that the slash command `/` also triggers a mention (and not just `@` and `#`)?

alunturner:
> Is it right that the slash command `/` also triggers a mention (and not just `@` and `#`)?

I _think_ I'm right in saying it will create a `Suggestion` in the rust state terminology. Is that right @aringenbach ?

platforms/web/src/App.tsx
alunturner: Updated

aringenbach:
> > Is it right that the slash command `/` also triggers a mention (and not just `@` and `#`)?
> 
> I _think_ I'm right in saying it will create a `Suggestion` in the rust state terminology. Is that right @aringenbach ?

Yes indeed, it creates the same kind of suggestion. We have two different new composer updates APIs that uses these, `set_link_suggestion` and `replace_text_suggestion`. It's expected that the client will use the latter when selecting a slash command. Currently there is no mechanism in Rust that would prevent creating a link from a slash suggestion, but perhaps it makes sense to add one ? (and I guess not the other way around because it seems fair to mention users or rooms with plain text rather than links)

PR #598, opened 2023-03-01
Fix leading whitespace getting ignored by parsers

aringenbach:
* Fix issue where parser would ignore leading whitespace that are actually inserted by the user.
* Replace them in the same fashion as trailing whitespaces (nbsp)
* Add tests for both whitespace replacements
* Update some impacted Rust tests

### DIFF
--- a/crates/wysiwyg/src/tests/test_characters.rs
+++ b/crates/wysiwyg/src/tests/test_characters.rs
@@ -316,21 +316,39 @@ fn insert_text_between_line_breaks_in_format_node() {
 }
 
 #[test]
-fn leading_whitespace_is_replace_with_nbsp() {
+fn leading_whitespace_is_replaced_with_nbsp() {
     let model = cm("<p> text|</p>");
     assert_eq!(tx(&model), "<p>&nbsp;text|</p>")
 }
 
 #[test]
-fn trailing_whitespace_is_replace_with_nbsp() {
+fn multiple_leading_whitespaces_are_replaced_with_nbsp() {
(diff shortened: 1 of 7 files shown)


crates/wysiwyg/src/tests/test_characters.rs
```diff
@@ -315,6 +315,24 @@ fn insert_text_between_line_breaks_in_format_node() {
...
+}
+
+#[test]
+fn trailing_whitespace_is_replace_with_nbsp() {
+    let model = cm("<p>text |</p>");
+    assert_eq!(tx(&model), "<p>text&nbsp;|</p>")
+}
+
+#[test]
+fn leading_and_trailing_whitespace_are_both_replaced_with_nbsp() {
```
alunturner, line 331: Do we need to test multiple white spaces and the combinations of multiple spaces (eg a tab with a space or whatever)?

aringenbach: I added tests for multiple whitespaces here: https://github.com/matrix-org/matrix-rich-text-editor/pull/598/commits/51008e1b989b9a9c00095970d13146583904b3fe

Can't add tests for tabulation, line feed or carriage return yet, as these are not handled right now (mostly because I am not sure yet how to fix them, it doesn't seem fair to replace them with `&nbsp;`). I assume we could use something like ZWSP + orignal character, but that would offset the positions, so that is not an easy fix (well, at least we'll have to add some platform specific code that understands we are actually doing that).

PR #599, opened 2023-03-02
Revert vite back to v3, turn coverage back off

alunturner:
The upgrade was fine in this repo, but an internal change to vite between versions 3 and 4 gives us issues when using in element-web.

Issue is due to the nullish coalescing operator no longer being transpiled and I couldn't find a quick fix. Reverting the change to put vite back to v3.

Having tried everything I could find on the web, we still get this issue intermittently. Including, it must be said, when running `yarn test`, so I don't think this is just a coverage thing. For now I've turned the coverage back off to avoid this issue.

### DIFF
--- a/platforms/web/package.json
+++ b/platforms/web/package.json
@@ -51,9 +51,9 @@
         "@types/react-dom": "^18.0.6",
         "@typescript-eslint/eslint-plugin": "^5.35.1",
         "@typescript-eslint/parser": "^5.6.0",
-        "@vitejs/plugin-react": "^3.1.0",
-        "@vitest/coverage-c8": "^0.25.8",
-        "@vitest/ui": "^0.25.8",
+        "@vitejs/plugin-react": "^2.1.0",
+        "@vitest/coverage-c8": "^0.23.4",
+        "@vitest/ui": "^0.23.4",
         "cypress": "^10.9.0",
         "eslint": "^8.23.1",
         "eslint-config-google": "^0.14.0",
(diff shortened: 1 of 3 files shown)


platforms/web/package.json
```diff
@@ -71,9 +71,9 @@
         "react-dom": "^18.2.0",
         "replace": "^1.2.1",
         "typescript": "^4.7.4",
-        "vite": "^4.1.0",
+        "vite": "^3.1.0",
         "vite-plugin-dts": "^1.5.0",
-        "vitest": "^0.25.8",
+        "vitest": "^0.23.4",
```
alunturner, line 76: These all revert to what they were before the upgrade attempt

platforms/web/vite.config.ts
```diff
@@ -59,6 +59,7 @@ export default defineConfig({
         onConsoleLog: (log) => {
             if (log.includes('wasm')) return false;
         },
+        teardownTimeout: 10000,
```
alunturner, line 62: Single line configuration change

jonnyandrew: Is the problem related to vitest-dev/vitest#2008?

Maybe we can add a comment here to remind us to revisit this config later once any related issue is fixed?

jonnyandrew (approved):

alunturner:
@jonnyandrew yes it is, I'll add a comment. I may actually raise something about this with vite as there's a discrepancy between documentation and behaviour (there's a 0 missing, it says the default is 10000 in the documentation vs it's 1000 in the code) which is why I took this value.

alunturner:
Been chasing my tail on this one a bit. Tried almost everything noted in that issue, including upgrading to multiple different versions of vitest, but it still happens intermittently.

Admitting defeat for now and turning the coverage off.

jonnyandrew (approved):

PR #600, opened 2023-03-02
[iOS] Mentions API & enable display in example app

aringenbach:
* Enable mention API in iOS platform code
* Use it in a basic implementation of pills in the example app

### DIFF
--- a/platforms/ios/example/Wysiwyg/Views/WysiwygSuggestionList.swift
+++ b/platforms/ios/example/Wysiwyg/Views/WysiwygSuggestionList.swift
@@ -1,0 +1,81 @@
+//
+// Copyright 2023 The Matrix.org Foundation C.I.C
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+import SwiftUI
+import WysiwygComposer
+
+struct WysiwygSuggestionList: View {
+    @EnvironmentObject private var viewModel: WysiwygComposerViewModel
+    var suggestion: SuggestionPattern
+
+    var body: some View {
+        HStack {
+            VStack(alignment: .leading) {
+                switch suggestion.key {
+                case .at:
+                    let users = Users.filtered(with: suggestion.text)
+                    if !users.isEmpty {
+                        Text(Users.title).underline()
+                        ForEach(users) { user in
+                            Button {
+                                viewModel.setAtMention(link: user.url, text: user.name)
+                            } label: {
+                                HStack(spacing: 4) {
+                                    Image(systemName: user.iconSystemName)
+                                    Text(user.name)
+                                }
+                            }
+                        }
+                    }
+                case .hash:
+                    let rooms = Rooms.filtered(with: suggestion.text)
+                    if !rooms.isEmpty {
+                        Text(Rooms.title).underline()
+                        ForEach(rooms) { room in
+                            Button {
+                                viewModel.setHashMention(link: room.url, text: room.name)
+                            } label: {
+                                HStack(spacing: 4) {
+                                    Image(systemName: room.iconSystemName)
+                                    Text(room.name)
+                                }
+                            }
+                        }
+                    }
+                case .slash:
+                    let commands = Commands.filtered(with: suggestion.text)
+                    if !commands.isEmpty {
+                        Text(Commands.title).underline()
+                        ForEach(Commands.allCases.filter { $0.name.contains("/" + suggestion.text.lowercased()) }) { command in
+                            Button {
+                                viewModel.setCommand(text: command.name)
+                            } label: {
+                                HStack(spacing: 4) {
+                                    Image(systemName: command.iconSystemName)
+                                    Text(command.name)
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            .padding(.horizontal, 8)
+            Spacer()
+        }
+        .overlay(Rectangle().stroke(Color.gray, lineWidth: 1))
+        .padding(.horizontal, 12)
+    }
+}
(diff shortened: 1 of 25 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -203,6 +209,54 @@ public extension WysiwygComposerViewModel {
...
+    /// - Parameters:
+    ///   - link: The link to the user.
+    ///   - name: The display name of the user.
+    func setAtMention(link: String, name: String) {
+        let update: ComposerUpdate
+        if let suggestionPattern, suggestionPattern.key == .at {
+            update = model.setLinkSuggestion(link: link, text: name, suggestion: suggestionPattern)
+        } else {
+            _ = model.setLinkWithText(link: link, text: name)
+            // FIXME: remove this if Rust adds this space for free
```
alunturner, line 225: I thought that the rust model add this at the moment? Or is this replacing an nbsp with a `" "` or something like that?

alunturner (commented):
Just one comment, but probably a good idea to get some more Swifty eyes on this as well

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
aringenbach: The Rust model adds it on the `set_link_suggestion` but not on `set_link_with_text`.
Will make this change in a subsequent PR to avoid more noise on this one.

aringenbach: (on the latter, it just adds nothing)

langleyd (approved):
Just a few minor things, overall it looks really good and was super easy to follow.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
```diff
@@ -62,6 +62,7 @@ public class WysiwygTextView: UITextView {
         guard content.text != attributedText || content.selection != selectedRange else { return }
 
         performWithoutDelegate {
+            WysiwygPillsFlusher.flush()
```
langleyd, line 65: Could you explain the idea of the flusher with a bit more context?

Having it as a singleton, will that mean if we had 2 editors on screen at the same time they could conflict with each other?

aringenbach: Yeah indeed it will have issues, I will definitely remove this as a singleton, it should be singularly attach to a `WysiwygComposerViewModel`/`UITextView`. 

I will add more documentation about that as well, but basically this is a fix for a nasty behaviour on iOS, that doesn't remove the views created by `NSTextAttachmentViewProvider` properly. 

Actually, II was also wondering if I can actually build something where the iOS platform code of the library owns the `NSTextAttachmentViewProvider`, so the hosting app just have to worry about providing the view itself, and doesn't have to care about providing things for the flusher. The main issue with that idea, is that if we build something like that the Pills of the Composer can't share the same UTType as the Pills of the Timeline/legacy composer. So I'm not sure what would be the best here 🤔

aringenbach: Note: The can't share UTType is due to the fact that registering the view provider class is this code: 

```swift
NSTextAttachment.registerViewProviderClass(WysiwygAttachmentViewProvider.self,
                                           forFileType: WysiwygAttachmentViewProvider.pillUTType)
```

So iOS would use the latest provided Provider for a given UTType whatever the kind of UI element we are using.

platforms/ios/example/Wysiwyg/Views/WysiwygSuggestionList.swift
```diff
@@ -0,0 +1,85 @@
...
+                                    Text(room.name)
+                                }
+                            }
+                            .accessibilityIdentifier(room.accessiblityIdentifier)
+                        }
+                    }
+                case .slash:
+                    let commands = Commands.filtered(with: suggestion.text)
+                    // FIXME: we should probably add the suggestion.start test at a deeper level (Rust?)
+                    if !commands.isEmpty, suggestion.start == 0 {
```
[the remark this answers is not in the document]
aringenbach: Hum, maybe, I'm guessing we could try to move some of the logic in the iOS platform library code, but here apart from the `suggestion.start == 0`, this is just the example app filtering its mock objects. 
I guess it could make sense to provide that kind of view model in the library, and make some protocol to represent a user/room/command and have the client provide objects that could conform to it (e.g. make MXUser/MXMember in Element conform)

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -363,6 +417,15 @@ private extension WysiwygComposerViewModel {
...
             break
         }
+
+        switch update.menuAction() {
+        case .keep:
+            break
+        case .none:
+            suggestionPattern = nil
+        case let .suggestion(suggestionPattern: suggestionPattern):
+            self.suggestionPattern = suggestionPattern
```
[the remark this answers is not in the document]
aringenbach: Would have to rename the extracted let though, but yeah might look nicer

platforms/ios/example/Wysiwyg/Views/WysiwygSuggestionList.swift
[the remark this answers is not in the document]
aringenbach: At least, I moved the /command filtering to Rust as it makes sense on all platforms.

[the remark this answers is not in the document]
aringenbach: I wonder if we should bother with my protocol idea up here ^ 
With the current implementation we are giving freedom to the hosting application to define it's own way of filtering. For example an hosting app could choose that searching for pattern "li" should return "Alice" (i.e. what Element-Web does), but another app could consider that it's mandatory to only allow pattern with matching leading characters.

langleyd (approved):
Looks great

PR #601, opened 2023-03-03
Correction to web script

alunturner:
Adds correction to the `update` script in package.json.

The script automates building the bindings, copying the files, then rebuilding the web code.

### DIFF
--- a/platforms/web/package.json
+++ b/platforms/web/package.json
@@ -37,7 +37,7 @@
         "test:watch": "vitest",
         "test:ui": "vitest --ui",
         "coverage": "vitest run --coverage",
-        "update": "cd ../../bindings/wysiwyg-wasm && npm run build && cp -f ./pkg/*wysiwyg* ../../platforms/web/generated && cd ../../platforms/web && yarn"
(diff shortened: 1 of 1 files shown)


aringenbach (approved):
LGTM

platforms/web/package.json
```diff
@@ -37,7 +37,7 @@
         "test:watch": "vitest",
         "test:ui": "vitest --ui",
         "coverage": "vitest run --coverage",
-        "update": "cd ../../bindings/wysiwyg-wasm && npm run build && cp -f ./pkg/*wysiwyg* ../../platforms/web/generated && cd ../../platforms/web && yarn"
+        "update": "cd ../../bindings/wysiwyg-wasm && npm run build && cp -f ./pkg/*wysiwyg* ../../platforms/web/generated && cd ../../platforms/web && yarn build"
```
florianduros, line 40: We should yarn instead of npm to run script since we are using yarn in the RTL

alunturner: I agree that we are in /platforms/web, but it doesn't appear to me that we are inside the /bindings/wysiwyg-wasm folder. It also says in the readme (in that folder) that `npm run build` is what to run there.

Am I missing something here? Not sure if the README should be updated or followed.

florianduros: Oh you're right, npm is used in the `wysiwyg-wasm`folder ! 

You can ignore my comment

PR #602, opened 2023-03-03
Add commands for web

alunturner:
This PR:

- integrates commands into the example app
- maps the suggestions output from the `useWysiwyg` hook to make them easier to use on the JS side of things
- adds test for the new functions relating to suggestions
- adds types for the new `MappedSuggestion` type

### DIFF
--- a/platforms/web/lib/composer.ts
+++ b/platforms/web/lib/composer.ts
@@ -80,6 +80,18 @@ export function processInput(
             }
             break;
         }
+        case 'insertCommand': {
+            if (suggestion && event.data) {
+                return action(
+                    composerModel.replace_text_suggestion(
+                        event.data,
+                        suggestion,
+                    ),
+                    'replace_text_suggestion',
+                );
+            }
+            break;
+        }
         case 'clear':
             return action(composerModel.clear(), 'clear');
         case 'deleteContentBackward':
--- a/platforms/web/lib/testUtils/Editor.tsx
+++ b/platforms/web/lib/testUtils/Editor.tsx
@@ -43,6 +43,7 @@ export const Editor = forwardRef<HTMLDivElement, EditorProps>(function Editor(
             key !== 'removeLinks' &&
             key !== 'getLink' &&
             key !== 'mention' &&
+            key !== 'command' &&
             key !== 'indent' &&
             key !== 'unindent',
     ) as Array<
@@ -53,6 +54,7 @@ export const Editor = forwardRef<HTMLDivElement, EditorProps>(function Editor(
             | 'removeLinks'
             | 'getLink'
             | 'mention'
+            | 'command'
             | 'indent'
             | 'unindent'
         >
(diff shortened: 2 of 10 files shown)


platforms/web/lib/composer.ts
```diff
@@ -80,6 +80,18 @@ export function processInput(
...
+                return action(
+                    composerModel.replace_text_suggestion(
+                        event.data,
+                        suggestion,
+                    ),
+                    'replace_text_suggestion',
+                );
+            }
+            break;
+        }
```
alunturner, line 94: This is to handle the input event

platforms/web/lib/testUtils/Editor.tsx
```diff
@@ -124,6 +126,14 @@ export const Editor = forwardRef<HTMLDivElement, EditorProps>(function Editor(
...
                 add @mention
             </button>
+            <button
+                type="button"
+                onClick={() => {
+                    wysiwyg.command('/test_command');
+                }}
+            >
+                add command
+            </button>
```
alunturner, line 136: All changes in this file are purely for testing purposes

platforms/web/lib/useWysiwyg.formatting.test.tsx
```diff
@@ -455,3 +453,44 @@ describe('mentions', () => {
...
+            data: prefixedInput,
+            inputType: 'insertText',
+        });
+        await userEvent.click(button);
+
+        // Then
+        // nb this information is hardcoded in the button for this test
+        expect(textbox).toContainHTML('/test_command');
+    });
+});
```
alunturner, line 496: Separated this out from the mentions tests, where I initially put it

platforms/web/src/App.tsx
```diff
@@ -200,6 +200,14 @@ function App() {
...
                             </button>
                         )}
+                        {suggestion && suggestion.key === 2 && (
+                            <button
+                                type="button"
+                                onClick={(_e) => wysiwyg.command('/spoiler ')}
+                            >
+                                Add /spoiler command
+                            </button>
+                        )}
```
alunturner, line 210: This could be neatened up with a util to determine the key. Potentially worth exporting too so that it can be used in Element web. Will give that some thought over the weekend.

aringenbach (approved):
LGTM

platforms/web/src/App.tsx
alunturner: I have not added an exported util, I've just slightly changed the suggestion that is exposed to React to include a character and a type to make it easier to handle in element-web

aringenbach (approved):
LGTM

BTW, any idea what changed with sonarcloud ? It seems to complain way more than before, including on main

alunturner:
> LGTM
> 
> BTW, any idea what changed with sonarcloud ? It seems to complain way more than before, including on main

Sonarcloud will now always complain about coverage, because it isn't run with coverage turned on (so it gets a 0).

Unfortunately, the upgrades I did to various pieces to get the coverage running ok stops us being able to use the output from this repo in element web, so I had to revert the changes.

When I get a chance, I'll do that upgrade again as a single job, making sure it works for element web too, then we'll be able to get the coverage turned back on.

PR #609, opened 2023-03-10
Fix plain text mode markdown with Pills

aringenbach:
* API update for PermalinkReplacer (rename link -> url, protocol is now decomposed between HTML parsing requirements and other composer requirements)
* Allow for the client to post process markdown right before it is set in the text view in plain text mode
* Allow for the client to restore altered markdown when switching to RTE
* Plain text mode content is now published in order for the client to implement some pattern detection in plain text mode
* Added Pills representation in plain text mode in the example app & UI tests for it

### DIFF
--- a/platforms/ios/example/Wysiwyg/Pills/WysiwygPermalinkReplacer.swift
+++ b/platforms/ios/example/Wysiwyg/Pills/WysiwygPermalinkReplacer.swift
@@ -15,18 +15,70 @@
 //
 
 import Foundation
-import HTMLParser
 import UIKit
+import WysiwygComposer
 
 final class WysiwygPermalinkReplacer: PermalinkReplacer {
-    func replacementForLink(_ link: String, text: String) -> NSAttributedString? {
+    func replacementForLink(_ url: String, text: String) -> NSAttributedString? {
         if #available(iOS 15.0, *),
-           link.starts(with: "https://matrix.to/#/"),
+           url.starts(with: "https://matrix.to/#/"),
            let attachment = WysiwygTextAttachment(displayName: text,
+                                                  url: url,
                                                   font: UIFont.preferredFont(forTextStyle: .body)) {
             return NSAttributedString(attachment: attachment)
         } else {
             return nil
         }
+    }
+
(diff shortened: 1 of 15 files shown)


platforms/ios/example/Wysiwyg/Pills/WysiwygPermalinkReplacer.swift
```diff
@@ -15,18 +15,70 @@
...
             return NSAttributedString(attachment: attachment)
         } else {
             return nil
         }
     }
+
+    func postProcessMarkdown(in attributedString: NSAttributedString) -> NSAttributedString {
+        // Create a regexp that detects markdown links.
+        let pattern = "\\[([^\\]]+)\\]\\(([^\\)\"\\s]+)(?:\\s+\"(.*)\")?\\)"
+        guard #available(iOS 15.0, *),
```
langleyd, line 37: What's the experience pre-iOS15, they see a link?

platforms/ios/example/WysiwygUITests/WysiwygUITests+PlainTextMode.swift
```diff
@@ -0,0 +1,62 @@
...
+            └>em
+              └>"italic"
+            """
+        )
+        // Re-toggling restores the markdown.
+        button(.plainRichButton).tap()
+        assertTextViewContent("text __bold__ *italic*")
+    }
+
+    func testPlainTextModePreservesPills() throws {
```
langleyd, line 42: 🌟

langleyd (approved):
Looks great

platforms/ios/example/Wysiwyg/Pills/WysiwygPermalinkReplacer.swift
aringenbach: Yeah, exactly.

Come to think maybe when we translate to markdown, we should display plain text display name rather (which means we will lose the link if toggling back to RTE). Right now toggling to plain after setting up a mention pre-iOS 15 would display the link mention in markdown.
Alternatively we could display both in plain text pre-iOS 15 which means people could just edit it like standard text, which is the behaviour of Element-iOS legacy composer.

PR #612, opened 2023-03-13
Fix attributed index computation with multiple pills

aringenbach:
Fixes an issue with multiple Pills inside the composer ending up with broken attributed indexes. 
This issue happened when Pills are close to each other, and the previous algorithm ended up counting out Pills that were actually not before the target position. 
The offset is now calculated along the way and it auto-excludes ranges that end up outside of the target: 

```swift
// Iterate replacement ranges in order and only account those
// that are still in range after previous offset update.
attributedIndex = replacementTextRanges(to: attributedIndex)
    .reduce(attributedIndex) { $1.range.location < $0 ? $0 + $1.offset : $0 }
```

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Range.swift
@@ -117,10 +117,9 @@ extension NSAttributedString {
     /// - Parameter attributedIndex: the index inside the attributed representation
     /// - Returns: Total offset of replacement ranges
     func replacementsOffsetAt(at attributedIndex: Int) -> Int {
-        let range = NSRange(location: 0, length: attributedIndex)
-        return replacementTextRanges(in: range)
-            .map { range.contains($0.range) ? $0.offset : 0 }
-            .reduce(0) { $0 - $1 }
+        replacementTextRanges(to: attributedIndex)
+            .compactMap { $0.range.upperBound <= attributedIndex ? Optional($0.offset) : nil }
+            .reduce(0, -)
     }
 
     /// Computes index inside the HTML raw text from the index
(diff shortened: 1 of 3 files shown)


langleyd:
@aringenbach Not sure I understand what is being fixed. Could you explain this part a bit more(or is there a visual that captures it?):

> counting out Pills that were actually not before the target position.

aringenbach:
Given this composer:

image

If I move on the two positions around the whitespace between the two pills, this is the selection I would get: 

```
2023-03-14 10:52:41.631437+0100 Wysiwyg[38404:24320943] [TextView] Sel(att): {1, 0} | textViewDidChangeSelection(_:)
2023-03-14 10:52:41.631942+0100 Wysiwyg[38404:24320943] [ViewModel] Sel(att): {1, 0} | Sel: {5, 0} | Text: "￼ ￼ " | select(range:)
2023-03-14 10:52:41.632363+0100 Wysiwyg[38404:24320943] [ViewModel] Sel(att): {-3, 0} | Sel: {5, 0} | applySelect(start:end:)
2023-03-14 10:52:41.889565+0100 Wysiwyg[38404:24320943] [TextView] Sel(att): {2, 0} | textViewDidChangeSelection(_:)
2023-03-14 10:52:41.889985+0100 Wysiwyg[38404:24320943] [ViewModel] Sel(att): {2, 0} | Sel: {6, 0} | Text: "￼ ￼ " | select(range:)
2023-03-14 10:52:41.890362+0100 Wysiwyg[38404:24320943] [ViewModel] Sel(att): {-2, 0} | Sel: {6, 0} | applySelect(start:end:)
```

We can see that the offset of the second Pill (which is 1 - 5 => -4) is subtracted from the position (despite being located before it), result is (-2, 0) and (-3, 0) instead of (2, 0) and (1, 0). 
This was due to the fact that we calculate a theoretical attributedIndex, and then apply all the offsets we find inside it in a single operation. What we do know is iterate over the potential offset and only apply while the calculated position is < than the next potential offset.

langleyd (approved):
Thanks for that, makes sense!

PR #613, opened 2023-03-13
[iOS] Disable link action when selection contains/touches Pills

aringenbach:
* Disable link button when a link is replaced by a pill
* Update tests for this use case

### DIFF
--- a/platforms/ios/example/WysiwygUITests/WysiwygUITests+Suggestions.swift
+++ b/platforms/ios/example/WysiwygUITests/WysiwygUITests+Suggestions.swift
@@ -32,6 +32,13 @@ extension WysiwygUITests {
             └>" "
             """
         )
+        // Removing the whitespace afterwards disables the
+        // link button as the caret is right after the pill.
+        app.keys["delete"].tap()
+        XCTAssertFalse(button(.linkButton).isEnabled)
+        // Link button can be re-enabled.
+        app.keys["space"].tap()
+        XCTAssertTrue(button(.linkButton).isEnabled)
     }
 
     func testHashMention() throws {
--- a/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Attributes.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSAttributedString+Attributes.swift
@@ -67,4 +67,33 @@ public extension NSAttributedString {
         let color = attribute(.backgroundColor, at: index, effectiveRange: nil) as? UIColor
         return color ?? .clear
     }
+
+    /// Computes whether given range or its surroundings contains
+    /// a link that has been replaced with something else (e.g.: a pill)
+    ///
+    /// - Parameter range: the range to lookup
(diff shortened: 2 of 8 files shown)


langleyd (approved):
Took a little bit of digging to understand the parts with `originalLength` attribute keys etc, but makes sense.

Maybe we should get together soon to discuss some of this cross-platform and whether there are things we could do to unify the approaches or simplify the index/selection handling.

Or if you have any initial thoughts you could put something down in a Discussion and we could try to progress it async.

aringenbach:
Oh actually `originalLength` could totally be renamed (it's internal so it's not annoying at all), perhaps it makes sense that, even if it is the only actual data we need to wrap it in another object and change the key. Something like that (and rename the key in the same way):

```swift
struct ReplacementContent {
    var originalLength: Int
}
```

It's kinda hard to see how and if we can share some helpers for these mappings, it's strongly tied to the NSAttributedString to HTML translation. On Android, we don't have issues with items like `listPrefixes`, and on Web AFAIK the location inside HTML is very different to an index as it's understood in a string context.

langleyd:
But as an example, if we moved more towards rendering/reconciling directly from the AST -> Attributed string vs HTML -> attributed might it it remove some of this complexity and potentially(I suppose we'd need to investigate) bring the platforms more in line?

aringenbach:
> But as an example, if we moved more towards rendering/reconciling directly from the AST -> Attributed string vs HTML -> attributed might it it remove some of this complexity and potentially(I suppose we'd need to investigate) bring the platforms more in line?

Yeah, if we get to go directly from AST to NSAttributedString, lots of things could change, it will probably require big changes on how selection works to get rid of these kind of computations altogether though. 
Right now, Rust expects that a call to `select` provides a `start` and a `end` which are both integers reflecting an index in the HTML. 

_What I imagine we could do:_

If we completely own how the representation is done, attributes could contain data towards the actual text node we are in. 

So if we imagine this with a Pill:  `"<a href="https://matrix.to/#/@alice:matrix.org">{Alice</a> }|"` with `{}|` representing the selection. Right now we have an attributed selection (0, 2) that gets translated to an HTML selection (0, 6), hence why we need that code. But if we created string attributes and equivalent on other platforms, the selection could be: 
```
start: node [0][0], index: 0 // index 0 in text node "Alice"
end: node [1], index: 1 // index 1 in node " " 
```
We would just have to be smart about replacements, and understand that with a Pill of length 1, the index 1 corresponds to the trailing of "Alice" node which is 5, but we wouldn't have to worry a bit about anything that is in between the two node locations we are giving (nor would we need to care about Pills positioned somewhere before our selection), because Rust knows from the AST what is actually between these two positions. 
Same thing goes the other way around, if Rust returns that the caret is at e.g. `node[1] index: 0` we would just have to find the attribute matching this node, and move the caret there, rather than translating some `HTML index` to the expected `attributed index`. 
I think something like that would be a lot less error-prone, I'm not sure how positively it will affect performance, getting rid of the two operations parsing to and from HTML on each update is a huge deal, but I'm not sure how performant looking up attributes would be (but I got a hunch that it will probably be better overall)

PR #614, opened 2023-03-13
Auto-insert a whitespace after creating a link with text

aringenbach:
Automatically inserts a whitespace after creating a link with text (e.g. using UI elements to create the link from scratch)

Note: we can see in the updated tests (e.g. the updated Android test) that if we do that in the middle of some text where current location is followed by a whitespace already, it looks a bit whacky to insert an additional space. Do we want to lookup the next character and only insert if the following character is either inexistant or something other than a whitespace ?

### DIFF
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
@@ -250,7 +250,7 @@ class EditorEditTextInputTests {
             .check(matches(TextViewMatcher {
                 it.editableText.getSpans<LinkSpan>().isNotEmpty()
             }))
-            .check(matches(withText("a Element to set")))
+            .check(matches(withText("a Element  to set")))
     }
 
     @Test
--- a/crates/wysiwyg/src/composer_model/hyperlinks.rs
+++ b/crates/wysiwyg/src/composer_model/hyperlinks.rs
@@ -58,8 +58,7 @@ where
         self.do_replace_text_in(S::default(), suggestion.start, suggestion.end);
         self.state.start = Location::from(suggestion.start);
         self.state.end = self.state.start;
-        self.set_link_with_text(link, text);
-        self.do_replace_text(" ".into())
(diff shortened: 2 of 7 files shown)


platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
```diff
@@ -250,7 +250,7 @@ class EditorEditTextInputTests {
             .check(matches(TextViewMatcher {
                 it.editableText.getSpans<LinkSpan>().isNotEmpty()
             }))
-            .check(matches(withText("a Element to set")))
+            .check(matches(withText("a Element  to set")))
```
jonnyandrew, line 253: If there is an explicit selection would we still want an extra space to be added?

I think it there are slightly different use cases between inserting a link at the cursor vs highlighting text and replacing with a new link and might want to check this with UX

jonnyandrew (commented):
Code looks good to me!

Just one question about the behaviour.

platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/test/EditorEditTextInputTests.kt
aringenbach: Hum, I suppose it could make sense in all cases where we remove something to not add the space. See my note in the initial comment as well, there are also cases where caret is at: `text | text` where we probably don't want to insert that space.

jonnyandrew: Yeah I'm not sure what the right thing is here - I think worth asking @callumu on it and seeing what other editors are doing

aringenbach: Cool, we do want that for Pills though, right ? @callumu

callumu: Yes correct:
Pills - yes
Links - no

PR #619, opened 2023-03-15
Add mention attributes

alunturner:
This adds `contenteditable=false` and `data-mention-type=...` fields to the mention types when they are added.

Extra work will be required to add the ability for clients to pass the mention type in when calling this function, so that's been added as a comment.

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -220,7 +220,10 @@ impl ComposerModel {
         let link = Utf16String::from_str(&link);
         let text = Utf16String::from_str(&text);
         Arc::new(ComposerUpdate::from(
-            self.inner.lock().unwrap().set_link_with_text(link, text),
+            self.inner
+                .lock()
+                .unwrap()
+                .set_link_with_text(link, text, None),
         ))
     }
 
--- a/crates/wysiwyg/src/composer_model/hyperlinks.rs
+++ b/crates/wysiwyg/src/composer_model/hyperlinks.rs
@@ -101,7 +101,7 @@ where
         let mention_type: Option<S> = match suggestion {
             Some(_sug) => match _sug.key {
                 crate::PatternKey::At => Some("user".into()),
-                crate::PatternKey::Hash => Some("user".into()),
+                crate::PatternKey::Hash => Some("room".into()),
                 crate::PatternKey::Slash => None,
             },
             None => None,
--- a/crates/wysiwyg/src/composer_model/hyperlinks.rs
+++ b/crates/wysiwyg/src/composer_model/hyperlinks.rs
@@ -58,7 +58,7 @@ where
         self.do_replace_text_in(S::default(), suggestion.start, suggestion.end);
         self.state.start = Location::from(suggestion.start);
         self.state.end = self.state.start;
-        self.set_link_with_text(link, text);
+        self.set_link_with_text(link, text, Some(suggestion));
         self.do_replace_text(" ".into())
     }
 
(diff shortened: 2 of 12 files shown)


platforms/web/lib/useWysiwyg.formatting.test.tsx
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 import { fireEvent, render, screen, waitFor } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { createRef, MutableRefObject } from 'react';
+import { preview } from 'vite';
```
aringenbach, line 20: Seems to break React build

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -88,13 +88,14 @@ where
         &mut self,
         link: S,
         text: S,
+        suggestion: Option<SuggestionPattern>,
```
aringenbach, line 91: This makes things look a bit weird, I suppose this should rather have some `data-mention-type` (or whatever we will name it in the future), instead of a `SuggestionPattern`. I suppose we can immediately determine this from the `PatternKey` element in the autocomplete case, and from data provided by the client in other cases.

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -220,7 +220,10 @@ impl ComposerModel {
         let link = Utf16String::from_str(&link);
         let text = Utf16String::from_str(&text);
         Arc::new(ComposerUpdate::from(
-            self.inner.lock().unwrap().set_link_with_text(link, text),
+            self.inner
+                .lock()
+                .unwrap()
+                .set_link_with_text(link, text, None),
```
aringenbach, line 226: Following my comment below, we should probably expose a new Enum for the `data-mention-type` in order for the client to give the composer knowledge about what kind of link it is trying to create.

platforms/web/lib/useWysiwyg.formatting.test.tsx
alunturner: No idea how that got there

crates/wysiwyg/src/composer_model/hyperlinks.rs
alunturner: Yep that makes sense, I'll try and rejig that

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
alunturner: I agree, but think that work will fit better with implementing the ability to pass in the mention type when calling the function

crates/wysiwyg/src/composer_model/hyperlinks.rs
alunturner: I've had a go at that in the most recent commits

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -88,13 +88,26 @@ where
...
         let e = s + text.len();
         let range = self.state.dom.find_range(s, e);
-        self.set_link_in_range(link, range)
+
+        // TODO instead of inferring the type from the suggestion, change the initial function
+        // call from the client to pass in the mention type when creating the link
+        let mention_type: Option<S> = match suggestion {
+            Some(_sug) => match _sug.key {
+                crate::PatternKey::At => Some("user".into()),
+                crate::PatternKey::Hash => Some("user".into()),
```
aringenbach, line 104:
```suggestion
                crate::PatternKey::Hash => Some("room".into()),
```

aringenbach (approved):
Seems fine for now, I guess we can wait to actually now what to expect for the attribute naming etc before creating explicit enums etc

PR #620, opened 2023-03-15
Write outside of links by default when cursor/selection is at the edge

### LINKED ISSUE #615: Uniformise behaviour when writing near links
Description:

When the composer model contains a link, writing at the position right before the link will end up in either writing inside the first text node of the link, or the previous text node. We should uniformise behaviour so it always prioritise writing outside of the link to avoid issues. 

This test currently succeeds:
```rust
#[test]
fn write_right_before_link_with_previous_text() {
    let mut model = cm("text|<a href=\"https://matrix.org\">Matrix</a>");
    model.replace_text("text".into());
    assert_eq!(
        tx(&model),
        "texttext|<a href=\"https://matrix.org\">Matrix</a>",
    )
}
```

This one fails though (even though the caret position seems to be inside the link and it might make sense, it is indeed an ambiguous position):
```rust
#[test]
fn write_right_before_link() {
    let mut model = cm("<a href=\"https://matrix.org\">|Matrix</a>");
    model.replace_text("text".into());
    assert_eq!(
        tx(&model),
        "text|<a href=\"https://matrix.org\">Matrix</a>",
    )
}
```

aringenbach:
Fixes #615 

Also fixes an issue were, when effectively writing outside of the link on edge with a formatting container as sibling, a new text node was created, instead of writing in the last/first text node of the sibling container.

### DIFF
--- a/crates/wysiwyg/src/dom/dom_methods.rs
+++ b/crates/wysiwyg/src/dom/dom_methods.rs
@@ -65,31 +65,42 @@ where
         #[cfg(any(test, feature = "assert-invariants"))]
         self.assert_invariants();
 
-        let extra_len = new_text.len() + 1;
+        let length = new_text.len();
         let range = self.find_range(start, end);
         let (start_block, end_block) =
             self.top_most_block_nodes_in_range(start, &range);
-        let deleted_handles = if range.is_empty() {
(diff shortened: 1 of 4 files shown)


alunturner (approved):
Looks good to me, I think the tests make it nice and clear what this is doing

PR #622, opened 2023-03-16
Allow updating both url and link text

aringenbach:
Fixes https://github.com/matrix-org/matrix-rich-text-editor/issues/616

* Provides tools to allow editing both the url and the text of a link
* Rename link API to use `url` and avoid ambiguity
* Update example apps for both Android & iOS

iOS demo:

https://user-images.githubusercontent.com/80891108/225628909-aee65427-adce-462d-b554-09de409402de.mp4

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -205,42 +205,54 @@ impl ComposerModel {
         Arc::new(ComposerUpdate::from(self.inner.lock().unwrap().redo()))
     }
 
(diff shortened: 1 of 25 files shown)


alunturner (commented):
This looks good to me and I think the rename of `link` to `url` is definitely a good move.

What's the plan with getting this merged in, do you want me to do the web side of things as a separate PR? From reading the PR it seems we'll have to amend some function calls in web, but I haven't worked on links yet so maybe that's an incorrect assumption.

Maybe a good way ahead is for me to do the web side of things so both PRs are approved then merging one after the other to ensure they're both in the same release? Open to suggestions.

PR #623, opened 2023-03-16
Add correct selection and deletion behaviour for mentions

alunturner:
This PR adds the selection/deletion handling behaviour around mentions to the example app, plus unit tests.

I think it's possible that this work could be shifted inside the rust, which may be a good idea but I only realised this after this work was finished.


jonnyandrew:
> it's possible that this work could be shifted inside the rust, which may be a good idea

@alunturner if there is a benefit to centralising this in the Rust model, might it be worth taking that approach sooner rather than later? As merging this change could make it more difficult to achieve later (as I assume we'd need to unpick and revert it in web)

alunturner:
Closing as this behaviour is now being implemented inside the rust model

PR #627, opened 2023-03-17
Enable parsing link attributes & move link button disabling for immutable links to Rust

aringenbach:
* Defines helpers for immutable content in Rust
* Disable link button when an immutable link is in range
* Allow parsing link attributes properly to get them in the model
* Removed the iOS code disabling buttons on Pills locations (this means that we could edit a Pill with the link button, if the link does not contain `contenteditable="false"`, but it should be the hosting application responsibility to provide proper HTML)

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -320,6 +320,16 @@ where
         matches!(self.kind, ContainerNodeKind::Link(_))
     }
 
+    pub fn is_immutable(&self) -> bool {
+        self.attributes()
+            .unwrap_or(&vec![])
+            .contains(&("contenteditable".into(), "false".into()))
+    }
+
+    pub fn is_immutable_link(&self) -> bool {
+        matches!(self.kind, ContainerNodeKind::Link(_) if self.is_immutable())
+    }
+
     pub fn is_list_item(&self) -> bool {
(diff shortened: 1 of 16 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -371,10 +381,15 @@ where
...
+        attributes: Option<Vec<(S, S)>>,
         children: Vec<DomNode<S>>,
         mention_type: Option<S>,
     ) -> Self {
-        let mut attrs = vec![("href".into(), url.clone())];
+        // FIXME: mention_type could be removed, and the hosting application could just provide attributes
+        // this would allow the Rust code to stay as generic as possible, since it should only care abouy
+        // `contenteditable="false"` to implement custom behaviours for immutable links.
+        let mut attrs =
+            attributes.unwrap_or_else(|| vec![("href".into(), url.clone())]);
```
alunturner, line 392: This is a really good idea - this could be used to add classes and styles to links in the markup which I think will be really helpful with implementing avatars on web

alunturner (approved):
This looks good to me

crates/wysiwyg/src/dom/nodes/container_node.rs
aringenbach: One important thing to note, is that these attributes **will** end up in the final HTML in order for message edition to work and re-create pills, so we still need to be careful about what ends up in the HTML or not.

PR #629, opened 2023-03-20
Backspace/delete non-editable links

alunturner:
Add tests for backspace/deletion behaviour in Rust around mention type links plus some code to handle those cases.

The approach is effectively what I had done in the client, ported into rust, namely:

- detect if we're pressing backspace/delete at the edge of/inside a mention
- if we are, expand the selection in the model before continuing on with the delete/backspace flow

Also a small bit of web specific code (and test) to handle the case where the selection is at 0 index and the first thing in the composer is not editable, ie a composer containing a single mention.

### DIFF
--- a/crates/wysiwyg/src/composer_model/delete_text.rs
+++ b/crates/wysiwyg/src/composer_model/delete_text.rs
@@ -46,32 +46,9 @@ where
 {
     pub fn backspace(&mut self) -> ComposerUpdate<S> {
         self.push_state_to_history();
+        self.handle_mention();
+
         let (s, e) = self.safe_selection();
-
-        // if we're inside a non-editable item, like a mention type link, expand the selection to cover the whole
-        // non-editable item and then allow the backspace flow to continue
-        let range = self.state.dom.find_range(s, e);
-
-        let first_leaf = range.locations.iter().find(|loc| {
-            loc.is_leaf() || (loc.kind.is_block_kind() && loc.is_empty())
-        });
-        if let Some(leaf) = first_leaf {
-            let parent_link_item_loc =
-                range.deepest_node_of_kind(Link, Some(&leaf.node_handle));
-            if let Some(link) = parent_link_item_loc {
-                if (self
-                    .state
-                    .dom
-                    .lookup_container(&link.node_handle)
-                    .is_immutable_link())
-                {
-                    self.state.start = Location::from(link.position);
-                    self.state.end =
-                        Location::from(link.position + link.length);
-                }
-            }
-        }
-
         if s == e {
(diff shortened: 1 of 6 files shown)


crates/wysiwyg/src/composer_model/delete_text.rs
```diff
@@ -92,9 +93,40 @@ where
...
+                    .is_immutable_link()
+                {
+                    self.select(
+                        Location::from(link.position),
+                        Location::from(link.position + link.length),
+                    );
+                }
+            }
+        }
+    }
```
alunturner, line 122: I feel like this is pretty clunky but wanted to get it working first. If this approach should be changed or the code moved to a different part of the codebase, let me know. Haven't done rust stuff for a little while.

aringenbach (approved):
Seems like a good starter. The fact that it relies on selection, makes it necessary to be a `ComposerModel` tool, so it's probably in the right place. I wonder though how much effort we want to put in this, because I guess there are some edge cases that probably won't work with this, e.g. (plus obviously the `delete` variants): 

```rust
let mut model = cm("<a contenteditable=\"false\" href=\"https://matrix.org\">Link with spa</a>ce|");
model.backspace_word();
```

```rust
let mut model = cm("<a contenteditable=\"false\" href=\"https://matrix.org\">Cros{s-sel</a>ection}|");
model.backspace();
```
The second one should almost never happen but we might have missing Pills that just get replaced by standard links that should still be immutable.

I'm not sure how complicated it will be to build a Dom Layer function that handles all this without tweaking the selection, but rather as an, `when we remove a part of a text node that is inside an immutable link, remove the entire link`.

alunturner:
> Seems like a good starter. The fact that it relies on selection, makes it necessary to be a `ComposerModel` tool, so it's probably in the right place. I wonder though how much effort we want to put in this, because I guess there are some edge cases that probably won't work with this, e.g. (plus obviously the `delete` variants):
> 
> ```rust
> let mut model = cm("<a contenteditable=\"false\" href=\"https://matrix.org\">Link with spa</a>ce|");
> model.backspace_word();
> ```
> 
> ```rust
> let mut model = cm("<a contenteditable=\"false\" href=\"https://matrix.org\">Cros{s-sel</a>ection}|");
> model.backspace();
> ```
> 
> The second one should almost never happen but we might have missing Pills that just get replaced by standard links that should still be immutable.
> 
> I'm not sure how complicated it will be to build a Dom Layer function that handles all this without tweaking the selection, but rather as a `when we remove a part of a text node that is inside an immutable link, remove the entire link`.

I'd just noticed the first case myself. I'll merge this and see if I can fix that first case

PR #630, opened 2023-03-20
Backspace/delete word into mentions

alunturner:
Adds tests and handling to deal with the case where a user attempts to backspace/delete by word that enters a mention

### DIFF
--- a/crates/wysiwyg/src/composer_model/delete_text.rs
+++ b/crates/wysiwyg/src/composer_model/delete_text.rs
@@ -204,7 +204,7 @@ where
     ) -> ComposerUpdate<S> {
         // we could have entered a non-editable node during this run, if this is the
         // case, we handle it by calling the delete method once which will adjust the
-        // selection to cover that node and then remove it, then stop the deletion
+        // selection to cover that node and then remove it, ending the recursive calls
         if self.cursor_is_inside_non_editable_text_node() {
             return self.delete();
         }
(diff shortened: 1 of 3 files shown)


crates/wysiwyg/src/composer_model/delete_text.rs
```diff
@@ -186,6 +202,12 @@ where
         direction: Direction,
         location: DomLocation,
     ) -> ComposerUpdate<S> {
+        // we could have entered a non-editable node during this run, if this is the
+        // case, we handle it by calling the delete method once which will adjust the
+        // selection to cover that node and then remove it, ending the recursive calls
+        if self.cursor_is_inside_non_editable_text_node() {
+            return self.delete();
```
aringenbach, line 209: This calls `push_state_to_history` which means an additional state for `undo`, I think we do have issues with this within the `delete` implementation, as I just noticed we don't have a `delete` and `do_delete` variants.

We need to have these variant for any bindings-facing API we might use internally as well. 

e.g.:

```
    pub fn delete(&mut self) -> ComposerUpdate<S> {
        self.push_state_to_history();

        self.do_delete()
    }
    
    pub(crate) fn do_delete(&mut self) -> ComposerUpdate<S> {
        [...]
    }
```

(probably something to look more deeply and fix everywhere we might have issues)

alunturner: Good point

aringenbach (approved):
LGTM

PR #636, opened 2023-03-28
Mention delete behaviour fix

alunturner:
I noticed some incorrect behaviour when deleting the second of multiple mentions - it would delete the first letter of the immutable link, not the whole link.

Example of side by side behaviour in video below, before on the left, after on the right:

https://user-images.githubusercontent.com/56027671/228244598-3e76a9f2-4edb-4ff2-aad9-21d5c6c72861.mov

This PR fixes the unwanted behaviour and adds tests to cover it


jonnyandrew (approved):
Looks good to me

PR #639, opened 2023-03-29
Use attributes for mentions

alunturner:
Draft PR to show the work in progress for adding the ability to pass custom attributes into mentions. 

This is required to allow us to implement avatars in web.


alunturner:
Opened too soon - this had multiple test issues that needed addressing

PR #641, opened 2023-03-29
Pass "data-mention-type" to model

langleyd:
- Pass `data-mention-type` to model
- Add `WysiwygMentionType` to clean up the ViewModel api/logic a bit.
- Fix tests


alunturner (approved):
This seems good to me

PR #642, opened 2023-03-30
Fix selection issue

### LINKED ISSUE #638: Text inserted at incorrect position
## Steps to reproduce
- Type four or five lines of text
- Apply the following formatting
  - Line 1: bold
  - Line 2: list
  - Line 3: bold
  - Line 4: list
  - Line 5: list
- Move the cursor to the middle of line 5
- Type any text
- _Note the text is inserted at the wrong position_

## Details
- Platform: Web (Desktop) - not tested on other platforms
- App version: Element Nightly version: 2023032601

## Screen recording

[Screencast from 29-03-23 09:40:16.webm](https://user-images.githubusercontent.com/4940864/228477525-06bd40b9-8f5e-4647-a915-e9fae2d6d624.webm)

alunturner:
Fixes #638 

This has probably been around a little while, since we introduced paragraphs. It was the particular edge case where you have a formatting node in a paragraph node that's a sibling of a list.

Have made the fix and added a test to cover both list cases.

### DIFF
--- a/platforms/web/lib/dom.ts
+++ b/platforms/web/lib/dom.ts
@@ -538,6 +538,8 @@ function isInlineNode(node: Node | ParentNode | null) {
 // _OFFSET_NODE_NAMES array will require the addition of an offset (of one).
 // Nb, whilst lists require this offset, we consider the offset to apply to each
 // list item, the enclosing list type tag does not add an extra offset.
+// We need the enclosing list tags in the array as we also use this check on
+// sibling nodes.
 function isNodeRequiringExtraOffset(node: Node) {
     return EXTRA_OFFSET_NODE_NAMES.includes(node.nodeName || '');
 }
--- a/platforms/web/lib/dom.ts
+++ b/platforms/web/lib/dom.ts
@@ -527,24 +527,18 @@ export function textNodeNeedsExtraOffset(node: Node | null) {
 }
 
 const INLINE_NODE_NAMES = ['EM', 'U', 'STRONG', 'DEL', 'CODE', 'A'];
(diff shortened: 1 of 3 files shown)


platforms/web/lib/dom.ts
```diff
@@ -538,6 +538,13 @@ function isNodeRequiringExtraOffset(node: Node) {
...
     // do not add an extra offset for the ol/ul tag, only for the list items
     // themselves
+
+    // with the paragraph refactor, we could end up in a situation where we have
+    // a p tag containing formatted text next to a list, so if this is the case,
+    // we do want to add an extra offset _if_ the list has any children
+    if (node.nodeName === 'UL' || node.nodeName === 'OL') {
+        return node.hasChildNodes();
+    }
     return EXTRA_OFFSET_NODE_NAMES.includes(node.nodeName || '');
```
jonnyandrew, line 548: Could `EXTRA_OFFSET_NODE_NAMES` be a bit misleading now as it doesn't contain `UL` or `OL`?

jonnyandrew (approved):
Looks good to me!

I see you've added unit tests for the offset checking function but would it be good to have one for the whole behaviour that is fixed?

Something like
- Given some affected HTML
- When cursor is placed at an affected position and type
- Then the text is inserted at the correct position

platforms/web/lib/dom.ts
alunturner: I think that's a fair point. I'll have a look at the comments and naming in this section and see if they can be improved.

alunturner: Thanks for the comment, rereading this I've actually realised I could remove this special case and add the list nodes to the `EXTRA_OFFSET_NODE_NAMES` array. Much clearerf.

PR #645, opened 2023-03-31
Use attributes - contenteditable from client

alunturner:
This PR:

- allows us to pass in attributes that get applied to the mentions
- relies on the client passing in `contenteditable=false` 
- changes the web and mobile bindings
- updates the iOS and web code to accommodate the binding changes (Android is not currently using this functionality so does not require change)

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -227,6 +227,9 @@ impl ComposerModel {
         ))
     }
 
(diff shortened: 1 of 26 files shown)


jonnyandrew (approved):
Looks good to me!

I have one small suggestion which is to add some docs to the public functions as it might not be so obvious how `attributes` should be used.

Also, IIRC, adding the attributes `contenteditable` and `data-mention-type` is a temporary workaround because we shouldn't really need to pollute the message content with this. So I'd say worth adding a TODO note explaining how we'd fix it

jonnyandrew (approved):

PR #651, opened 2023-04-24
Expose markdown content setter

aringenbach:
* Expose markdown setter
* Add tests for matching content with setting HTML or Markdown
* Add tests for `attributedContent` publisher


langleyd (approved):
Looks great

PR #652, opened 2023-04-26
Disable autocorrection if input is a command

aringenbach:
* Disable autocorrection when message starts with one slash
* Add unit tests
* Add UI tests for potential cases not covered by unit tests


langleyd (approved):
Looks great

PR #653, opened 2023-04-26
Move translating mention type to attributes from wrapper to view model

aringenbach:
Minor code move as `ComposerModelWrapper` shouldn't host any logic other than the panic recovery mechanism (plan is to be able to remove it if possible, and just use the bindings-built `ComposerModel` - e.g. if Rust code manages to be 100% reliable).


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygMentionType.swift
```diff
@@ -32,15 +36,14 @@ public extension WysiwygMentionType {
...
-public extension PatternKey {
-    var mentionType: WysiwygMentionType? {
-        switch self {
-        case .at:
-            return .user
-        case .hash:
-            return .room
-        case .slash:
-            return nil
-        }
```
alunturner, line 44: There's a bunch of similar code in web too. I wonder if there would be value in us storing the actual character in the

alunturner (approved):

PR #654, opened 2023-04-27
Fix NBSP not being replaced properly inside code blocks

aringenbach:
* Fixed an issue with NBSP replacement by ZWSP in code blocks
* Reworked post DTCoreText cleanup to be more efficient (browse the AST once instead of twice)
* Add unit tests covering previously failing cases


platforms/ios/lib/WysiwygComposer/Tests/HTMLParserTests/HTMLParserTests.swift
```diff
@@ -67,4 +67,28 @@ final class HTMLParserTests: XCTestCase {
...
+    }
+
+    func testSingleNbspInCodeBlockIsReplacedByZwsp() throws {
+        let html = "<pre><code>\(String.nbsp)</code></pre>"
+        let attributed = try HTMLParser.parse(html: html)
+        XCTAssertEqual(attributed.string, "\(String.zwsp)")
+    }
+
+    func testTrailingCodeBlockNbspIsReplacedByZwsp() throws {
+        let html = "<pre><code>Test\n\(String.nbsp)</code></pre>"
```
alunturner, line 84: Are the cases where we have `\n\nbsp` different to where we have something like `text\nbsp`?

It may not be worth testing, however. Asking just in case it is.

alunturner (approved):
It looks good to me - one q that is non-blocking

platforms/ios/lib/WysiwygComposer/Tests/HTMLParserTests/HTMLParserTests.swift
aringenbach: Technically if my code block is something like `<pre><code>Test </code></pre>`, that space after `Test` is actually a standard space and not an `nbsp` in our current implementation. It does seem fair to create a test for it just in case. Also it made me realise we should only add our line break element if we find `\n`

PR #655, opened 2023-04-27
Add attributes to all link APIs

aringenbach:
* Adds attributes to all link APIs
* Also rename `link` parameter to `url` where it is relevant

Note: API breaking change that will require slight updates on hosting applications code.


alunturner (approved):
This looks good to me

PR #657, opened 2023-04-28
Fix failure on autocorrection tests if the keyboard is hidden on the simulator

aringenbach:
Autocorrection tests that use the delete key can sometimes fail if the simulator is not showing the keyboard. 
This fixes it by using the rawValue instead of looking for the keyboard button.


langleyd (approved):
LGTM

PR #658, opened 2023-04-28
Fix concurrency cancelling unrelated builds

aringenbach:
Change concurrency to avoid issue with current setup cancelling unrelated content. 
Concurrency should now allow to run all `main` builds and allow only a single build on each PR


langleyd (approved):
Makes sense

PR #661, opened 2023-05-03
Upload rust coverage to codecov

langleyd:
Use `cargo-llvm-cov` for coverage and upload to codecov.

Added a couple of flags so we can easily see all unit tests together and platforms separately.


aringenbach (approved):
LGTM

PR #662, opened 2023-05-03
Add iOS unit test coverage upload to codecov

langleyd:
Add iOS unit test coverage upload to codecov


.github/workflows/ios.yml
```diff
@@ -46,3 +46,9 @@ jobs:
     - name: Tests
       working-directory: platforms/ios/example
       run: exec ./ios-tests.sh
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v3
+      with:
+        fail_ci_if_error: true
```
aringenbach, line 53: I suppose it should have `directory: platforms/ios/example/DerivedData` (or something slightly deeper in that hierarchy) since we are saving to a custom derived data folder.

langleyd: Codecov doesn’t support the apple format for coverage. So I have some changes to integrate slather, the same library used on the main app.

.slather.yml
```diff
@@ -0,0 +1,10 @@
+coverage_service: cobertura_xml
+xcodeproj: platforms/ios/example/Wysiwyg.xcodeproj
+scheme: Wysiwyg
+source_directory: platforms/ios/example
+output_directory: platforms/ios/example
+build_directory: platforms/ios/example/DerivedData
+ignore:
+  - Wysiwyg/Mocks/*
+  - Wysiwyg/AppDelegate.swift
```
aringenbach, line 9: I guess these two should be `example/Wysiwyg/...` but should we just exclude the entire `example` directory from coverage stats ? Supposing that all we care about is the `lib` folder which is what is actually extracted to our Swift Package repo.

.github/workflows/ios.yml
```diff
@@ -46,3 +49,20 @@ jobs:
...
+      env:
+        GIT_BRANCH: ${{ steps.get_branch.outputs.branch }}
+        CI_PULL_REQUEST: ${{ github.event.number }}
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v3
+      with:
+        directory: platforms/ios/example
+        fail_ci_if_error: true
+        flags: unittests-ios, unittests
```
aringenbach, line 68: Also I wonder if we want to separate things or slightly change that flag because technically, we currently run both UI and Unit tests on iOS, and the resulting coverage is the merged coverage of the two test suites.

aringenbach (approved):

.slather.yml
langleyd: I think from he examples I saw the ignores were relative to the source_directory above, but will double check this.

But yea you are probably right we probably don't need to do coverage on the sample, as it's probably more valuable to be testing the libraries integration in the real work examples (EI and EIX).

.github/workflows/ios.yml
langleyd: Yea I noticed this later on, initially I thought was only including unit tests.
How would we split them Unit and UI tests? Snapshot fitting under UI tests?

aringenbach: Actually I'm currently wondering if it's worth it to try adding another flag for UI tests. Running them to prevent an issue is definitely good, but in terms of pure line coverage, I'm pretty sure they don't cover much more than the UnitTests (and if they do, I suppose that could be a missing Unit Test in the Package).

.slather.yml
```diff
@@ -0,0 +1,7 @@
+coverage_service: cobertura_xml
+xcodeproj: platforms/ios/example/Wysiwyg.xcodeproj
+source_directory: platforms/ios/example/lib
```
aringenbach, line 3:
```suggestion
source_directory: platforms/ios/lib
```

aringenbach: (not sure what implies each of these directories, but that one is not a valid address :) )

PR #675, opened 2023-05-12
Read attributes from markdown for mentions

alunturner:
When there was the refactor to add the ability to pass attributes in when creating links, this function was missed.

Amending this function to pass through (specified) attributes means that now we will be able to pass a mention from the plain text editor (ie "markdown") into the rich text editor and have the rust model be instantiated containing that mention with the required attributes.


aringenbach (commented):
Looks fine but could you add a couple tests for this use case ? 🙏

alunturner:
> Looks fine but could you add a couple tests for this use case ? 🙏

I'm trying to do this but don't really understand the code structure here. Where would this be tested? It's the greyed out code for this moduel that's confusing me

jonnyandrew (approved):

PR #677, opened 2023-05-15
[Android] Fix display of multiple identical custom link replacements

jonnyandrew:
Fix display of multiple identical custom link replacements.

Previously, when a second link replacement was added, the first disappeared. After the fix, both are displayed as expected.

### Screenshots
| Before | After |
|-|-|
|pill-before|pill-after|

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
@@ -16,6 +16,7 @@ import io.element.android.wysiwyg.view.StyleConfig
 import io.element.android.wysiwyg.view.models.InlineFormat
 import io.element.android.wysiwyg.view.spans.BlockSpan
 import io.element.android.wysiwyg.view.spans.CodeBlockSpan
+import io.element.android.wysiwyg.view.spans.CustomReplacementSpan
 import io.element.android.wysiwyg.view.spans.ExtraCharacterSpan
 import io.element.android.wysiwyg.view.spans.InlineCodeSpan
 import io.element.android.wysiwyg.view.spans.LinkSpan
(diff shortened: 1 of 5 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -314,11 +315,11 @@ internal class HtmlToSpansParser(
             ?: LinkDisplay.Plain
         when(linkDisplay) {
             is LinkDisplay.Custom -> {
-                val span = linkDisplay.customSpan
+                val span = CustomReplacementSpan(linkDisplay.customSpan)
                 replacePlaceholderWithPendingSpan(last.span, span, last.start, text.length, Spanned.SPAN_EXCLUSIVE_EXCLUSIVE)
             }
             LinkDisplay.Pill -> {
-                val span = PillSpan((last.span).link, resourcesHelper.getColor(styleConfig.pill.backgroundColor))
+                val span = PillSpan(resourcesHelper.getColor(styleConfig.pill.backgroundColor))
```
alunturner, line 322: - I don't know kotlin, so perhaps I'm missing something, but I don't think I can see the signature to `PillSpan` changing in this PR. Was this incorrect before and is being updated to the correct signature now? Or am I missing something obvious? Or, option C, am I'm completely misunderstanding the change?

alunturner (approved):
This looks good to me, I've got one question but it's probably just me misunderstanding an unfamiliar language

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
jonnyandrew: I spotted that the `url` field of `PillSpan` was not used so have removed in this PR ([link to that change](https://github.com/matrix-org/matrix-rich-text-editor/pull/677/files#diff-fe66889a904ee3e8f6d18fb80b16848051764003c6f93f88b7471f2edd6fc843L11))

PR #680, opened 2023-05-16
[Android] Support text replacement and keyword display

jonnyandrew:


platforms/android/example/src/main/java/io/element/android/wysiwyg/poc/matrix/MatrixRoomKeywordDisplayHandler.kt
```diff
@@ -0,0 +1,19 @@
+package io.element.android.wysiwyg.poc.matrix
+
+import io.element.android.wysiwyg.display.KeywordDisplayHandler
+import io.element.android.wysiwyg.display.TextDisplay
+
+/**
+ * Convenience implementation of a [KeywordDisplayHandler] that detects Matrix @room
+ * mentions displays them as default pills.
+ */
+class MatrixRoomKeywordDisplayHandler : KeywordDisplayHandler {
```
jonnyandrew, line 10: This is not exposed in the library, only in the example app

alunturner (approved):

PR #683, opened 2023-05-18
Version 2.2.0

jonnyandrew:


langleyd:
Should we run all tests on releases do you think @jonnyandrew ? Either by adding all the platform tag or adding a release tag and checking it?

PR #685, opened 2023-05-22
Disable invariant assertions by default

jonnyandrew:
Disable invariant assertions by default so that they are not run in release builds.

Note
- The invariant assertions will still be run in tests.
- After this change, the invariant assertions will no longer run in iOS and web example apps so follow-up work is needed to fix that.


aringenbach (approved):
LGTM

Issue #679, opened 2023-05-16
Add ability to output html and markdown formatted specifically for message content


alunturner:
We have discussed on multiple occasions the benefits of creating and exposing a function that would allow us to output the html and markdown from the Rust model but _specifically_ for sending as message content. 

This issue records some of the things we think it should do, noting that some of these approaches may change depending on other design decisions:
**HTML output (to attach to formatted body)**
- remove empty `<p>` tags from the message (ie `<p>&nbsp;</p>`), which we currently use to represent an empty paragraph that the user may potentially put their cursor inside whilst editing https://github.com/matrix-org/matrix-rich-text-editor/pull/686
- strip any attributes from `<a>` tags that aren't href (to remove custom attributes used for display) https://github.com/matrix-org/matrix-rich-text-editor/pull/716
- replace `at-room` mentions with the plain text `@room` as this is what the timeline seems to expect https://github.com/matrix-org/matrix-rich-text-editor/pull/716
- replace the inner text of a room mention link with the mxId (think this is what the spec dictates, should this be the same for users?)

**Markdown output (to attach to body)**
- replace mention link representations with just the mxId, prefixed with @ or # as appropriate

There may be more work required, but this is what we've come across so far.

```[tasklist]
### Tasks
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/686
- [ ] Expose new get_content_as_message_html() function from platform libraries
```

jonnyandrew:
> we will add a function that outputs the Rust state as markdown that meets the Matrix spec

Would we need two different ways to get markdown output if mentions will not be represented as links in the Rust model?

alunturner:
> > we will add a function that outputs the Rust state as markdown that meets the Matrix spec
> 
> Would we need two different ways to get markdown output if mentions will not be represented as links in the Rust model?

Hmm. It's an interesting question. Initially I thought so, but perhaps not. It wouldn't affect web implementation I don't think  but I'd be concerned about whether it would affect iOS at all.

PR #686, opened 2023-05-22
Add functions to get content in clean message format

jonnyandrew:
## Changes
- Add `to_message_html()`/`get_content_as_message_html()` functions which render HTML in a format that is suitable for sending as a message.
  - Output empty `<p>` tags as a newline `\n` when asking for message HTML. This behaviour is consistent with the legacy Element message editor.
- _Note: keep `to_html()`/`get_content_as_html()` functions which render HTML in a format that preserves structures and attributes needed during editing._

## Context
- matrix-org/matrix-rich-text-editor#679

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/container_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/container_node.rs
@@ -709,7 +709,7 @@ impl<S: UnicodeString> ContainerNode<S> {
         let name = self.name();
 
         if self.is_empty() {
-            self.fmt_tag_empty(&S::from("br"), formatter, &self.attrs);
+            formatter.push('\n');
         } else {
             self.fmt_tag_open(name, formatter, &self.attrs);
             self.fmt_children_html(
@@ -813,27 +813,6 @@ impl<S: UnicodeString> ContainerNode<S> {
         formatter.push("</");
         formatter.push(name);
         formatter.push('>');
-    }
-
-    fn fmt_tag_empty(
-        &self,
-        name: &S::Str,
-        formatter: &mut S,
-        attrs: &Option<Vec<(S, S)>>,
-    ) {
-        formatter.push('<');
-        formatter.push(name);
-        if let Some(attrs) = attrs {
-            for attr in attrs {
-                let (attr_name, value) = attr;
-                formatter.push(' ');
-                formatter.push(&**attr_name);
-                formatter.push("=\"");
-                formatter.push(&**value);
-                formatter.push('"');
-            }
-        }
-        formatter.push(" />");
     }
 
     fn updated_state(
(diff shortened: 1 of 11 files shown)


crates/wysiwyg/src/dom/to_html.rs
```diff
@@ -25,11 +25,18 @@ where
         buf: &mut S,
         selection_writer: Option<&mut SelectionWriter>,
         state: ToHtmlState,
+        clean: bool,
```
jonnyandrew, line 28: Not sure if it makes more sense to add this as an configuration option on the `ToHtml` trait functions (as I have done here) or to add a new trait.

crates/wysiwyg/src/dom/to_html.rs
```diff
@@ -25,11 +25,18 @@ where
...
     );
 
     fn to_html(&self) -> S {
         let mut buf = S::default();
-        self.fmt_html(&mut buf, None, ToHtmlState::default());
+        self.fmt_html(&mut buf, None, ToHtmlState::default(), true);
+        buf
+    }
+
+    fn to_internal_html(&self) -> S {
```
jonnyandrew, line 37: My thinking was that `to_html` should output the final/best HTML and the new function should be more specialised. I'm not 100% on the naming of the new function...

Other options I considered:
- `to_ast()` / `to_ast_html()` (i.e. abstract syntax tree - though we'd then need to consider the output of `codeblock` which is split into `<pre>`/`<code>`)
- `to_editor_html()` (i.e. the HTML used by the editor UI?)

crates/wysiwyg/src/dom/to_html.rs
jonnyandrew, line 28: I couldn't settle on the naming of this parameter. 'Cleaning' the HTML seemed closest to what we'd be doing. I also considered 'as_internal_representation` as the inverse but it's a bit long.

crates/wysiwyg/src/dom/to_html.rs
jonnyandrew: Settled on `get_content_as_html()` and `get_content_as_message_html()`

So in both cases we're getting the editor content but in the special case we're getting it in a minimal 'message' format

crates/wysiwyg/src/dom/to_html.rs
jonnyandrew: Updated to `as_message` (related to https://github.com/matrix-org/matrix-rich-text-editor/pull/686#discussion_r1203765574)

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -661,30 +686,56 @@ impl<S: UnicodeString> ContainerNode<S> {
...
+        formatter: &mut S,
+        selection_writer: Option<&mut SelectionWriter>,
+        state: ToHtmlState,
+    ) {
+        let as_message = true;
+        assert!(matches!(self.kind, ContainerNodeKind::Paragraph));
+        let name = self.name();
+
+        if self.is_empty() {
+            formatter.push('\n');
```
jonnyandrew, line 712: Format empty paragraphs as newlines, behaviour consistent with the legacy Element composer.

PR #691, opened 2023-05-24
Add ContainerNodeKind::Mention

alunturner:
This PR adds a mention type so that we can differentiate between links and mentions. Opening for CI checks.

This PR:
- amends `set_link` and `set_link_with_text` so that these functions no longer need to be passed an `attributes` argument (and updates the binding functions for web and mobile to account for this change of signature)
- renames `set_link_suggestion` to `set_mention_from_suggestion`
- makes it so that creating a mention automatically adds a `contenteditable=false` attribute to the mention and adjusts the test cases as required

Recommend to look at this commit by commit.

### DIFF
--- a/crates/wysiwyg/src/dom/insert_parent.rs
+++ b/crates/wysiwyg/src/dom/insert_parent.rs
@@ -129,10 +129,10 @@ mod test {
         let (start, end) = model.safe_selection();
         let range = model.state.dom.find_range(start, end);
 
-        model.state.dom.insert_parent(
-            &range,
-            DomNode::new_link(utf16("link"), vec![], vec![]),
-        );
+        model
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(utf16("link"), vec![]));
 
         assert_eq!(model.state.dom.to_html(), r#"A<a href="link">B</a>C"#)
     }
@@ -143,10 +143,10 @@ mod test {
         let (start, end) = model.safe_selection();
         let range = model.state.dom.find_range(start, end);
 
-        model.state.dom.insert_parent(
-            &range,
-            DomNode::new_link(utf16("link"), vec![], vec![]),
-        );
+        model
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(utf16("link"), vec![]));
 
         assert_eq!(model.state.dom.to_html(), r#"<a href="link">AB</a>C"#)
     }
@@ -157,10 +157,10 @@ mod test {
         let (start, end) = model.safe_selection();
         let range = model.state.dom.find_range(start, end);
 
-        model.state.dom.insert_parent(
-            &range,
-            DomNode::new_link(utf16("link"), vec![], vec![]),
-        );
+        model
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(utf16("link"), vec![]));
 
         assert_eq!(model.state.dom.to_html(), r#"A<a href="link">BC</a>"#)
     }
@@ -171,10 +171,10 @@ mod test {
         let (start, end) = model.safe_selection();
         let range = model.state.dom.find_range(start, end);
 
-        model.state.dom.insert_parent(
-            &range,
-            DomNode::new_link(utf16("link"), vec![], vec![]),
-        );
+        model
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(utf16("link"), vec![]));
 
         assert_eq!(model.state.dom.to_html(), r#"<a href="link">ABC</a>"#)
     }
@@ -185,10 +185,10 @@ mod test {
         let (start, end) = model.safe_selection();
         let range = model.state.dom.find_range(start, end);
 
-        model.state.dom.insert_parent(
-            &range,
-            DomNode::new_link(utf16("link"), vec![], vec![]),
-        );
+        model
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(utf16("link"), vec![]));
 
         assert_eq!(
             model.state.dom.to_html(),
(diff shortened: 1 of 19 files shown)


crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -326,8 +328,9 @@ where
             .contains(&("contenteditable".into(), "false".into()))
     }
 
-    pub fn is_immutable_link(&self) -> bool {
+    pub fn is_immutable_link_or_mention(&self) -> bool {
         matches!(self.kind, ContainerNodeKind::Link(_) if self.is_immutable())
+            || matches!(self.kind, ContainerNodeKind::Mention(_))
```
alunturner, line 333: This can probably change as we will no longer have the ability to create  an immutable `ContainerNodeKind::Link`, that is now a `ContainerNodeKind::Mention`

crates/wysiwyg/src/dom/insert_parent.rs
```diff
@@ -129,10 +129,10 @@ mod test {
...
         let range = model.state.dom.find_range(start, end);
 
-        model.state.dom.insert_parent(
-            &range,
-            DomNode::new_link(utf16("link"), vec![], vec![]),
-        );
+        model
+            .state
+            .dom
+            .insert_parent(&range, DomNode::new_link(utf16("link"), vec![]));
```
alunturner, line 135: All changes in this test file are changes due to function signature change.

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -269,19 +269,39 @@ mod sys {
...
-                .collect();
-
-            DomNode::Container(ContainerNode::new_link(
-                child.get_attr("href").unwrap_or("").into(),
-                Vec::new(),
-                attributes,
-            ))
+
+            // initial implementation, firstly check if we have either `contenteditable=false` or `data-mention-type=`
+            // attributes, if so then we're going to add a mention instead of a link
```
alunturner, line 273: I went with this initially to avoid multiple test case failures. I actually think it will be better to use just the presence of `data-mention-type` to determine if a link is a mention or a plain link. I do this further down, just wanted to make sure all the tests were passing for the time being

crates/wysiwyg/src/tests/test_deleting.rs
```diff
@@ -887,7 +887,7 @@ fn backspace_mention_multiple() {
     model.backspace();
     assert_eq!(
         restore_whitespace(&tx(&model)),
-        "<a contenteditable=\"false\" href=\"https://matrix.org\">first|</a>"
+        "<a href=\"https://matrix.org\" contenteditable=\"false\">first|</a>"
```
alunturner, line 890: All fixes in this file are just attribute ordering changes

crates/wysiwyg/src/tests/test_links.rs
```diff
@@ -846,30 +772,3 @@ fn replace_text_right_after_link_with_next_formatted_text() {
...
-    model.set_link_with_text(
-        "https://matrix.org".into(),
-        "link".into(),
-        vec![("customattribute".into(), "customvalue".into())],
-    );
-    assert_eq!(
-        tx(&model),
-        "<a customattribute=\"customvalue\" href=\"https://matrix.org\">link|</a>"
-    )
-}
```
alunturner, line 875: We no longer do this, these things are more accurately represented as mentions

platforms/web/lib/composer.ts
```diff
@@ -73,17 +73,14 @@ export function processInput(
         case 'insertSuggestion': {
             if (suggestion && isSuggestionEvent(event)) {
                 const { text, url, attributes } = event.data;
-                const defaultMap = new Map();
-                defaultMap.set('contenteditable', 'false');
-                Object.entries(attributes).forEach(([key, value]) => {
-                    defaultMap.set(key, value);
-                });
+                const attributesMap = new Map(Object.entries(attributes));
+
```
alunturner, line 77: No longer a requirement to pass in `contenteditable`

crates/wysiwyg/src/dom/nodes/container_node.rs
```diff
@@ -418,28 +418,6 @@ where
         }
     }
 
-    pub fn new_mention(
```
alunturner: For some reason got repeated in the merge

crates/wysiwyg/src/dom/nodes/dom_node.rs
```diff
@@ -138,16 +138,6 @@ where
         ))
     }
 
-    pub fn new_mention(
```
alunturner: Accidental repetition in the merge

jonnyandrew (commented):
I have only had a quick look but a few comments:
- should a Mention be a container node? It cannot contain other elements and it doesn't behave like a container in the UI (can't be split for example).
- do we need to consider `@room` mentions too?
- related to the above two points, I wonder if we might run into issues if we treat mentions too much like links when they don't behave like links

alunturner:
> I have only had a quick look but a few comments:
> 
> * should a Mention be a container node? It cannot contain other elements and it doesn't behave like a container in the UI (can't be split for example).
I mean it _can_ contain a text node, like a link, but I think it's a fair point about not being able to split it. Because we were initially treating mentions as immutable links I went this way to avoid breaking selection/deletion logic but I think this is a good point. I'll have ago at creating a completely new node type today.

> * do we need to consider `@room` mentions too?
Another good point, as these would only have a placeholder href

> * related to the above two points, I wonder if we might run into issues if we treat mentions too much like links when they don't behave like links

Will update the PR when I've tried shifting to a completely new node type and have got it compiling and passing tests.

alunturner:
Having made an attempt to split off into a new node type, I'm not sure that is a better approach than this, given the complexity of adding new `DomNode` and `DomNodeKind` types. Attempting to progress this PR and get it tidied up a bit now

alunturner:
Closing as superseded by implementation of MentionNode as a separate node

Issue #692, opened 2023-05-24
Parse Matrix mentions from HTML


jonnyandrew:
## Room and user mentions
When setting the content of the editor from HTML, the editor should parse links matching the Matrix permalink format as mentions.

A URL is a [mention](https://spec.matrix.org/v1.6/client-server-api/#user-and-room-mentions) if it has the [Matrix URI scheme](https://spec.matrix.org/v1.6/appendices/#matrix-uri-scheme) or is a [Matrix.to link](https://spec.matrix.org/v1.6/appendices/#matrixto-navigation):
- Matrix URI scheme: `matrix:[//{authority}/]{type}/{id without sigil}[/{type}/{id without sigil}...][?{query}][#{fragment}]`
- Matrix.to: `https://matrix.to/#/<identifier>[/<extra parameter>][?<additional arguments>]`

Apps also seem to create and accept custom matrix.to style mentions:
- Custom app mention: `https://<app-host>/#/<type>/<identifier>[/<optional-event-id>][?<optional-query>]`

where
- `<app-host>` is one of a set of supported hosts defined by the host application (e.g. `app.element.io`)
- `<type>` is one of `room`, `user` or `group` (see [MatrixToConverter](https://github.com/vector-im/element-android/blob/d5d5c9009b69c8fb6f28472e3b9fce41f51205a8/matrix-sdk-android/src/main/java/org/matrix/android/sdk/api/session/permalinks/MatrixToConverter.kt#L24)).

## Whole room mentions (`@room`)
When setting the content of the editor from HTML, the editor should parse text that matches the pattern `@room` as a mention.

jonnyandrew:
Most of the functionality required here should already be part of [ruma](https://github.com/ruma/ruma) (see [matrix_uri.rs](https://github.com/ruma/ruma/blob/main/crates/ruma-common/src/identifiers/matrix_uri.rs))

alunturner:
Addressed in 1bf905db7aff3a11cb171eebce5bb63d9897e0cd

PR #695, opened 2023-05-25
Add Matrix mentions utils [wip]

jonnyandrew:
This PR should add utilities to help with creating Matrix spec compliant mention HTML.

- Part of https://github.com/matrix-org/matrix-rich-text-editor/issues/692

As input, functions might accept:
- a Matrix mention URL
- a potentially ambiguous display name
- any room aliases that are available

As output, functions should return:
- a valid mention URL and anchor/body text combination
- any other useful information required by the library (e.g. the mention type required by matrix-org/matrix-rich-text-editor#709)

### Reference

Valid mention URL and anchor/body text combinations according to the Matrix spec:

> * When mentioning users, use the user’s potentially ambiguous display name for the anchor’s text. If the user does not have a display name, use the user’s ID.
> * When mentioning rooms, use the canonical alias [fallling back to alias then room ID] for the room. In all cases, use the alias/room ID being linked to as the anchor’s text.

> The text component of the anchor should be used in the event’s `body` where the mention would normally be represented, as shown in the example above.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,15 @@
 version = 3
 
 [[package]]
+name = "aho-corasick"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "67fc08ce920c31afb70f013dcce1bfc3a3195de6a228474e45e1f145b36f8d04"
+dependencies = [
+ "memchr",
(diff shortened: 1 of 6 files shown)


jonnyandrew:
Closing in favor of #717

PR #698, opened 2023-05-30
Add mention insertion methods

alunturner:
This change makes breaking changes to the API. This PR incorporates changes to web, iOS and Android to account for the selection logic changes required and the use of the new lib functions.

This PR:

- changes name of internal function to `set_mention_from_suggestion` and updates bindings
- expands some of the ffi tests for this function
- exposes a new function in the bindings files
- creates a new `mentions.rs` file to handle the renamed function
- creates a new `insert_node_at_cursor.rs` file to allow insertion of a node at a cursor with unit tests
- adds extensive tests for mentions in `test_mentions.rs`
- tweaks behaviour of suggestion monitoring to disallow it inside links
- implements the corresponding selection changes required in web due to mention nodes now having length 1
[- implements iOS changes](https://github.com/matrix-org/matrix-rich-text-editor/pull/698/commits/a3cadae066c83d6bcf8f926d55921f8352968b69)
[- implements Android changes ](https://github.com/matrix-org/matrix-rich-text-editor/pull/698/commits/cd8b8397b09bb29207b0da56bc338e112b2506dc)


crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -59,15 +59,84 @@ where
...
+
+        // do nothing if we don't have a cursor
+        // TODO determine behaviour if we try to insert when we have a selection
+        if range.is_selection() {
+            ComposerUpdate::<S>::keep();
+        }
+
+        // manipulate the state of the dom as required
+        if range.is_empty() {
+            // this happens if we replace the whole of a single text node in the composer
```
jonnyandrew, line 101: I'm a bit confused about what this means. Wouldn't an empty range just imply there are no nodes in the DOM?

jonnyandrew (commented):
Looks great so far, and I can follow the logic of the new `insert` method.

Just a few comments about where this fits in the project.

crates/wysiwyg/src/composer_model/hyperlinks.rs
```diff
@@ -59,15 +59,84 @@ where
...
         self.do_replace_text_in(S::default(), suggestion.start, suggestion.end);
         self.state.start = Location::from(suggestion.start);
         self.state.end = self.state.start;
-        self.set_mention_with_text(url, text, attributes);
-        self.do_replace_text(" ".into())
+
+        self.set_mention_with_text(url, text, attributes)
+    }
+
+    fn set_mention_with_text(
```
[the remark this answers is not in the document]
alunturner: Yep - there's definitely rearrangement to be done in terms of function locations, just wanted to get it working to start off with

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -284,7 +284,7 @@ impl ComposerModel {
             self.inner
                 .lock()
                 .unwrap()
-                .set_link_suggestion(url, text, suggestion, attrs),
+                .set_mention_from_suggestion(url, text, suggestion, attrs),
```
jonnyandrew, line 287: The calling function is still called `set_link_suggestion` - is that right?

crates/wysiwyg/src/dom/insert_node_at_range.rs
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            self.insert_node_at_selection(range, new_node)
+        }
+    }
+
+    fn insert_node_at_cursor(
+        &mut self,
+        range: &Range,
+        new_node: DomNode<S>,
+    ) -> DomHandle {
```
jonnyandrew, line 39: Currently this method could be called with a range that is not a cursor. Is there any way we can make this function safer and not allow that to happen? I'm thinking you could change it to accept the cursor location for example:

```rust
    fn insert_node_at_cursor(
        &mut self,
        cursor: Option<&Location>,
        new_node: DomNode<S>,
    )
```

crates/wysiwyg/src/dom/insert_node_at_range.rs
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    fn insert_node_at_selection(
+        &mut self,
+        range: &Range,
+        mut new_node: DomNode<S>,
+    ) -> DomHandle {
+        // TODO
+        return DomHandle::new_unset();
```
jonnyandrew, line 94: I think we should either introduce the whole implementation as part of this change or limit the functionality to `insert_node_at_cursor`

crates/wysiwyg/src/tests/test_mentions.rs
jonnyandrew, line 1: Could you add a test or two that documents the behaviour when there is a selection?

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -0,0 +1,69 @@
...
+
+        self.push_state_to_history();
+
+        let handle = self.state.dom.insert_node_at_range(&range, new_node);
+
+        // manually move the cursor to the end of the mention
+        self.state.start = Location::from(new_cursor_index);
+        self.state.end = self.state.start;
+
+        // add a trailing space in cases when we do not have a next sibling
```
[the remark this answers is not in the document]
jonnyandrew: Personally I think the behaviour you've implemented makes sense but could be worth checking with UX if you're not sure

jonnyandrew (commented):
Looks great and thanks for the comprehensive tests

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
alunturner: It is at the moment yes as I wasn't 100% on `set_mention_from_suggestion`. I feel like what we're actually doing is `replace_suggestion_with_mention` or `insert_mention_at_suggestion` or something like that.

alunturner: It is at the moment yes as I wasn't 100% on the name `set_mention_from_suggestion`. I feel like what we're actually doing is `replace_suggestion_with_mention` or `insert_mention_at_suggestion` or something like that.

crates/wysiwyg/src/dom/insert_node_at_range.rs
alunturner: I don't think it can as I think the only exposed function is `insert_node_at_range` which then checks if it's a cursor before calling `insert_node_at_cursor`. Could be wrong though.

crates/wysiwyg/src/dom/insert_node_at_range.rs
alunturner: Functions have been amended to handle the selection case and amended appropriately, with new tests added

alunturner: I took the following approach:

- if user tries to insert a mention to replace a suggestion, remove the suggestion, then try to insert a mention
- if user tries to insert a mention directly (like by clicking on a name), first check if we have a selection or not
    - if we have a selection, remove it
    - if we have a cursor, do nothing
    - then insert a mention at the cursor position

alunturner (commented):
Setting back to draft whilst I figure out the selection changes required for web

alunturner (commented):
Setting back to draft whilst I figure out the selection changes required for web

bindings/wysiwyg-ffi/src/ffi_composer_update.rs
```diff
@@ -121,35 +121,118 @@ mod test {
...
-                },
-                Attribute {
-                    key: "data-mention-type".into(),
-                    value: "user".into(),
-                },
-            ],
+            vec![Attribute {
+                key: "data-mention-type".into(),
+                value: "user".into(),
+            }],
```
jonnyandrew, line 234: We should soon be able to remove this so could add a TODO here (issue https://github.com/matrix-org/matrix-rich-text-editor/issues/709)

platforms/web/lib/dom.test.ts
jonnyandrew, line 1: Same as above, we could link a TODO for these tests https://github.com/matrix-org/matrix-rich-text-editor/issues/709

jonnyandrew (commented):
Looks good to me, but we'll need to update all platform code as part of this PR as it is a breaking change

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -0,0 +1,96 @@
...
+    ) -> ComposerUpdate<S> {
+        let (start, end) = self.safe_selection();
+        let range = self.state.dom.find_range(start, end);
+
+        if range.locations.iter().any(|l: &DomLocation| {
+            l.kind.is_link_kind() || l.kind.is_code_kind()
+        }) {
+            return ComposerUpdate::keep();
+        }
+
```
aringenbach, line 61: `self.push_state_to_history();`

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -0,0 +1,96 @@
...
+        self.insert_mention(url, text, attributes)
+    }
+
+    /// Inserts a mention into the composer. It uses the following rules:
+    /// - If the selection or cursor contains/is inside a link, do nothing (see
+    /// https://github.com/matrix-org/matrix-rich-text-editor/issues/702)
+    /// - If the composer contains a selection, remove the contents of the selection
+    /// prior to inserting a mention at the cursor.
+    /// - If the composer contains a cursor, insert a mention at the cursor
+    pub fn insert_mention(
```
aringenbach, line 47: I'm guessing this function should be exposed to Uniffi / WASM for creating mentions in place.

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -0,0 +1,96 @@
...
+        text: S,
+        attributes: Vec<(S, S)>,
+    ) -> ComposerUpdate<S> {
+        let (start, end) = self.safe_selection();
+        let range = self.state.dom.find_range(start, end);
+
+        let new_node = DomNode::new_mention(url, text, attributes);
+        let new_cursor_index = start + new_node.text_len();
+
+        self.push_state_to_history();
```
aringenbach, line 81: We should probably never `push_state_to_history` inside inner variant `do_...` in case of reuse somewhere else (since it creates an undo/redo state - it should always be called in the public API function)

crates/wysiwyg/src/composer_model/mentions.rs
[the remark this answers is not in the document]
aringenbach: Yeah this was already discussed at some point, the conversation lies somewhere in our Room history, but yeah adding a space after the inserted trailing function only is the way to go.

bindings/wysiwyg-ffi/src/ffi_composer_update.rs
```diff
@@ -121,35 +121,120 @@ mod test {
     }
 
     #[test]
-    fn test_set_link_suggestion_ffi() {
-        let model = Arc::new(ComposerModel::new());
+    fn test_replace_whole_suggestion_with_mention_ffi() {
+        let mut model = Arc::new(ComposerModel::new());
         let update = model.replace_text("@alic".into());
```
aringenbach, line 126: `update` is unused now

alunturner:
I've made those changes @aringenbach . Extracting the logic check to a util is probably a good move too as we're going to change that behaviour when issue #702 gets addressed anyway.

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -0,0 +1,115 @@
...
+        if self.should_not_insert_mention() {
+            return ComposerUpdate::keep();
+        }
+
+        self.replace_text_in(S::default(), suggestion.start, suggestion.end);
+        self.state.start = Location::from(suggestion.start);
+        self.state.end = self.state.start;
+
+        self.push_state_to_history();
+        self.do_insert_mention(url, text, attributes)
```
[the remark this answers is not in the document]
aringenbach: I'm guessing it is rather this, otherwise `undo` will not restore the text that got removed.

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -0,0 +1,115 @@
...
+        if self.should_not_insert_mention() {
+            return ComposerUpdate::keep();
+        }
+
+        if self.has_selection() {
+            self.replace_text(S::default());
+        }
+
+        self.push_state_to_history();
+        self.do_insert_mention(url, text, attributes)
```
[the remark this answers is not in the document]
aringenbach: Same here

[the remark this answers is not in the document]
aringenbach: To be honest sometimes I wonder if we should wrap this `ComposerModel` into another object that manages the undo stack and has all the public/FFI APIs, because it seems there are a lot very easy ways to push multiple states to history when we shouldn't.

bindings/wysiwyg-ffi/src/wysiwyg_composer.udl
```diff
@@ -47,7 +47,7 @@ interface ComposerModel {
     ComposerUpdate unindent();
     ComposerUpdate set_link(string url, sequence<Attribute> attributes);
     ComposerUpdate set_link_with_text(string url, string text, sequence<Attribute> attributes);
-    ComposerUpdate set_link_suggestion(string url, string text, SuggestionPattern suggestion, sequence<Attribute> attributes);
+    ComposerUpdate insert_mention_at_suggestion(string url, string text, SuggestionPattern suggestion, sequence<Attribute> attributes);
     ComposerUpdate remove_links();
```
aringenbach, line 51:
```suggestion
    ComposerUpdate remove_links();
    ComposerUpdate insert_mention(string url, string text, sequence<Attribute> attributes);
    ComposerUpdate insert_mention_at_suggestion(string url, string text, SuggestionPattern suggestion, sequence<Attribute> attributes);
```

crates/wysiwyg/src/composer_model/mentions.rs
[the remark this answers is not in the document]
alunturner: Maybe I'm misunderstanding how this undo stuff works, but the `replace_text*` function variants all seem to call `self.push_state_to_history` internally. So I went with these functions only using the `replace_text*` variants (so they handle the undo stack) then only adding something to the undo stack immediately before we insert the mention. Perhaps I don't need to add something to the undo stack before we insert the mention as it probably makes sense for the user to be able to hit undo to go back to _before_ any of the text replacement happened, not immediately before the mention was inserted (but after text removal may have happened).

aringenbach (approved):
LGTM ! I also triggered the mobile platforms tests that seem to be all green.

PR #700, opened 2023-06-01
Android: Allow pasting of non text content

langleyd:
- Only special-case onTextContextMenuItem paste behaviour if it is text that is pasted, otherwise default to EditText implementation.
- Add tests for pasting of images, plain text and html

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/UriContentListener.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/UriContentListener.kt
@@ -1,0 +1,29 @@
+package io.element.android.wysiwyg.inputhandlers
+
+import android.content.ClipData
+import android.net.Uri
+import android.view.View
+import androidx.core.view.ContentInfoCompat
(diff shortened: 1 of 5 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/UriContentListener.kt
```diff
@@ -0,0 +1,29 @@
+package io.element.android.wysiwyg.inputhandlers
+
+import android.content.ClipData
+import android.net.Uri
+import android.view.View
+import androidx.core.view.ContentInfoCompat
+import androidx.core.view.OnReceiveContentListener
+
+class UriContentListener(
```
jonnyandrew, line 9: If this is only used as a test utility, could you move it into the test directory?

(Or if it is supposed to be exposed as a library utility, then we should add some documentation)

jonnyandrew (commented):
Looks good to me, just one comment about the `UriContentListener`

jonnyandrew (approved):

PR #701, opened 2023-06-01
Add docs to example format selection writer

jonnyandrew:
## Changes

- Add documentation to example format selection writer (used in tests)
- Improve naming within selection writer code

## Context

This part of the code base has caused some confusion for a maintainers including myself and could do with refactoring. As a stop-gap I've clarified some of the naming and added documentation.


jonnyandrew:
> Hang on a minute - code cov is showing an improvement. Something fishy there.

I suspect it's something related to the fact that Android and iOS tests are not running here but they are on the main branch. I have created an issue:

- https://github.com/matrix-org/matrix-rich-text-editor/issues/703

PR #706, opened 2023-06-05
Update Android to use new mention insertion APIs

jonnyandrew:
## Changes

- Update Android library to use mention insertion API added in matrix-org/matrix-rich-text-editor#698
- Ensure character index mapping between the editor UI and composer model works. As mentions are treated as a single character in the Rust model, we can achieve this by by marking all but one character of the mention anchor text with `ExtraCharacterSpan`s so that they are not counted.

## Context

- Part of matrix-org/matrix-rich-text-editor#698


alunturner (approved):
Looks good to me - I wonder if we should hold merge until the target branch is approved though?

PR #707, opened 2023-06-06
Draft: idea for selection decoupling

alunturner:
Opening as an idea for giving us a way to decouple selection logic from Rust and client.

Adds a `WebSelection` type that simulates what you'd get back from doing `window.getSelection()` in web.


alunturner:
Closing (but not deleting) in case this is useful to anybody in the future.

If work gets done to export an AST from the Rust model, I expect it would render this idea obsolete.

PR #710, opened 2023-06-08
Add onBeforeInput listener

alunturner:
To handle a special case in element web (regarding pasting stuff from an iPhone to a Mac, see issue https://github.com/vector-im/element-web/issues/25327)  we require a before input listener to be added to the rich text editor.

This PR adds that listener and ensures it passes the event through when required to handle the case listed in the issue above. This listener will be required anyway to implement IME behaviour correctly (ie allowing composition of characters in certain foreign languages)

### DIFF
--- a/platforms/web/lib/useListeners/useListeners.ts
+++ b/platforms/web/lib/useListeners/useListeners.ts
@@ -123,15 +123,24 @@ export function useListeners(
         const onInput = (e: Event) => isInputEvent(e) && _handleInput(e);
         editorNode.addEventListener('input', onInput);
 
-        const onPaste = (e: Event) => {
-            if (!isClipboardEvent(e)) {
-                return;
+        // Can be called by onPaste or onBeforeInput
+        const onPaste = (e: ClipboardEvent | InputEvent) => {
+            // this is required to handle edge case image pasting in Safari, see
+            // https://github.com/vector-im/element-web/issues/25327
+            const isSpecialCaseInputEvent =
+                isInputEvent(e) &&
+                e.inputType === 'insertFromPaste' &&
+                e.dataTransfer !== null;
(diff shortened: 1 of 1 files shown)


platforms/web/lib/useListeners/useListeners.ts
```diff
@@ -123,15 +123,24 @@ export function useListeners(
...
+                e.dataTransfer !== null;
+
+            const isEventToHandle =
+                isClipboardEvent(e) || isSpecialCaseInputEvent;
+
+            if (isEventToHandle) {
+                e.preventDefault();
+                e.stopPropagation();
+
+                _handleInput(e);
```
alunturner, line 142: Before we only used to call this from `onPaste` and returned early (ie did nothing) if it was not an event from the clipboard.

Now we need to handle a special case as a clipboard event, so the check has become "check if it is a clipboard event or the special case, and if it's either of those, do stuff. Otherwise, do nothing"

platforms/web/lib/useListeners/useListeners.ts
```diff
@@ -185,6 +199,7 @@ export function useListeners(
             editorNode.removeEventListener('paste', onPaste);
             editorNode.removeEventListener('wysiwygInput', onWysiwygInput);
             editorNode.removeEventListener('keydown', onKeyDown);
+            editorNode.removeEventListener('beforeinput', onBeforeInput);
```
alunturner, line 202: These lines create a reference to the `onPaste` function which will be called whenever a `beforeInput` event occurs. The `addEventListener` call attaches that function when the composer is created, the `removeEventListener` call removes it when the composer is removed from the page

aringenbach (approved):
LGTM

PR #711, opened 2023-06-08
[iOS] Mention node updates

aringenbach:
Fixes most of the iOS cases for mention node integration.

Things that are missing though: 
* Support for plain text mode transitions (should rework the `MentionReplacer` further to be able to re-create pills)
* Currently EI will NOT re-insert the `data-mention-type` if it is not present for a message edition (still have to think if we can afford Rust to do that with a bit of help from the hosting app)

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/DTCoreText/DTHTMLElement.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/DTCoreText/DTHTMLElement.swift
@@ -21,6 +21,16 @@ extension DTHTMLElement {
     func sanitize() {
         guard let childNodes = childNodes as? [DTHTMLElement] else { return }
 
+        if tag == .a,
+           attributes["data-mention-type"] != nil,
+           let textNode = self.childNodes.first as? DTTextHTMLElement {
+            let mentionTextNode = MentionTextNodeHTMLElement(from: textNode)
+            removeAllChildNodes()
+            addChildNode(mentionTextNode)
+            mentionTextNode.inheritAttributes(from: self)
+            mentionTextNode.interpretAttributes()
+        }
+
         if childNodes.count == 1, let child = childNodes.first as? DTTextHTMLElement {
             if child.text() == .nbsp {
                 // Removing NBSP character from e.g. <p>&nbsp;</p> since it is only used to
@@ -66,11 +76,14 @@ extension DTHTMLElement {
 private enum DTHTMLElementTag: String {
     case pre
     case code
+    case a
 }
 
 private extension DTHTMLElement {
     var tag: DTHTMLElementTag? {
-        DTHTMLElementTag(rawValue: name)
+        guard let name else { return nil }
+
+        return DTHTMLElementTag(rawValue: name)
     }
 
     func createDiscardableElement() -> PlaceholderTextHTMLElement {
--- a/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSMutableAttributedString.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSMutableAttributedString.swift
@@ -45,17 +45,17 @@ extension NSMutableAttributedString {
     /// a new attributed string part provided by the hosting app `HTMLPermalinkReplacer`.
     ///
     /// - Parameter permalinkReplacer: The permalink replacer providing new attributed strings.
-    func replaceLinks(with permalinkReplacer: HTMLPermalinkReplacer) {
-        enumerateTypedAttribute(.link) { (url: URL, range: NSRange, _) in
-            if let replacement = permalinkReplacer.replacementForLink(
-                url.absoluteString,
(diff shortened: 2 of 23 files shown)


platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/NSMutableAttributedString.swift
```diff
@@ -41,21 +41,21 @@ extension NSMutableAttributedString {
...
             ) {
-                let originalText = self.attributedSubstring(from: range).string
                 self.replaceCharacters(in: range, with: replacement)
-                self.addAttribute(.originalContent,
-                                  value: OriginalContent(text: originalText),
-                                  range: .init(location: range.location, length: replacement.length))
+                // TODO: find a way to avoid re-applying the attribute (make the replacement mutable ?)
+                self.addAttribute(.mention,
+                                  value: originalContent,
+                                  range: NSRange(location: range.location, length: replacement.length))
```
aringenbach, line 58: Fixing that would be great. The issue is that if the hosting app is giving us a part of an attributed string it would erase our mention attributes. And we can't really trust the hosting application to give us an attributed string part that saved our custom attributes. 

Maybe a solution would be to explicitly expect a NSTextAttachment to display the Pill, which covers the Element-iOS case but has less versatility than an NSAttributedString part.

platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Components/WysiwygComposerView/WysiwygComposerViewModelTests+Suggestions.swift
```diff
@@ -116,7 +116,7 @@ extension WysiwygComposerViewModelTests {
         XCTAssertEqual(
             viewModel.content.html,
             """
-            <a data-mention-type="room" contenteditable="false" href="https://matrix.to/#/#room1:matrix.org">Room 1</a> Text
```
aringenbach, line 119: Due note the space removal in those cases, maybe from a UX perspective we should add a space after the mention when we insert it without a suggestion 🤔 (which was the previous behaviour for this on the iOS impl)

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/DTCoreText/DTHTMLElement.swift
```diff
@@ -21,6 +21,16 @@ extension DTHTMLElement {
...
 
+        if tag == .a,
+           attributes["data-mention-type"] != nil,
+           let textNode = self.childNodes.first as? DTTextHTMLElement {
+            let mentionTextNode = MentionTextNodeHTMLElement(from: textNode)
+            removeAllChildNodes()
+            addChildNode(mentionTextNode)
+            mentionTextNode.inheritAttributes(from: self)
+            mentionTextNode.interpretAttributes()
+        }
```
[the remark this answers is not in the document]
alunturner: Having read a bit about it yesterday and today, I think the custom tag could be a really good way to go. I think it would be good to discuss what this solution would look like before starting work on it. For example would it be split out into `<mention-user>`, `<mention-room>` etc? I think to be valid custom html they have to be hyphenated, but I think this would be a change that helps neaten up some client side code.

platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Components/WysiwygComposerView/WysiwygComposerViewModelTests+Suggestions.swift
alunturner: I think that the check we do in Rust probably needs to be a bit better than it is. Currently it just looks for a next sibling and, if there isn't one, adds a space. I don't feel this is sufficient.

Perhaps it's easier to define when we don't add a space. To me it seems that we shouldn't add a space if there is a leaf following the mention we've just added _and_ that leaf doesn't start with a space. In all other cases add a space. Does that sound closer to what we want? It's quite hard trying to imagine all the cases of mention next to X element.

platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/DTCoreText/DTHTMLElement.swift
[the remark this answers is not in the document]
alunturner: I think it makes sense too as we're coding the type into the tag, which seems neater to me than having an attribute that needs parsing (and so also needs stripping)

platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Components/WysiwygComposerView/WysiwygComposerViewModelTests+Suggestions.swift
alunturner: Yep, got it the wrong way around, will edit

alunturner (commented):
Looks good to me, not sure if @langleyd also wants to take a look?

alunturner (approved):

Issue #714, opened 2023-06-09
Move from `<a>` tags to custom html tags for mentions


alunturner:
It is possible in html to use custom tags. To prevent us having to constantly parse attributes to check if something is a mention, we could move to using a custom tag. This has many benefits:

- It will tidy up how we identify what is a mention. In some places we check for the `contenteditable` attribute, others we check for `data-mention-type` often these are combined with checking the tag type too. All this repetition/confusion could be tidied up if we could simply always check the tag type in the html
- It would really help to tidy up Rust tests, which suffer from the same confusion, and make them easier to read. For example when reading a test in Rust, if I see a test with an `<a>` tag in it, is the test for links or is it for mentions? It's not easy to tell at a glance
- Discrimination between links and mentions becomes explicit in the html markup, not just in the inner workings of the Rust model

Proposed solution would be to move from `<a data-mention-type="X" ... />` to:

- `<mention-user contenteditable="false" style="{}" href="">user name</mention-user>`
- `<mention-room contenteditable="false" style="{}" href= "">room name</mention-room>`
- `<mention-at-room contenteditable="false" style="{}" href= "">text</mention-at-room>`

I think that this arrangement of information would satisfy all clients, as we have defined the type of mention (in the tag) then if it is editable (needed for web), custom styling (needed for web) and href (needed for all clients) in attributes. It's possible that another attribute may be required to make this legitimate html (I'm not completely familiar with custom components) but hopefully this serves as a reasonable start point.

PR #716, opened 2023-06-12
Amend mention message html output

alunturner:
This PR:

- makes it so that when you translate a mention to "message format" html then the following rules are applied:
    - if it's an at-room mention, only output the raw text `@room`
    - if it's a room or user mention, output the mention as `<a href="...">display_text</a>`

### DIFF
--- a/crates/wysiwyg/src/composer_model/mentions.rs
+++ b/crates/wysiwyg/src/composer_model/mentions.rs
@@ -77,7 +77,13 @@ where
         let (start, end) = self.safe_selection();
         let range = self.state.dom.find_range(start, end);
 
-        let new_node = DomNode::new_mention(url, text, attributes);
+        // use the display text to determine if it's an at-room or regular mention
+        let new_node = if text == "@room".into() {
+            DomNode::new_at_room_mention(attributes)
+        } else {
+            DomNode::new_mention(url, text, attributes)
+        };
+
         let new_cursor_index = start + new_node.text_len();
 
         let handle = self.state.dom.insert_node_at_cursor(&range, new_node);
--- a/crates/wysiwyg/src/dom/nodes/mention_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/mention_node.rs
@@ -129,12 +129,16 @@ impl<S: UnicodeString> MentionNode<S> {
         let cur_pos = formatter.len();
         match self.kind() {
             MentionNodeKind::MatrixUrl { display_text, url } => {
-                let mut attributes = self.attributes.clone();
+                let mut attributes = if as_message {
+                    vec![]
+                } else {
+                    self.attributes.clone()
+                };
+
                 attributes.push(("href".into(), url.clone()));
 
                 if !as_message {
                     attributes.push(("contenteditable".into(), "false".into()))
-                    // TODO: data-mention-type = "user" | "room"
                 }
 
                 self.fmt_tag_open(tag, formatter, &Some(attributes));
@@ -144,7 +148,18 @@ impl<S: UnicodeString> MentionNode<S> {
                 self.fmt_tag_close(tag, formatter);
             }
             MentionNodeKind::AtRoom => {
-                formatter.push(self.display_text());
+                if as_message {
+                    formatter.push(self.display_text())
+                } else {
+                    let mut attributes = self.attributes.clone();
+                    attributes.push(("href".into(), "#".into()));
+                    attributes.push(("contenteditable".into(), "false".into()));
+                    self.fmt_tag_open(tag, formatter, &Some(attributes));
+
+                    formatter.push("@room");
+
+                    self.fmt_tag_close(tag, formatter);
+                };
             }
         }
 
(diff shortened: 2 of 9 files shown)


crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -77,7 +77,16 @@ where
         let (start, end) = self.safe_selection();
         let range = self.state.dom.find_range(start, end);
 
-        let new_node = DomNode::new_mention(url, text, attributes);
+        // use the display text or the presence of a `data-mention-type => at-room` attribute to decide the mention type
+        let new_node = if text == "@room".into()
+            || attributes.iter().any(|(k, v)| {
+                k == &S::from("data-mention-type") && v == &S::from("at-room")
```
jonnyandrew, line 83: I believe it's an `@room` mention if and only if there is an occurrence of `@room` in the text making this condition redundant

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -77,7 +77,16 @@ where
         let (start, end) = self.safe_selection();
         let range = self.state.dom.find_range(start, end);
 
-        let new_node = DomNode::new_mention(url, text, attributes);
+        // use the display text or the presence of a `data-mention-type => at-room` attribute to decide the mention type
+        let new_node = if text == "@room".into()
```
jonnyandrew, line 81: Eventually this logic could be shared with the HTML parsing logic (matrix-org/matrix-rich-text-editor#695 / matrix-org/matrix-rich-text-editor#692). So we could add a note here to remind us to update when it's ready.

aringenbach (approved):
Looks great

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -77,7 +77,16 @@ where
...
+            || attributes.iter().any(|(k, v)| {
+                k == &S::from("data-mention-type") && v == &S::from("at-room")
+            }) {
+            DomNode::new_at_room_mention(attributes)
```
jonnyandrew, line 85: As this function is able to create an `@room` mention, we might want to reconsider the API, and either:
- make the `url` argument optional, or
- accept an enum (`AtRoom` | `Matrix(text, url)` - the actual types might come from matrix-org/matrix-rich-text-editor#695).

Alternatively, we could add a new function so that we have:
- `do_insert_mention(text, url) // with mandatory args`
- `do_insert_at_room_mention()`

crates/wysiwyg/src/dom/nodes/mention_node.rs
```diff
@@ -129,22 +129,34 @@ impl<S: UnicodeString> MentionNode<S> {
...
-                formatter.push(self.display_text());
+                // if formatting as a message, simply use the display text (@room)
+                if as_message {
+                    formatter.push(self.display_text())
+                } else {
+                    let mut attributes = self.attributes.clone();
+                    attributes.push(("href".into(), "#".into())); // designates a placeholder link in html
+                    attributes.push(("contenteditable".into(), "false".into()));
+
+                    self.fmt_tag_open(tag, formatter, &Some(attributes));
```
jonnyandrew, line 156: Should this tag be `<span>` (or `<mention>`), rather than `<a>`?

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -77,7 +77,16 @@ where
...
+        } else {
+            DomNode::new_mention(url, text, attributes)
```
jonnyandrew, line 87: When trying to insert an invalid mention, should the editor still do that? I'm wondering if it could return an error result and/or fall back to `DomNode::new_link()`.

It might not be possible to do without being able to parse mentions but could be worth adding a note

jonnyandrew (commented):
Looks good!

crates/wysiwyg/src/dom/nodes/mention_node.rs
alunturner: For the time being keeping everything as `<a>` - [there's a separate issue for moving to a custom html tag type here](https://github.com/matrix-org/matrix-rich-text-editor/issues/714) to try and keep these PRs small and digestible

crates/wysiwyg/src/composer_model/mentions.rs
alunturner: I think I'd be tempted to go a different way and have this function call `DomNode::new_mention` with url, display_text, attributes as we currently have  and then have `MentionNode::new` create either a room, user, or at-room mention as appropriate.

What do you think to that solution? 

If you're happy with me adding a note as the work may be further informed by the WIP mentions utils work (which is what I'm picking up next), I'll do that and address in that subsequent PR.

crates/wysiwyg/src/composer_model/mentions.rs
alunturner: That's a good question, and for now I believe we don't have a way of telling if it's valid or not. Will add a note to handle that (after parsing gives us that ability).

crates/wysiwyg/src/composer_model/mentions.rs
alunturner: Great - I wanted to think that was the case but wasn't sure. Updated in

PR #717, opened 2023-06-13
Mentions utils

alunturner:
This PR sets up the mention parsing crate and allows us to:

- parse mentions from a uri
- parse mentions from a uri and display text

It is designed with the expected usage of:

- checking mentions prior to inserting them into a composer (ie on a client calling to insert a mention)
- parsing mentions from messages prior to editing


aringenbach (approved):
LGTM

PR #719, opened 2023-06-13
Integrate mention parsing

alunturner:
This PR integrates the new parsing functions in the following ways:
- adds an `is_valid_uri` helper function to the `matrix_mentions` crate
- updates the `MentionNode` type quite a bit to accommodate the parsing changes
- updates the output as html/md of the `MentionNode` to address eg using mx_id inside room mentions and in markdown output
- enables us to detect when a link is a mention (for editing messages etc)
- updates tests where required
- enables the matrix mention parsing to handle non matrix permalinks

This will close the outstanding points on #679 , namely 
```
replace the inner text of a room mention link with the mxId (think this is what the spec dictates, should this be the same for users?)
Markdown output (to attach to body)

replace mention link representations with just the mxId, prefixed with @ or # as appropriate
```
Closes https://github.com/matrix-org/matrix-rich-text-editor/issues/679

### DIFF
--- a/crates/matrix_mentions/src/mention.rs
+++ b/crates/matrix_mentions/src/mention.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use ruma_common::{matrix_uri::MatrixId, MatrixToUri, MatrixUri};
+use ruma_common::{matrix_uri::MatrixId, MatrixToUri, MatrixUri, UserId};
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Mention {
@@ -147,10 +147,41 @@ impl Mention {
     }
 }
 
+/// Determines if a uri can be parsed for a matrix id. Attempts to treat the uri in three
+/// ways when parsing:
+/// 1 - As a matrix uri
+/// 2 - As a matrix to uri
+/// 3 - As a custom uri
+///  
+/// If any of the above succeed, return Some<MatrixIdI. Else return None.
 fn parse_matrix_id(uri: &str) -> Option<MatrixId> {
     if let Ok(matrix_uri) = MatrixUri::parse(uri) {
         Some(matrix_uri.id().to_owned())
     } else if let Ok(matrix_to_uri) = MatrixToUri::parse(uri) {
+        Some(matrix_to_uri.id().to_owned())
+    } else if let Some(matrix_id) = parse_external_id(uri) {
+        Some(matrix_id.to_owned())
+    } else {
+        None
+    }
+}
+
+/// Splits an external uri on `/#/` and then attempts to find the relevant matrix information
+/// from after the split
+///
+/// If successful return Some<MatrixId>. Else return None.
+fn parse_external_id(uri: &str) -> Option<MatrixId> {
+    // first split the string into the parts we need
+    let parts: Vec<&str> = uri.split("/#/").collect();
+    let before_hash = parts[0];
+    let after_hash = parts[1];
+
+    // now rebuild the uri and try again...
+    let hacked_uri = format!("{}{}", "https://matrix.to/#/", after_hash);
+
+    if let Ok(matrix_uri) = MatrixUri::parse(&*hacked_uri) {
+        Some(matrix_uri.id().to_owned())
+    } else if let Ok(matrix_to_uri) = MatrixToUri::parse(&*hacked_uri) {
         Some(matrix_to_uri.id().to_owned())
     } else {
         None
@@ -261,6 +292,18 @@ mod test {
     }
 
     #[test]
+    fn parse_uri_external_room() {
+        let uri =
+            "https://custom.custom.com/?secretstuff/#/!roomid:example.org";
+        let parsed = Mention::from_uri(uri).unwrap();
+
+        assert_eq!(parsed.uri(), uri);
+        assert_eq!(parsed.mx_id(), "!roomid:example.org");
+        assert_eq!(parsed.display_text(), "!roomid:example.org");
+        assert_eq!(parsed.kind(), &MentionKind::Room);
+    }
+
+    #[test]
     fn parse_link_user_text() {
         let uri = "https://matrix.to/#/@alice:example.org";
         let display_text = "Alice";
@@ -284,7 +327,7 @@ mod test {
 
         assert_eq!(parsed.uri(), uri);
         assert_eq!(parsed.mx_id(), "!room:example.org");
-        assert_eq!(parsed.display_text(), "!room:example.org"); // note the display_text is overridden
+        assert_eq!(parsed.display_text(), "!room:example.org");
         assert_eq!(parsed.kind(), &MentionKind::Room);
     }
 
--- a/crates/matrix_mentions/src/mention.rs
+++ b/crates/matrix_mentions/src/mention.rs
@@ -338,7 +338,7 @@ mod test {
 
         assert_eq!(parsed.uri(), uri);
         assert_eq!(parsed.mx_id(), "!room:example.org");
-        assert_eq!(parsed.display_text(), "!room:example.org");
+        assert_eq!(parsed.display_text(), "!room:example.org"); // note the display_text is overridden
         assert_eq!(parsed.kind(), &MentionKind::Room);
     }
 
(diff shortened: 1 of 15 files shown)


aringenbach (commented):
Looks good to me

What is our plan for supporting custom permalink URls ?

alunturner:
> Looks good to me
> 
> What is our plan for supporting custom permalink URls ?

After our discussion this morning, I've gone down the route of splitting the custom permalinks on "/#/" then treating them like we treat other links. Can be seen in the most recent commits.

crates/wysiwyg/src/dom/nodes/mention_node.rs
```diff
@@ -22,47 +23,73 @@ use crate::dom::to_tree::ToTree;
...
 
 #[derive(Clone, Debug, PartialEq, Eq)]
-pub enum MentionNodeKind<S>
-where
-    S: UnicodeString,
-{
-    MatrixUrl { display_text: S, url: S },
+pub enum MentionNodeKind {
+    Room { mention: Mention },
+    User { mention: Mention },
```
jonnyandrew, line 45: Could he type of mention Room / User be derived from the `mention` field? Otherwise we create the possibility of invalid states

E.g. `MentionNodeKind::Room(Mention { kind: MentionKind::User })`

I'd suggest keeping 

```
pub enum MentionNodeKind {
    MatrixUrl { mention: Mention },
    AtRoom,
}
```

Or perhaps better, integrating `AtRoom` into `matrix_mentions` crate.

crates/wysiwyg/src/dom/nodes/mention_node.rs
```diff
@@ -96,9 +123,19 @@ where
...
+
+    /// Util function to check if the display text is that of an at-room mention
+    pub fn is_at_room_display_text(text: &S) -> bool {
+        text == &S::from("@room")
+    }
+
+    /// Util function to get the display text for an at-room mention
+    pub fn get_at_room_display_text() -> S {
+        S::from("@room")
+    }
```
jonnyandrew, line 138: Could these utils and the `"@room"` constant be part of `matrix_mentions` crate?

crates/wysiwyg/src/dom/nodes/mention_node.rs
```diff
@@ -222,42 +281,25 @@ where
...
             S: UnicodeString,
         {
-            // TODO make this use mxId, for now we use display_text
-            buffer.push(this.display_text());
-            Ok(())
-        }
+            let text = match this.kind() {
+                // for User/Room type, we use the mx_id in the md output
+                MentionNodeKind::User { mention }
+                | MentionNodeKind::Room { mention } => S::from(mention.mx_id()),
```
jonnyandrew, line 298: For a user mention, the Matrix spec says we should use the potentially ambigious display name (if available) and only fall back to the mx ID when this is not available.

If we are concerned about being able to disambiguate the user given a Matrix message, this should still be possible by looking at the formatted HTML which contains the full mention URL.

jonnyandrew (commented):
Looks good so far!

crates/wysiwyg/src/dom/nodes/mention_node.rs
aringenbach: I think the long term plan to disambiguate at app level is https://github.com/matrix-org/matrix-spec-proposals/pull/3952 

Otherwise at the RTE level we will always parse the HTML so it will be fine, and as previously mentioned we need a custom Markdown to handle plain text mode with pills.

crates/wysiwyg/src/dom/nodes/mention_node.rs
aringenbach: ➕ on that last bit. 

When possible, things that apply the matrix mentions spec should be in that separate crate.

crates/wysiwyg/src/dom/nodes/mention_node.rs
alunturner: Done in [a2e3361](https://github.com/matrix-org/matrix-rich-text-editor/pull/719/commits/a2e336146bac9ba49d380c279f7813345508f66d)

crates/wysiwyg/src/dom/nodes/mention_node.rs
alunturner: Ok I have reverted it back to the MatrixURI style that it was before. I don't think that `AtRoom` should live in the mentions crate though - I'd argue it's more a node type as it contains none of the information that  MatrixURI mention would.

crates/matrix_mentions/src/mention.rs
```diff
@@ -142,16 +149,45 @@ impl Mention {
     }
 }
 
+/// Determines if a uri can be parsed for a matrix id. Attempts to treat the uri in three
+/// ways when parsing:
+/// 1 - As a matrix uri
+/// 2 - As a matrix to uri
+/// 3 - As a custom uri
```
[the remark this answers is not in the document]
aringenbach: I don't see any evidence that customisation of permalinks is part of the matrix spec, so yeah maybe it lies somewhere very specific. But I think it should still be easy to parameter,  not sure it lies in a non-default crate, as the platform code of the library is shipped with specific crate configuration. I think it could be a big drawback if people wanting to use custom permalink will have to maintain a fork of e.g. RTE's Swift Package.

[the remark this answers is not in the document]
alunturner: Addressed in [8417740](https://github.com/matrix-org/matrix-rich-text-editor/pull/719/commits/841774099585d1f22d6302d551a06b182c63303d)

crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -119,23 +112,21 @@ where
...
 
-    /// Creates a new mention node then inserts the node at the cursor position. If creation fails due to
-    /// an invalid uri, it will return `ComposerUpdate::keep()`.
-    /// It adds a trailing space when the inserted mention is the last node in it's parent.
+    /// Inserts the node at the cursor position. It adds a trailing space when the inserted
+    /// mention is the last node in it's parent.
     fn do_insert_mention(
         &mut self,
-        mention_node: DomNode<S>,
+        mention_node: MentionNode<S>, // TODO can we type this as a mention node somehow
```
jonnyandrew, line 119:
```suggestion
        mention_node: MentionNode<S>,
```

crates/wysiwyg/src/dom/nodes/mention_node.rs
```diff
@@ -41,8 +47,7 @@ where
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum MentionNodeKind {
-    Room { mention: Mention },
-    User { mention: Mention },
+    MatrixURI { mention: Mention },
```
jonnyandrew, line 50: nitpick: In UpperCamelCase, acronyms and contractions of compound words count as one word: use Uuid rather than UUID, Usize rather than USize or Stdin rather than StdIn ([source](https://rust-lang.github.io/api-guidelines/naming.html))
```suggestion
    MatrixUri { mention: Mention },
```

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -321,7 +321,7 @@ mod sys {
             );
 
             match creation_result {
-                Ok(node) => node,
+                Ok(node) => DomNode::Mention(node),
```
jonnyandrew, line 324: nitpick: We could implement `From` trait so that this becomes
```suggestion
                Ok(node) => node.into(),
```

jonnyandrew (approved):

crates/wysiwyg/src/dom/parser/parse.rs
alunturner: Will postpone this to a later bit of work, added a TODO in [46f48cb](https://github.com/matrix-org/matrix-rich-text-editor/pull/719/commits/46f48cb0f9e7c353e673b4e1179333f54bd8342c)

crates/wysiwyg/src/dom/nodes/mention_node.rs
alunturner: Addressed in [46f48cb](https://github.com/matrix-org/matrix-rich-text-editor/pull/719/commits/46f48cb0f9e7c353e673b4e1179333f54bd8342c)

crates/wysiwyg/src/composer_model/mentions.rs
alunturner: Addressed in [46f48cb](https://github.com/matrix-org/matrix-rich-text-editor/pull/719/commits/46f48cb0f9e7c353e673b4e1179333f54bd8342c)

PR #720, opened 2023-06-14
Alter attribute input for mentions

alunturner:
This PR:

- "unhooks" the attributes from the mobile side of things - they no longer need to pass in any attributes as per this issue https://github.com/matrix-org/matrix-rich-text-editor/issues/721, but still require removal from the mobile side of the codebase
- makes the addition of `data-mention-type` automatic
- updates the tests to reflect this new behaviour
- stops web passing in the unneeded `data-mention-type` attribute in web, and changes the type for attributes to a map
- associated event/type changes in web to allow us to get the formatted message output and also to handle at room mentions using the new exposed api function

### DIFF
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -282,7 +282,7 @@ impl ComposerModel {
         self: &Arc<Self>,
         url: String,
         text: String,
-        attributes: Vec<Attribute>, // TODO remove attributes
+        _attributes: Vec<Attribute>, // TODO remove attributes
     ) -> Arc<ComposerUpdate> {
         let url = Utf16String::from_str(&url);
         let text = Utf16String::from_str(&text);
@@ -324,7 +324,7 @@ impl ComposerModel {
         url: String,
         text: String,
         suggestion: SuggestionPattern,
-        attributes: Vec<Attribute>, // TODO remove attributes
+        _attributes: Vec<Attribute>, // TODO remove attributes
     ) -> Arc<ComposerUpdate> {
         let url = Utf16String::from_str(&url);
         let text = Utf16String::from_str(&text);
--- a/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
+++ b/bindings/wysiwyg-ffi/src/ffi_composer_model.rs
@@ -282,19 +282,11 @@ impl ComposerModel {
         self: &Arc<Self>,
         url: String,
         text: String,
-        attributes: Vec<Attribute>,
+        attributes: Vec<Attribute>, // TODO remove attributes
     ) -> Arc<ComposerUpdate> {
         let url = Utf16String::from_str(&url);
         let text = Utf16String::from_str(&text);
-        let attrs = attributes
-            .iter()
-            .map(|attr| {
-                (
-                    Utf16String::from_str(&attr.key),
-                    Utf16String::from_str(&attr.value),
-                )
-            })
-            .collect();
+        let attrs = vec![];
         Arc::new(ComposerUpdate::from(
             self.inner.lock().unwrap().insert_mention(url, text, attrs),
         ))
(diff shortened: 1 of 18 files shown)


aringenbach (commented):
Looks good to me apart from the failing tests. 

What is our plan for supporting custom permalink URls ?

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -1081,27 +1081,10 @@ mod js {
...
-        #[wasm_bindgen_test]
-        fn a_with_bad_attribute() {
-            let html = r#"<a invalidattribute="true" href="http://example.com">a user mention</a>"#;
-            let dom = HtmlParser::default().parse::<Utf16String>(html).unwrap();
-            assert_eq!(
-                dom.to_string(),
-                r#"<a href="http://example.com">a user mention</a>"#
-            );
-        }
-
```
alunturner, line 1100: No longer good tests, as we don't have links with attributes

platforms/web/lib/composer.ts
```diff
@@ -71,27 +72,38 @@ export function processInput(
...
+                return action(
+                    composerModel.insert_at_room_mention_at_suggestion(
+                        suggestion,
+                        attributes,
+                    ),
+                    'insert_at_room_mention_at_suggestion',
+                );
+            }
+            break;
+        }
```
alunturner, line 92: This replaces the `if text === ...` check from before to completely separate mention and at mention handling

platforms/web/lib/useListeners/types.ts
```diff
@@ -28,15 +26,17 @@ export type LinkEvent = Omit<InputEvent, 'data'> & {
...
-    | keyof HTMLAttributes<HTMLAnchorElement>
-    | `data-${string}`;
-
-export type Attributes = {
-    [K in AnchorElementAttributes]?: string;
-};
+export type AllowedMentionAttributes = Map<
+    'style' | 'data-mention-type',
+    string
+>;
```
alunturner, line 32: This change the type to be a `Map`, which fixes another open issue and we also make it less permissive in terms of what it accepts.

platforms/web/lib/useWysiwyg.ts
```diff
@@ -108,5 +108,6 @@ export function useWysiwyg(wysiwygProps?: WysiwygProps) {
             traceAction: testUtilities.traceAction,
         },
         suggestion: memoisedMappedSuggestion,
+        messageContent: composerModel?.get_content_as_message_html() ?? null,
```
alunturner, line 111: Exposing the ability to use the rust model to output message format html

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -282,19 +271,11 @@ impl ComposerModel {
...
-        let attrs = attributes
-            .iter()
-            .map(|attr| {
-                (
-                    Utf16String::from_str(&attr.key),
-                    Utf16String::from_str(&attr.value),
-                )
-            })
-            .collect();
+        let attrs = vec![];
```
jonnyandrew, line 278: While this is a TODO to be removed, I'd prefer keep passing the `attributes` argument through to the composer. That way it still works as expected until it's removed.

Also, should we link an issue?

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
```diff
@@ -332,20 +304,12 @@ impl ComposerModel {
         url: String,
         text: String,
         suggestion: SuggestionPattern,
-        attributes: Vec<Attribute>,
+        _attributes: Vec<Attribute>, // TODO remove attributes
```
jonnyandrew, line 307: Same as above

crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -831,8 +831,8 @@ mod js {
                         self.current_path.push(DomNodeKind::Link);
 
                         let mut attributes = vec![];
-                        let valid_attributes =
-                            ["contenteditable", "data-mention-type", "style"];
```
jonnyandrew, line 835: :tada:

jonnyandrew (approved):
I'm happy to see this change happen!

bindings/wysiwyg-ffi/src/ffi_composer_model.rs
alunturner: Yep I think that's a fair call. 
Reverted in: [2ed5d77](https://github.com/matrix-org/matrix-rich-text-editor/pull/720/commits/2ed5d77d54c9d9d65a6eef00e31c86e5f887aecf)
Issue: https://github.com/matrix-org/matrix-rich-text-editor/issues/721

PR #723, opened 2023-06-20
Ime composing

alunturner:
Closes https://github.com/vector-im/element-web/issues/23649
Closes https://github.com/vector-im/element-web/issues/25247

...or at least aims to. I have managed to add a test that simulates what a browser does, as it doesn't seem to be doable using `userEvent` from my previous attemtps.

Adds the ability for the composer to handle IME type composition, for letters with accents/CJK etc.

New behaviour:

https://github.com/matrix-org/matrix-rich-text-editor/assets/56027671/7c7d07fc-cf88-459d-89b1-d926142b9bfc


platforms/web/src/App.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+import App from './App';
+
+test('User can enter an umlaut using IME type input', async () => {
+    const user = userEvent.setup();
+
+    render(<App />);
+    const textbox = screen.getByRole('textbox');
+    await user.type(textbox, '{Shift>}^{/Shift}a');
+    screen.debug();
```
alunturner, line 28: This will be implemented/expanded if I can figure out how to get it working in the tests

platforms/web/lib/useListeners/useListeners.ts
```diff
@@ -188,9 +191,25 @@ export function useListeners(
...
         // https://github.com/vector-im/element-web/issues/25327
-        const onBeforeInput = onPaste;
+        const onBeforeInput = (e: ClipboardEvent | InputEvent) => {
+            if (isInputEvent(e) && e.isComposing) return;
+            return onPaste(e);
+        };
         editorNode.addEventListener('beforeinput', onBeforeInput);
 
+        const onCompositionEnd = (e: CompositionEvent): void => {
+            console.log(e);
```
jonnyandrew, line 201: Should this log be removed?

platforms/web/lib/useListeners/useListeners.ts
```diff
@@ -188,9 +191,25 @@ export function useListeners(
...
+            // create a new inputEvent for us to process
+            const inputEvent = new InputEvent('input', {
+                data: e.data,
+                inputType: 'insertCompositionText',
+            });
+
+            // now process that new event
+            onInput(inputEvent);
```
jonnyandrew, line 209: I had a look at the [`compositionend` event](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event) and it sounds like this is also fired when the composition is cancelled.

If that happens, will we need to do something different?

jonnyandrew (commented):
Looks good - just one question about the `compositionend` event handler

platforms/web/lib/useListeners/useListeners.ts
alunturner: Yes it should

platforms/web/lib/useListeners/useListeners.ts
alunturner: This is a good question, I think this is the correct behaviour.

For instance, on a mac, if you press `option + i` to insert a "composition style" ^ (ie the next letter will sit beneath the hat), then click off the textbox we _still_ process it as an input event, so that the `^` is being shown in the textbox.

To me that seems to be the correct behaviour for cancellation, but I'm not super familiar with this input method.

jonnyandrew (approved):

PR #731, opened 2023-06-21
Use correct message output format in conversion function

alunturner:
I missed that this function would also require switching over to the new message format output function. Have updated the function to allow us to choose if we want the output from this call in message format or not, then added tests.

This is required due to where we call this from in the client, as we call it when converting between composers and also when sending a message.

### DIFF
--- a/platforms/web/lib/conversion.test.ts
+++ b/platforms/web/lib/conversion.test.ts
@@ -99,46 +99,60 @@ describe('Plain text <=> markdown', () => {
     });
 });
 
-describe('Mentions conversion from plain text to rich', () => {
-    it('converts at-room mentions as expected', async () => {
+describe('Mentions', () => {
+    it('converts at-room mentions for composer as expected', async () => {
         const input = '@room';
         const asComposerHtml = await plainToRich(input);
-        const asMessageHtml = await plainToRich(input, true);
 
-        // for this case, we'd expect the composer html to be an <a> tag, whereas
-        // the message html should be the equivalent plain text
         expect(asComposerHtml).toBe(
             '<a data-mention-type="at-room" href="#" contenteditable="false">@room</a>',
         );
+    });
+
+    it('converts at-room mentions for message as expected', async () => {
+        const input = '@room';
+        const asMessageHtml = await plainToRich(input, true);
+
         expect(asMessageHtml).toBe('@room');
     });
 
-    it('converts user mentions as expected', async () => {
+    it('converts user mentions for composer as expected', async () => {
         const input =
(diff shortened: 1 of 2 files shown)


platforms/web/lib/conversion.test.ts
```diff
@@ -98,3 +98,49 @@ describe('Plain text <=> markdown', () => {
...
+        const asComposerHtml = await plainToRich(input);
+        const asMessageHtml = await plainToRich(input, true);
+
+        // for this case, we'd expect the composer html to be an <a> tag, whereas
+        // the message html should be the equivalent plain text
+        expect(asComposerHtml).toBe(
+            '<a data-mention-type="at-room" href="#" contenteditable="false">@room</a>',
+        );
+        expect(asMessageHtml).toBe('@room');
+    });
```
jonnyandrew, line 114: nitpick:

In these cases, it feels like there are two things being tested per test. I'd be tempted to split this into two e.g.

```typescript
    const atRoomInput = '@room';
    
    it('converts at-room mentions to message html', async () => {
        const asMessageHtml = await plainToRich(atRoomInput, true);

        expect(asMessageHtml).toBe('@room');
    });
    
    it('converts at-room mentions to composer html', async () => {
        const asComposerHtml = await plainToRich(atRoomInput);

        // for this case, we'd expect the composer html to be an <a> tag, whereas
        // the message html should be the equivalent plain text
        expect(asComposerHtml).toBe(
            '<a data-mention-type="at-room" href="#" contenteditable="false">@room</a>',
        );
    });
```

This way, when something breaks, it's easier to see exactly which input causes the failure.

platforms/web/lib/conversion.ts
```diff
@@ -67,7 +67,7 @@ export async function richToPlain(richText: string) {
     return plainText;
 }
 
-export async function plainToRich(plainText: string) {
+export async function plainToRich(plainText: string, inMessageFormat = false) {
```
jonnyandrew, line 70: nitpick: Would it be nicer to have the message format be the default so that the composer internals are only exposed if explicitly requested?

jonnyandrew (approved):
Looks good to me!

platforms/web/lib/conversion.test.ts
alunturner: I think that's a fair point, will separate them out

platforms/web/lib/conversion.ts
alunturner: I actually think that it's better to remove the default completely. I added it to avoid having to change all the function calls, but it opens the option of accidentally misusing the function. Will amend that in the next commit.

PR #733, opened 2023-06-22
Support initialising composer with mentions

alunturner:
In order to be able to initialise the composer with mentions properly, we can't simply slap the html from the message being edited into the composer (as the rust model needs to deduce and add some attributes to it first, like `data-mention-type` and `contenteditable`).

This PR:

- stops us using the `initialContent` to set any state directly (as this wouldn't work correctly for mentions)
- creates a model from `initialContent` then reads from the model, using the model output to set the composer content
- adds tests for parsing mentions
- adds a test file for the whole `useComposer` hook

### DIFF
--- a/platforms/web/lib/useComposerModel.ts
+++ b/platforms/web/lib/useComposerModel.ts
@@ -85,6 +85,7 @@ export function useComposerModel(
                         );
                     }
                 } catch (e) {
+                    // if the initialisation fails, due to a parsing failure of the html, fallback to an empty composer
                     setComposerModel(new_composer_model());
                 }
             } else {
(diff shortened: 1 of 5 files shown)


platforms/web/lib/useWysiwyg.ts
```diff
@@ -71,11 +71,6 @@ export function useWysiwyg(wysiwygProps?: WysiwygProps) {
 
     const formattingFunctions = useFormattingFunctions(ref, composerModel);
 
-    const onError = useCallback(
-        (content?: string) => initModel(content),
-        [initModel],
-    );
-
```
alunturner, line 78: This was needless wrapping of initModel in another `useCallback`. This has effectively been moved inside the `useComposerModel` hook

platforms/web/lib/useComposerModel.ts
```diff
@@ -68,23 +68,26 @@ export function useComposerModel(
...
                             editorRef.current,
-                            initialContent,
+                            modelContent,
                             0,
-                            initialContent.length,
+                            modelContent.length,
                         );
                     }
                 } catch (e) {
+                    // if the initialisation fails, due to a parsing failure of the html, fallback to an empty composer
```
aringenbach, line 90: maybe we can try to display a raw version of the text in that case ? (just an idea for later)

aringenbach (approved):
LGTM

platforms/web/lib/useComposerModel.ts
alunturner: Yes that is a good idea. The failure here will be if we have poorly formed HTML though, and I don't know of a simple way to translate poorly formed HTML into raw text. I imagine it would be hard because of the number of ways the HTML could be incorrect.

PR #734, opened 2023-06-23
Linebreaks for blockquotes

alunturner:
Trying to address a P2 I noticed we had an issue with linebreaks around blockquotes in general when parsing from markdown.

Previous behaviour when parsing:

https://github.com/matrix-org/matrix-rich-text-editor/assets/56027671/83f33149-ff58-405d-b69c-660fa9014b2b

New behaviour (also dependent on other PRs, but this is what we're aiming for):

https://github.com/matrix-org/matrix-rich-text-editor/assets/56027671/8649b32b-7d61-405d-85cc-6f4fa1410850

This is achieved by:
- adding some more replacements in the markdown parsing step
- adding a test

### DIFF
--- a/crates/wysiwyg/src/dom/parser/markdown/markdown_html_parser.rs
+++ b/crates/wysiwyg/src/dom/parser/markdown/markdown_html_parser.rs
@@ -50,8 +50,7 @@ impl MarkdownHTMLParser {
             }
         };
 
-        // We need to remove any trailing newline characters from block tags, then
-        // finally replace any remaining newline characters with linebreaks
+        // Remove any trailing newline characters from block tags
         let html = html
             .replace("<ul>\n", "<ul>")
             .replace("</ul>\n", "</ul>")
--- a/crates/wysiwyg/src/dom/parser/markdown/markdown_html_parser.rs
+++ b/crates/wysiwyg/src/dom/parser/markdown/markdown_html_parser.rs
@@ -62,8 +62,7 @@ impl MarkdownHTMLParser {
             .replace("<blockquote>\n", "<blockquote>")
             .replace("</blockquote>\n", "</blockquote>")
             .replace("<p>\n", "<p>")
-            .replace("</p>\n", "</p>")
-            .replace('\n', "<br />");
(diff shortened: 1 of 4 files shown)


crates/wysiwyg/src/tests/test_set_content.rs
```diff
@@ -74,3 +74,19 @@ fn set_contents_with_line_break_in_code_block() {
...
+    assert_eq!(tx(&model), "<blockquote><p>quote|</p></blockquote>");
+}
+
+#[test]
+fn set_content_from_markdown_with_newlines() {
+    let mut model = cm("|");
+    model
+        .set_content_from_markdown(&utf16("hello\nthere"))
+        .unwrap();
+    assert_eq!(tx(&model), "hello<br />there|");
```
aringenbach, line 91: Shouldn't this be `<p>hello</p><p>there</p>` ? AFAIK `br` is not well supported yet for all operations of the RTE, e.g. lists, code blocks, etc

alunturner: Yes I think it should. I'll take another look on Monday because I've been working on the element web side of this which is also related.

alunturner: I have removed this check because fixing it is not a small job. I'm going to attack that later this week in a larger piece of markdown conversion work if I get the time

jonnyandrew (approved):

crates/wysiwyg/src/dom/parser/markdown/markdown_html_parser.rs
```diff
@@ -50,14 +50,19 @@ impl MarkdownHTMLParser {
             }
         };
 
-        // Format new lines.
+        // We need to remove any trailing newline characters from block tags, then
+        // finally replace any remaining newline characters with linebreaks
```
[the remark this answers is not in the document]
alunturner: Sorry - this comment is out of date. I removed the `finally replacing...with linebreaks` behaviour. Will update now.

aringenbach (approved):
LGTM

PR #735, opened 2023-06-23
Amend web conversion functions

alunturner:
Dependent on the blockquote rust fix. In draft until work is carried out to update the state management for the rich text editor in element web (required to be able to use this properly).

This supports work in element web to enable us to better parse the output from the plain text composer by reading from `.innerText` instead of from `.innerHTML`. Required to address P2 for blockquote markdown not working in element web rte in plain mode.

This work removes all the mentions tests as previously, mentions would persist when moving from plain to rich, but not from rich to plain modes. The upcoming PR on element web means that mentions will now become plain text when moving between modes in either direction.

### DIFF
--- a/platforms/web/lib/conversion.test.ts
+++ b/platforms/web/lib/conversion.test.ts
@@ -14,12 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import {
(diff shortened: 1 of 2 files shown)


alunturner:
Closing in favour of another branch

PR #736, opened 2023-06-27
Update web text parsing

alunturner:
Allows us to parse "markdown" from a plain text composer into a rust model correctly. This needs to work because when the user uses the plain text composer, the a rust model is built to convert their input into the `formatted_body` html type output that is appended to the matrix message.

### DIFF
--- a/platforms/web/lib/conversion.ts
+++ b/platforms/web/lib/conversion.ts
@@ -100,22 +100,22 @@ export function amendInnerHtmlButBetter(composer: HTMLDivElement): string {
     while (node !== null) {
         const isTopLevelTextNode =
             node.nodeName === '#text' &&
-            node.parentElement.isSameNode(composer);
+            node.parentElement?.isSameNode(composer);
         const isNestedTextNode =
             node.nodeName === '#text' &&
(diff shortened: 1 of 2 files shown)


platforms/web/lib/conversion.ts
```diff
@@ -88,3 +77,106 @@ export async function plainToRich(plainText: string, inMessageFormat: boolean) {
...
+            // content is the html of the mention i.e. <a ...attributes>text</a>
+            let content = node.firstChild?.parentElement?.outerHTML ?? '';
+            if (shouldAddNewlineCharacter(node)) {
+                content += NEWLINE_CHAR;
+            }
+            markdownOutput += content;
+        }
+
+        node = iterator.nextNode();
+    }
```
aringenbach, line 154: A bit scared about this code, if we step upon nodes that are not expected here, they will just get ignored IIUC. 
Does it make sense to, at least, log nodes getting ignored ? (at least on debug if that affects performance)

aringenbach: Seems fine with the logging. 
I guess we'll see if that ever triggers.

aringenbach (approved):
LGTM

PR #742, opened 2023-07-19
Revamp xcframework generation to avoid conflicts with other Rust libraries

aringenbach:
Hopefully the move to a documented shell script makes all the whacky things we have to do for the iOS generation a bit clearer.


langleyd (approved):
Nice, LGTM

PR #745, opened 2023-07-21
Update uniffi version to 0.24.1

aringenbach:
Also adds a set of tools to help keep the same version everywhere. 

Also rename the ffi crate back to its previous non-standard name to avoid some issues with the Gradle plugin, see https://github.com/matrix-org/matrix-rich-text-editor/blob/main/platforms/android/README.md

Link to Uniffi-rs issue with multiple Rust libraries: https://github.com/mozilla/uniffi-rs/issues/1670

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,8 @@ resolver = "2"
 rust-version = "1.71"
 
 [workspace.dependencies]
+# We use a custom version of Uniffi that renames symbols that might clash
+# with other Rust libraries. See https://github.com/mozilla/uniffi-rs/issues/1670
 uniffi = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
 uniffi_macros = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -4,12 +4,22 @@ members = [
...
+    "crates/wysiwyg",
+    "crates/matrix_mentions",
 ]
 resolver = "2"
 
 [workspace.package]
 rust-version = "1.71"
 
+[workspace.dependencies]
+uniffi = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
```
jonnyandrew, line 19: Could you add a comment linking to the issue in UniFFI to help explain why this is needed?

bindings/wysiwyg-ffi/README.md
```diff
@@ -7,18 +7,7 @@ Rust code that can be used to generate Kotlin and Swift bindings for wysiwyg-rus
 ### 1. Install Rust
 See the [installation guide](https://www.rust-lang.org/tools/install).
 
-### 2. Install `uniffi-bindgen`
```
jonnyandrew, line 10: Thanks, I'm glad that we can now remove this step!

jonnyandrew (approved):
Looks good to me, I'd just like to see a comment about using the UniFFI fork

Cargo.toml
aringenbach: Added to description

jonnyandrew (approved):
Thanks!

PR #748, opened 2023-07-25
Set WysiwygComposerView as MainActor

aringenbach:
Might fix/silence some SwiftUI warnings about Publishing properties.
Issue seems to be with the `focused` property of the model but it's called from this view's context:

```swift
        public func textViewDidBeginEditing(_ textView: UITextView) {
            focused.wrappedValue = true
        }
        
        public func textViewDidEndEditing(_ textView: UITextView) {
            focused.wrappedValue = false
        }
```


aringenbach:
Oh well, looks like UI tests are unhappy with this change, I'll put this on draft for now.

aringenbach:
Issue should be fixed with https://github.com/matrix-org/matrix-rich-text-editor/pull/761

Issue #315, opened 2022-11-16
[Android] Expose a Composable rich text editor


bmarty:
To integrate the editor to an Android application written using Jetpack Compose, the Android library should expose a Composable editor, i.e. a method annotated with `@Composable`.

Will be useful for https://github.com/vector-im/element-x-android-poc/issues/1, however this is not blocking us.

### Tasks
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/754
  - [x] Theming and dark mode
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/755
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/758
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/760
  - [x] Add usage documentation
- [x] Fix: https://github.com/matrix-org/matrix-rich-text-editor/pull/764
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/766
  - [x] Focus API
  - [x] Line count API
  - [x] Text and cursor style settings
- [x] https://github.com/matrix-org/matrix-rich-text-editor/pull/767
- [x] Release

### Out of scope
- Feature parity with View library

PR #754, opened 2023-08-01
[Android] Add skeleton for Compose support

jonnyandrew:
## Changes

- Add a new library project, `:library-compose`, that exposes the editor as a Composable
- Add a new example project, `:example-compose`, that shows how to integrate the new UI
- Migrate to Gradle version catalogs now that we have 4 modules in the project
- Add some basic functionality (bold, italic, setHtml)
- Add an integration test

### Breaking changes
- `EditorEditText` is no longer a descendant of `TextInputEditText` as this was preventing use in non-AppCompat themes

### Screenshots
| Light | Dark |
|-|-|
| Screenshot from 2023-08-01 16-54-22 | Screenshot from 2023-08-01 16-52-14 |

### Future work

- Bring Compose library to parity with the View based library
- Publish the library for apps to consume

## Context

- Part of #315


platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -0,0 +1,95 @@
...
+    onActionsChanged: (actions: Map<ComposerAction, ActionState>) -> Unit = { },
+    onSelectionChanged: (start: Int, end: Int) -> Unit = { _, _ -> },
+    onMenuActionChanged: (menuAction: MenuAction) -> Unit = { },
+) {
+    val scope = rememberCoroutineScope()
+
+    var textWatcher by remember { mutableStateOf<((Editable?) -> Unit)>({ }) }
+
+    AndroidView(
+        modifier = modifier.fillMaxWidth(),
```
julioromano, line 51: Since this is a library we shouldn't apply any modifier here (unless it's strictly needed) and leave full control of this view's dimensions to the caller.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -0,0 +1,95 @@
...
+ * @param onHtmlChanged Called whenever the editor content changes with the new HTML
+ * @param onMarkdownChanged Called whenever the editor content changes with the new Markdown
+ * @param onActionsChanged Called whenever the editor action states change
+ * @param onSelectionChanged Called whenever the editor selection changes
+ * @param onMenuActionChanged Called whenever the editor menu action changes
+ */
+@Composable
+fun RichTextEditor(
+    modifier: Modifier = Modifier,
+    actions: Flow<EditorAction> = emptyFlow(),
```
julioromano, line 39: This looks odd.
This kind of internal state of the RTE should be modeled with the "state holder" pattern : https://developer.android.com/jetpack/compose/state#managing-state

platforms/android/settings.gradle
```diff
@@ -13,4 +13,6 @@ dependencyResolutionManagement {
     }
 }
 rootProject.name = "Rich Text Editor"
-include ':example', ':library'
+include ':example-view', ':example-compose', ':library', ':library-compose'
+
+enableFeaturePreview("VERSION_CATALOGS")
```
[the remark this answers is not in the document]
jonnyandrew: Agreed and is the plan - I didn't want to make this PR too big by mixing the Gradle upgrade into this PR

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jonnyandrew: Yeah it doesn't seem to be idiomatic Compose - the reason I added this flow is to work with the underlying non-Compose View. But I'd definitely prefer not to force this pattern on users of the compose library.

I thought about exposing some actions by following the pattern of `value` +`onValueChanged` i.e.
```kotlin
@Composable
fun MyComposable(
    value: String,
    onValueChange: (String) -> Unit,
)
```

But I was having difficulty propogating this down to the underlying View class without triggering all the different setters every time the update block is run.

I was also unsure about how best to extract one-shot actions such as `undo()`/`redo()` into a state.

Perhaps you know a way I can get around these issues?

julioromano: See code snippets:

https://developer.android.com/topic/architecture/ui-layer/stateholders#ui-logic

AND

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/SheetDefaults.kt;l=97?q=SheetState&sq=&ss=androidx

These show how to implement a state holder class.

Ideally that in your case that class will hold state regarding whether or not "bold" is enabled, and also offers methods to change those settings.

Then there will be a `rememberRichTextEditor()` or similar function that remembers an instance of that class. It will be used as a default value for the `RichTextEditor` composable so that, if needed, can be hoisted out of it and used by other code.

julioromano: If you can't your get around it let's book a meeting and pair a bit on it.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jonnyandrew: https://github.com/matrix-org/matrix-rich-text-editor/pull/754/commits/7b67bdd146c2c6ea8ca498c5950e5069d4fd7c93

jmartinesp:
Hey, I just wanted to add some info here, since there could be a possibility of creating a 100% Compose library using `PlatformTextInputAdapter` like this, wrapping our existing `InputConnection`: https://android-review.googlesource.com/c/platform/frameworks/support/+/2406439/38/compose/foundation/foundation/integration-tests/foundation-demos/src/main/java/androidx/compose/foundation/demos/text/PlatformTextInputAdapterDemo.kt

I think this is what we should implement in the long term but if we're constrained by time, we can start with this view-based implementation and iterate on it once the API is more mature.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -23,9 +27,32 @@ fun RichTextEditor(
     modifier: Modifier = Modifier,
 ) {
     val context = LocalContext.current
+    val lifecycleOwner = LocalLifecycleOwner.current
+
+    // Clean up the connection between view and state holder
+    DisposableEffect(lifecycleOwner) {
```
julioromano, line 33: This can be done without lifecycle.
```
DisposableEffect(Unit) {
onDispose {
state.viewConnection = null
}
}
```
This will set `viewConnection` as null whenever `RichTextEditor` leaves the composition.

julioromano (approved):
LGTM!

jonnyandrew:
@jmartinesp great point but yes the reason for wrapping the legacy View library is indeed time constraints; it allows us to maintain both view and compose libraries in a single place.

But as this library and compose both mature, I suspect there come a time where view support is less important and 100% compose is able to give better flexibility in terms of layout and style than we currently have with `TextView` and `Span`. At that point it'd be great to move in that direction.

PR #755, opened 2023-08-08
[Android] Add formatting actions to compose library

jonnyandrew:
## Changes

Add formatting actions to the Android compose library by exposing new methods on `RichTextEditorState`:
- Inline formatting (bold, italic, underline, strike-through, code)
- Lists
- Code blocks
- Quotes

Also
- Expose undo / redo actions
- Add everything to the example app

### Screenshot

image

## Context

### DIFF
--- a/platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
+++ b/platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
@@ -12,10 +12,8 @@ import androidx.compose.material3.Surface
 import androidx.compose.ui.Modifier
 import io.element.android.wysiwyg.compose.RichTextEditor
 import io.element.android.wysiwyg.compose.rememberRichTextEditorState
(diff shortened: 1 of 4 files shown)


platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
```diff
@@ -38,13 +40,25 @@ class MainActivity : ComponentActivity() {
...
-                            onBoldClick = {
-                                state.toggleBold()
-                            },
-                            onItalicClick = {
-                                state.toggleItalic()
-                            },
                             actionStates = state.actions,
+                            onActionClick = { action ->
+                                when(action) {
+                                    ComposerAction.BOLD -> state.toggleInlineFormat(InlineFormat.Bold)
```
julioromano, line 46: This looks like boilerplate. If users of the library must always handle all actions, shouldn't we put this logic in the state holder and just expose `state.handleAction()` or something?

julioromano (approved):
Proactive LGTM With some comment.

PR #756, opened 2023-08-09
[Android] Fix coverage for compose library

jonnyandrew:
- Fix coverage not running for compose library
- Fix coverage report generation for unit tests and instrumented tests


julioromano (approved):
Nothing against jacoco, though on EAX we use https://github.com/Kotlin/kotlinx-kover
I leave to you the choice.

PR #757, opened 2023-08-09
[Android] Update dependencies

jonnyandrew:
## Changes
This PR can be reviewed commit by commit.
- Update all dependencies
- Update Gradle and Android Gradle Plugin

## Context

Resolves https://github.com/matrix-org/matrix-rich-text-editor/pull/754#discussion_r1282732490


platforms/android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Wed Jul 13 15:15:46 CEST 2022
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.0-bin.zip
```
julioromano, line 3: You can safely bump to the latest (8.2.1)

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditorState.kt
```diff
@@ -30,21 +32,21 @@ class RichTextEditorState internal constructor() {
      *
      * @param inlineFormat which format to toggle (e.g. [InlineFormat.Bold])
      */
-    fun toggleInlineFormat(inlineFormat: InlineFormat) {
+    suspend fun toggleInlineFormat(inlineFormat: InlineFormat) = withContext(Dispatchers.Main) {
```
julioromano, line 35: Why this change? Is it related to the dependency updates?

jonnyandrew: This is related to an issue with tests since updating compose. The UI tests are failing with:

```
android.view.ViewRootImpl$CalledFromWrongThreadException: Only the original thread that created a view hierarchy can touch its views.
```

This change makes the problem go away but I'm not sure it is the right solution.

jonnyandrew: If `state.toggleInlineFormat()` is not called from the main thread in a non-test environment, then this protects the caller from causing a crash.

But that's not possible or not something we care about, we can just ensure tests call these functions on the main thread.

julioromano: Indeed. It is very much preferable that our public API is not dictated by test necessities. Better change the tests.

platforms/android/gradle.properties
```diff
@@ -21,6 +21,8 @@ kotlin.code.style=official
 # resources declared in the library itself and none from the library's dependencies,
 # thereby reducing the size of the R class for that library
 android.nonTransitiveRClass=true
```
julioromano, line 23: This can be remove. it's true by default since agp 8

platforms/android/gradle.properties
```diff
@@ -21,6 +21,8 @@ kotlin.code.style=official
...
+android.defaults.buildfeatures.buildconfig=true
```
julioromano, line 24: This can be removed and replaced by `android.enableBuildConfigAsBytecode=true` to speed up a bit compilation.
you can then enable buildconfig generation on a per module basis only for the modules that actually make us of it.

You can enable it in the gradle config file of the module with:

```
android{
  buildFeatures {
    buildConfig = true
  }
}
```

platforms/android/gradle.properties
```diff
@@ -21,6 +21,8 @@ kotlin.code.style=official
...
+android.nonFinalResIds=false
```
julioromano, line 25: Defaults to true in AGP 8.
Do you really need this false? Are you relying on the actual value of some res id being const?

platforms/android/library/build.gradle
```diff
@@ -43,11 +44,11 @@ android {
...
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
+        targetCompatibility JavaVersion.VERSION_17
     }
     kotlinOptions {
-        jvmTarget = '1.8'
+        jvmTarget = '17'
```
julioromano, line 51: This can be replaced by:

```
kotlin {
  jvmToolchain(17)
}
```

NB: This goes outside of the `android{}` block.

julioromano (approved):
Proactive LGTM! Wrote some other comments.

platforms/android/gradle.properties
jonnyandrew: TIL, thanks!

platforms/android/gradle.properties
jonnyandrew: That's the first and last time I'll use the AGP update assistant - I'd trusted that it would output sensible defaults but apparently not!

PR #758, opened 2023-08-10
[Android] Expose style configuration for compose

jonnyandrew:
## Changes

- Expose programmatic view API for setting style
- Expose compose API for setting style

## Context

Previously, the only way to apply styles was through the view XML properties.

### DIFF
--- a/platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
+++ b/platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
@@ -11,6 +11,7 @@ import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.Surface
 import androidx.compose.ui.Modifier
 import io.element.android.wysiwyg.compose.RichTextEditor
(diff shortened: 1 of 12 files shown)


platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
```diff
@@ -33,6 +34,7 @@ class MainActivity : ComponentActivity() {
                         RichTextEditor(
                             state = state,
                             modifier = Modifier.fillMaxWidth(),
+                            style = RichTextEditorDefaults.style(),
                         )
```
julioromano, line 38: Can't we use the default value instead?

julioromano (approved):

platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
jonnyandrew: That's true although, this being an example app, I wanted to show this being set explicitly. Perhaps it'd make more sense to use a non-default value here though!

PR #759, opened 2023-08-10
[iOS] Handle paste and key commands

aringenbach:
* Add an API to delegate items paste handling to the hosting application.
* Add an API to allow the hosting application to override some key commands behaviour (e.g. press `enter` to send a message) => This could be expanded further in order to handle delete word / delete line for iOS user using a physical keyboard if the default behaviour needs improvement.

Unfortunately I couldn't build some UITests for this bit because the key input API for XCTest is only available for a MacOS environment and not for iOS. Will come back with some tests for this later if I find something useful.


alfogrillo (approved):
Don't have much context yet.
But the code LGTM!

PR #760, opened 2023-08-10
[Android] Publish compose library artifact

jonnyandrew:
## Changes

- Publish the `library-compose` module as `io.element.android:wysiwyg-compose`
- Add usage documentation ([rendered](https://github.com/matrix-org/matrix-rich-text-editor/blob/jonny/publish/platforms/android/README.md)

## Context

### DIFF
--- a/platforms/android/library-compose/build.gradle
+++ b/platforms/android/library-compose/build.gradle
@@ -1,3 +1,5 @@
+import com.vanniktech.maven.publish.AndroidSingleVariantLibrary
+
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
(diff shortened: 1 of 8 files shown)


platforms/android/library-compose/build.gradle
```diff
@@ -9,6 +11,16 @@ if (project.hasProperty("coverage")) {
     apply plugin: 'jacoco'
 }
 
+apply plugin: "com.vanniktech.maven.publish.base"
```
julioromano, line 14: Apply plugin statements should be avoided, can you use this plugin from the `plugins{}`` block instead?

```
id 'com.vanniktech.maven.publish'
```

julioromano (approved):
Proactive LGTM with a comment!

platforms/android/library-compose/build.gradle
jonnyandrew: Good point, I've updated this in the view library as well.

PR #761, opened 2023-08-11
Fix `focused` publishing

aringenbach:
This change:

* Fixes an issue around publishing the focused value within view update (i.e. Xcode runtime purple warning)
* Moves the library iOS requirement to iOS 15 (in order to use `FocusState` and remove the need of an external `focused` binding)
* Adds a UI test for toggling the focus state (i.e. simulates the interaction with a hosting app updating from outside)
* Revamps a bit how the buttons for debug actions of the example app work to save some place for debug AST / sent message, etc:

Simulator Screenshot - iPhone 14 Pro - 2023-08-11 at 14 11 48


alfogrillo (approved):
LGTM

PR #767, opened 2023-08-16
[Android] Add preview support for compose

jonnyandrew:
## Changes

- Allow state to be created with an initial value
- Add preview to the example app
- Support compose preview by replacing the rich text editor with a plain EditText when in preview mode

## Context

### DIFF
--- a/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
+++ b/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
@@ -2,11 +2,13 @@ package io.element.android.wysiwyg.compose
 
 import android.os.Build
 import android.view.View
+import androidx.appcompat.widget.AppCompatEditText
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.DisposableEffect
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.toArgb
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalInspectionMode
 import androidx.compose.ui.viewinterop.AndroidView
 import androidx.core.content.ContextCompat
 import androidx.core.widget.addTextChangedListener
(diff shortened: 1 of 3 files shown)


platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -106,12 +120,51 @@ fun RichTextEditor(
...
+    if (!LocalInspectionMode.current) {
+        throw IllegalStateException("PreviewEditor should only be used in preview mode")
+    }
+
+    val context = LocalContext.current
+
+    AndroidView(
+        modifier = modifier,
+        factory = {
+            val view = AppCompatEditText(context).apply {
```
julioromano, line 143: Why not just a material `TextField` ? It'd be a one-liner

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditorState.kt
```diff
@@ -22,7 +22,19 @@ import uniffi.wysiwyg_composer.MenuAction
...
  */
 @Stable
-class RichTextEditorState internal constructor() {
+class RichTextEditorState internal constructor(
+    initialHtml: String = "",
+) {
+    /**
+     * The content of the editor as HTML formatted for sending as a message.
+     */
+    var messageHtml by mutableStateOf("")
```
julioromano, line 31: `var messageHtml by mutableStateOf(initialHtml)` ? (removing the `init{}` block)

julioromano (approved):
Just a few suggestion, but feel free to ignore them

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jonnyandrew: I wanted the preview editor to match the real editor as closely as possible. Using an `EditText` for the preview allows us to make changes to the real editor that will automatically get reflected in the preview, whereas using `TextField` would make it harder to get it looking pixel perfect.

PR #768, opened 2023-08-17
Fix iOS lib size

aringenbach:
* Fixes an issue with library path uniffi build with `lib` crate type that made the iOS static lib go from 35-40 to 80 MB.
* Uses this original fix: https://github.com/mozilla/uniffi-rs/pull/1671

I haven't checked if this affects Android dylib size as well, but it might

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,9 +18,9 @@ rust-version = "1.71"
 [workspace.dependencies]
 # We use a custom version of Uniffi that renames symbols that might clash
 # with other Rust libraries. See https://github.com/mozilla/uniffi-rs/issues/1670
-uniffi = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
-uniffi_macros = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
-uniffi_build = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -18,9 +18,9 @@ rust-version = "1.71"
 [workspace.dependencies]
 # We use a custom version of Uniffi that renames symbols that might clash
 # with other Rust libraries. See https://github.com/mozilla/uniffi-rs/issues/1670
-uniffi = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
-uniffi_macros = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
-uniffi_build = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "7fea5def9b7a00cec35b94bd1916172cd0ca60fb" }
+uniffi = { git = "https://github.com/aringenbach/uniffi-rs.git", rev = "737286e8fb43cda66f17b7efd6d3b351d72d70e1" }
```
jonnyandrew, line 21: I see you've cherry-picked https://github.com/mozilla/uniffi-rs/pull/1671 into your fork.

I think it's OK but perhaps it'd be easier to maintain the fork if we wait for an upstream release and rebase your changes onto it?

jonnyandrew (approved):

Cargo.toml
aringenbach: Yeah definitely, this change is not released yet I think so I figured this would be the easier to get this fix right away.

I saw discussions again in the uniffi team public channel around the issue we are solving with that fork, so hopefully we don't need to maintain that for too long 🤞

jonnyandrew: Ah great

PR #769, opened 2023-08-17
Rework composer text placeholder

aringenbach:
Reworks the current placeholder of the composer into a more straightforward SwiftUI component.
It also fixes an issue with the ideal height not being published right away (hence, staying at its default `CGFloat.zero`) that could lead to a misplaced placeholder.


langleyd (approved):
LGTM

PR #770, opened 2023-08-17
Fix invalid markdown output caused by special characters

jonnyandrew:
## Problem

Invalid markdown is generated if a user types special characters.

For example, it would be impossible to create the following formatted text because the underscores would be interpreted as markdown special characters:

_italic\_ italic \_italic_

## Solution

Escape special characters in markdown output.

|        | Before | After |
|--------|--------|-------|
Markdown | `_italic_ italic _italic_` | `_italic\_ italic \_italic_` |
Rendered | _italic_ italic _italic_ | _italic\_ italic \_italic_

### DIFF
--- a/crates/wysiwyg/src/dom/nodes/text_node.rs
+++ b/crates/wysiwyg/src/dom/nodes/text_node.rs
@@ -293,10 +293,31 @@ where
     fn fmt_markdown(
         &self,
         buffer: &mut S,
-        _options: &MarkdownOptions,
+        options: &MarkdownOptions,
         _as_message: bool,
     ) -> Result<(), MarkdownError<S>> {
-        buffer.push(self.data.to_owned());
+        if options.contains(MarkdownOptions::NO_ESCAPE) {
+            buffer.push(self.data());
+            return Ok(());
+        }
+
+        let mut escaped = S::default();
+
+        for c in self.data().chars() {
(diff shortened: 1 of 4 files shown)


crates/wysiwyg/src/dom/nodes/text_node.rs
```diff
@@ -293,10 +293,31 @@ where
...
+
+        let mut escaped = S::default();
+
+        for c in self.data().chars() {
+            match c {
+                // https://spec.commonmark.org/0.30/#ascii-punctuation-character
+                '!' | '"' | '#' | '$' | '%' | '&' | '\'' | '(' | ')' | '*'
+                | '+' | ',' | '-' | '.' | '/' | ':' | ';' | '<' | '=' | '>'
+                | '?' | '@' | '[' | '\\' | ']' | '^' | '_' | '`' | '{'
+                | '|' | '}' | '~' => {
```
jonnyandrew, line 312: Though it's possible to escape all of these characters, I'm not sure it's actually necessary. I couldn't find a good source for which characters would always need to be escaped though.

aringenbach (approved):
LGTM

Don't think we need to overthink the set of escaped characters too much as editing a rich text mode created content in markdown mode doesn't seem like such a frequent thing.

PR #771, opened 2023-08-18
[Android] Revert Java toolchain JDK

jonnyandrew:
## Problem

Updating Gradle and AGP to version 8 (#757), it was necessary to update the JDK used to run Gradle to 17. However, during this update, the toolchain JDK used to compile the source code was also updated to 17 which is not compatible with Android SDK API 33.

## Solution

Revert to Java toolchain JDK to version 11.

### DIFF
--- a/platforms/android/example-compose/build.gradle
+++ b/platforms/android/example-compose/build.gradle
@@ -26,8 +26,8 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
     }
     buildFeatures {
         compose true
@@ -43,7 +43,7 @@ android {
 }
 
 kotlin {
-    jvmToolchain(17)
+    jvmToolchain(11)
 }
 
 dependencies {
--- a/platforms/android/example-view/build.gradle
+++ b/platforms/android/example-view/build.gradle
@@ -23,8 +23,8 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
     }
 
     viewBinding {
@@ -33,7 +33,7 @@ android {
 }
 
 kotlin {
-    jvmToolchain(17)
+    jvmToolchain(11)
 }
 
 dependencies {
--- a/platforms/android/library-compose/build.gradle
+++ b/platforms/android/library-compose/build.gradle
@@ -44,8 +44,8 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility JavaVersion.VERSION_11
+        targetCompatibility JavaVersion.VERSION_11
     }
 
     testCoverage {
@@ -64,7 +64,7 @@ android {
 }
 
 kotlin {
-    jvmToolchain(17)
+    jvmToolchain(11)
 }
 
 dependencies {
--- a/platforms/android/library/build.gradle
+++ b/platforms/android/library/build.gradle
@@ -45,8 +45,8 @@ android {
     }
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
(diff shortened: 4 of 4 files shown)


julioromano:
If api33 weren't comptaible with java 17 bytecode how could https://github.com/matrix-org/matrix-rich-text-editor/pull/757 pass CI checks?

julioromano:
> I'm not 100% sure how it worked. My assumption is that it's because the test app is also configured to compile with JDK 17. But importing it into Element Android (compileSdk 33 / JDK 11) reveals a [problem](https://github.com/vector-im/element-android/actions/runs/5901258208/job/16007097128):
> 
> ```
> bad class file: [..] jetified-wysiwyg-2.5.0-api.jar(/io/element/android/wysiwyg/[..].class)
>     class file has wrong version 61.0 [Java 17], should be 55.0 [Java 11]
> ```
> 
> When I checked [Java versions in Android builds](https://developer.android.com/build/jdks#jdk-android-studio) docs, I saw the following compatibility table:
> 
> > Use the table below to determine which Java version is supported by
> > each Android API, and where to find details on which Java APIs are available.
> > 
> > 
> > 
> > Android
> > Java
> > 
> > 
> > 
> > 
> > 14 (API 34)
> > 17
> > 
> > 
> > 13 (API 33)
> > 11

Oh I see,

> I'm not 100% sure how it worked. My assumption is that it's because the test app is also configured to compile with JDK 17. But importing it into Element Android (compileSdk 33 / JDK 11) reveals a [problem](https://github.com/vector-im/element-android/actions/runs/5901258208/job/16007097128):
> 
> ```
> bad class file: [..] jetified-wysiwyg-2.5.0-api.jar(/io/element/android/wysiwyg/[..].class)
>     class file has wrong version 61.0 [Java 17], should be 55.0 [Java 11]
> ```
> 
> When I checked [Java versions in Android builds](https://developer.android.com/build/jdks#jdk-android-studio) docs, I saw the following compatibility table:
> 
> > Use the table below to determine which Java version is supported by
> > each Android API, and where to find details on which Java APIs are available.
> > 
> > 
> > 
> > Android
> > Java
> > 
> > 
> > 
> > 
> > 14 (API 34)
> > 17
> > 
> > 
> > 13 (API 33)
> > 11



> I'm not 100% sure how it worked. My assumption is that it's because the test app is also configured to compile with JDK 17. But importing it into Element Android (compileSdk 33 / JDK 11) reveals a [problem](https://github.com/vector-im/element-android/actions/runs/5901258208/job/16007097128):
> 
> ```
> bad class file: [..] jetified-wysiwyg-2.5.0-api.jar(/io/element/android/wysiwyg/[..].class)
>     class file has wrong version 61.0 [Java 17], should be 55.0 [Java 11]
> ```
> 
> When I checked [Java versions in Android builds](https://developer.android.com/build/jdks#jdk-android-studio) docs, I saw the following compatibility table:
> 
> > Use the table below to determine which Java version is supported by
> > each Android API, and where to find details on which Java APIs are available.
> > 
> > 
> > 
> > Android
> > Java
> > 
> > 
> > 
> > 
> > 14 (API 34)
> > 17
> > 
> > 
> > 13 (API 33)
> > 11

Oh yes, that's because EA is still build with v11 bytecode. If we wish to keep compatibility with EA too then we have no choice but to lower ours too.

julioromano (approved):
Can you please first check if it works by keeping: `jvmToolchain(17)`

and only changing :
```
   sourceCompatibility JavaVersion.VERSION_11
   targetCompatibility JavaVersion.VERSION_11
```
 
 If it doesn't then lower the jvmToolchain too. Thanks!

jonnyandrew:
Ok, I tried this but the build fails with an error:

```
> 'compileDebugJavaWithJavac' task (current target is 11) and 'compileDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
```

If I remove `targetCompatibility JavaVersion.VERSION_11`, then it's the same story as before - bytecode is not compatible with the EA build.

julioromano:
> Ok, I tried this but the build fails with an error:
> 
> ```
> > 'compileDebugJavaWithJavac' task (current target is 11) and 'compileDebugKotlin' task (current target is 17) jvm target compatibility should be set to the same Java version.
> ```
> 
> If I remove `targetCompatibility JavaVersion.VERSION_11`, then it builds but it's the same story as before - bytecode is not compatible with the EA build.

Thanks for checking! We'll have to keep everything at 11 then as long as we want to maintain compatibility with legacy element.

jonnyandrew:
Merging with test failures as these are unrelated and fixed in matrix-org/matrix-rich-text-editor#772

PR #772, opened 2023-08-18
[Android] Update failing test

jonnyandrew:
## Problem

Markdown behaviour changed due to https://github.com/matrix-org/matrix-rich-text-editor/pull/770 and Android tests were not updated to reflect this.

## Solution

Update the tests.

Note that this particular failure is caused by a comma `,` being escaped in the markdown output. Although we could probably leave commas unescaped, in absence of a proper strategy for which characters to escape and under what conditions, we're escaping all ASCII punctuation characters: https://github.com/matrix-org/matrix-rich-text-editor/pull/770#discussion_r1297304953


aringenbach (approved):
LGTM

PR #773, opened 2023-08-18
Disable SwiftFormat's `redundantStaticSelf`

aringenbach:
Needed until https://github.com/nicklockwood/SwiftFormat/issues/1510 is fixed

### DIFF
--- a/platforms/ios/example/.swiftformat
+++ b/platforms/ios/example/.swiftformat
@@ -1,5 +1,6 @@
 --swiftversion 5.6
 
+--disable redundantStaticSelf
 --disable wrapMultiLineStatementBraces
 --disable hoistPatternLet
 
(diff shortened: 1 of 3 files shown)


platforms/ios/example/.swiftformat
```diff
@@ -1,5 +1,6 @@
 --swiftversion 5.6
 
+--disable redundantStaticSelf
```
jonnyandrew, line 3: Perhaps worth adding a link to the related issue here (unless we plan on keeping this rule disabled forever)?

jonnyandrew (approved):

platforms/ios/example/.swiftformat
aringenbach: [ae07ae3](https://github.com/matrix-org/matrix-rich-text-editor/pull/773/commits/ae07ae314af7e10ac9835c1c6b029afe33205811)

PR #774, opened 2023-08-18
Version 2.6.0

jonnyandrew:


aringenbach (approved):
LGTM

jonnyandrew:
The iOS tests seem to be failing - @aringenbach will it be OK to fix in a follow-up release? Or should we wait to fix them?

aringenbach:
> The iOS tests seem to be failing - @aringenbach will it be OK to fix in a follow-up release? Or should we wait to fix them?

@jonnyandrew Definitely fine to fix later, these are related to plain text mode anyway, and only fails on GH Actions.

PR #782, opened 2023-08-23
Set Swift library to be dynamic

aringenbach:
Setting the Swift library as dynamic should avoid having issues with duplicated symbols. 

Note: we are though still using a fork of `uniffi-rs` that solves that issue through renaming, because it also contains https://github.com/mozilla/uniffi-rs/pull/1671 that we currently need in order to build in library mode without library size issues (see https://github.com/matrix-org/matrix-rich-text-editor/pull/768). I'll create an issue to replace that fork when a version of uniffi containing the fix lands.


langleyd (approved):
Sounds good, thanks

PR #793, opened 2023-08-31
Remove `WysiwygAction`

aringenbach:
This can easily be confused with the `ComposerAction` from Rust bindings. 
Bindings are always exposed to the hosting app anyway, so this makes the implementation and usage more straightforward


alfogrillo (approved):
LGTM!

PR #796, opened 2023-09-07
[Android] Add fake mode for compose state for testing

jonnyandrew:
## Changes

- Add fake mode for the compose state for testing and preview purposes
- Relax visibility of the compose state constructor

## Context

As this library inherently requires the underlying Rust model to function properly, it makes testing in a JVM environment difficult. A simple fake implementation allows unit tests to observe some simple state changes after calling functions on the state holder.

- See https://github.com/vector-im/element-x-android/pull/1172#discussion_r1317449784


langleyd (approved):
Nice, LGTM

PR #798, opened 2023-09-08
[Android] Fix performance issue in recomposition

jonnyandrew:
## Problem

Setting the style of the underlying `EditText` is an expensive operation due to the fact that it forces a re-render. In a traditional view-based app this isn't a problem however, in compose, the wrapper's `update` block may be called frequently regardless of whether the `style` argument has changed, causing the the expensive operation to be triggered frequently.

## Solution

This change ensures that setting the style in the underlying `EditText` only takes effect and triggers an expensive re-render if the new style is different to the old style.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/view/StyleConfig.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/view/StyleConfig.kt
@@ -1,7 +1,7 @@
 package io.element.android.wysiwyg.view
 
-import android.graphics.drawable.Drawable
 import androidx.annotation.ColorRes
+import androidx.annotation.DrawableRes
 import androidx.annotation.Px
 import io.element.android.wysiwyg.EditorEditText
 
@@ -50,10 +50,10 @@ data class InlineCodeStyleConfig(
     @Px val horizontalPadding: Int,
     @Px val verticalPadding: Int,
     val relativeTextSize: Float,
-    val singleLineBg: Drawable,
-    val multiLineBgLeft: Drawable,
-    val multiLineBgMid: Drawable,
-    val multiLineBgRight: Drawable,
+    @DrawableRes val singleLineBg: Int,
+    @DrawableRes val multiLineBgLeft: Int,
+    @DrawableRes val multiLineBgMid: Int,
+    @DrawableRes val multiLineBgRight: Int,
 )
 
 /**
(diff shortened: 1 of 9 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/view/StyleConfig.kt
```diff
@@ -68,7 +68,7 @@ data class CodeBlockStyleConfig(
     @Px val leadingMargin: Int,
     @Px val verticalPadding: Int,
     val relativeTextSize: Float,
-    val backgroundDrawable: Drawable,
+    @DrawableRes val backgroundDrawable: Int,
```
jonnyandrew, line 71: Switch to storing the resource IDs here so that the style config classes can be compared for equality.

julioromano:
Just wanted to make sure we aren't fixing the wrong part here: Why is the `update` lambda called so often? In theory it should only be called:
1) On composition (right after the `factory` lambda has run).
    This should generally not be a performance problem.
2) Whenever any state reads within the `update` lambda itself change.
     This looks more like the possible culprit. Though it's definitely not normal. So it is worth investigating.

Could we try to do this optimization closer to the source of the issue by reducing the number of times unwanted state changes trigger call to `update`?

My 2 cents:
- The compose compiler doesn't see the `RichTextEditorStyle` as stable so it will trigger unnecessary recompositions (hence reruns of the `update` lambda).
- The `onError` lambda is not stable by definition so it might also cause unwanted recompositions (but take this with a grain of salt, I hereby summon @jmartinesp for more help about it).

This page explains how to debug stability issues in compose: https://developer.android.com/jetpack/compose/performance/stability/diagnose

jmartinesp:
> My 2 cents:
> 
> * The compose compiler doesn't see the `RichTextEditorStyle` as stable so it will trigger unnecessary recompositions (hence reruns of the `update` lambda).
> * The `onError` lambda is not stable by definition so it might also cause unwanted recompositions (but take this with a grain of salt, I hereby summon @jmartinesp for more help about it).

AFAIK it shouldn't be any of those 2 possible issues that's the root cause here:

1. `RichTextEditorStyle` is a data class containing other data classes with either primitives or stable classes (unless I missed something), so it should also be stable.
2. The `onError` lambda passed is a function reference in EXA which (if I'm not mistaken) can be compared, unlike a lambda, so it shouldn't cause recompositions.

julioromano:
Then it's even more worth doing some debugging into why that lambda is called so often.

jonnyandrew:
Aha, I thought it seemed a bit fishy that the update block would be triggered despite the variables used there not being updated!

Thanks for the info on debugging stability issues - it turns out instability is the root cause. I'll open a new PR with the fix.

PR #803, opened 2023-09-12
[Android] Fix performance issue caused by excessive recomposition

jonnyandrew:
## Problem

The `update` block of the `RichTextEditor` composable is called regardless of whether the style argument has changed, causing an expensive operation to be triggered frequently, reducing performance.

## Solution

The root cause is that the `RichTextEditorState` class is not stable due to the instability of its `viewConnection` property, causing the `RichTextEditor` composable to recompose more than is necessary. The solution is to make the `viewConnection` property stable by wrapping it in `MutableState`.

#### Before
```
unstable class RichTextEditorState {
  runtime var viewConnection: ViewConnection?
  ...
  <runtime stability> = Unstable
}
```

#### After
```
stable class RichTextEditorState {
  stable var viewConnection$delegate: MutableState<ViewConnection?>
  ...
  <runtime stability> = Stable
}
```


julioromano (approved):
Super!

PR #810, opened 2023-09-14
[iOS] Add support to custom placeholder color

alfogrillo:


jonnyandrew (commented):
Does this take dark mode into account?

alfogrillo:
> Does this take dark mode into account?

Yes, it does. A `Color` has both states (dark, light) into it.

nimau (approved):
SGTM

PR #813, opened 2023-09-15
[Android] Add customisation options for links and code

jonnyandrew:
## Changes

There are a few semi-related changes in this PR which can be reviewed commit by commit (or split into separate PRs if needed).

- Allow customisation of link color
- Allow customisation of code and inline code backgrounds
- Refactor code and inline code background styles (so that apps do not need to handle drawables directly)
- Add useful compose debugging config
- Fix a few mutability/instability issues in the example app and core view library

## Context

- https://github.com/vector-im/element-x-android/issues/1315

### DIFF
--- a/platforms/android/library/build.gradle
+++ b/platforms/android/library/build.gradle
@@ -51,6 +51,11 @@ android {
 
     buildFeatures {
         buildConfig = true
+        compose true // Enable classes to be marked as stable
+    }
+
+    composeOptions {
+        kotlinCompilerExtensionVersion '1.5.1'
     }
 
     testOptions {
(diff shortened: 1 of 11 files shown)


platforms/android/library/build.gradle
```diff
@@ -51,6 +51,11 @@ android {
 
     buildFeatures {
         buildConfig = true
+        compose true // Enable classes to be marked as stable
+    }
+
+    composeOptions {
+        kotlinCompilerExtensionVersion '1.5.1'
```
julioromano, line 58: NIT: we are at 1.5.3 now.

julioromano (approved):
LGTM! Just a minor nit!

platforms/android/library/build.gradle
jonnyandrew: Thanks, I used this version to align with the other modules in the project but will update separately :+1:

PR #817, opened 2023-09-19
[iOS] Add workaround for the "out of bounds index" error

alfogrillo:
This PR is work around an "out of bounds index" on the `WysiwygComposerViewModel`.
It triggered when the `func setHtmlContent(_ html:)` was fed with a text like this `<p><strong>rteoff</strong></p>\n`.
The same string without a trailing `\n` worked well.

related issue: https://github.com/matrix-org/matrix-rich-text-editor/issues/818

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
@@ -451,9 +451,16 @@ private extension WysiwygComposerViewModel {
             let attributed = try HTMLParser.parse(html: html,
                                                   style: parserStyle,
                                                   mentionReplacer: mentionReplacer)
-            // FIXME: handle error for out of bounds index
             let htmlSelection = NSRange(location: Int(start), length: Int(end - start))
-            let textSelection = try attributed.attributedRange(from: htmlSelection)
+            let textSelection: NSRange
+
+            // Workaround for the "out of bounds index" error
+            if start == end, start > attributed.length {
+                textSelection = .init(location: attributed.length, length: 0)
+            } else {
+                textSelection = try attributed.attributedRange(from: htmlSelection)
+            }
+
             attributedContent = WysiwygComposerAttributedContent(text: attributed,
                                                                  selection: textSelection,
                                                                  plainText: model.getContentAsPlainText())
--- a/platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Components/WysiwygComposerView/WysiwygComposerViewModelTests+SetContent.swift
+++ b/platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Components/WysiwygComposerView/WysiwygComposerViewModelTests+SetContent.swift
@@ -20,6 +20,7 @@ import XCTest
 
 private enum Constants {
     static let sampleHtml = "some <strong>bold</strong> text"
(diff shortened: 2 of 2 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -451,9 +451,16 @@ private extension WysiwygComposerViewModel {
             let attributed = try HTMLParser.parse(html: html,
                                                   style: parserStyle,
                                                   mentionReplacer: mentionReplacer)
-            // FIXME: handle error for out of bounds index
             let htmlSelection = NSRange(location: Int(start), length: Int(end - start))
-            let textSelection = try attributed.attributedRange(from: htmlSelection)
+            let textSelection: NSRange
+
+            // Workaround for the "out of bounds index" error
```
jonnyandrew, line 457: Given we think this workaround is needed because the rust model is outputting invalid selection indices, perhaps we can create an issue for that and link it here?

alfogrillo: Here -> https://github.com/matrix-org/matrix-rich-text-editor/issues/818

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -451,9 +451,16 @@ private extension WysiwygComposerViewModel {
             let attributed = try HTMLParser.parse(html: html,
                                                   style: parserStyle,
                                                   mentionReplacer: mentionReplacer)
-            // FIXME: handle error for out of bounds index
             let htmlSelection = NSRange(location: Int(start), length: Int(end - start))
```
nimau, line 454: Should we check that start <= end?

nimau (commented):
SGTM, just a small question.

alfogrillo:
> Is there any way we can add an UI/integration test for this fix? I.e. call `setHtml()` on the editor and check the editor still has content and doesn't go blank?

Added one [here](https://github.com/matrix-org/matrix-rich-text-editor/pull/817/commits/826ba2ec6e52da78be28bd268b5c9e2f754e9920)

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
alfogrillo: Not sure this will ever happen. 
But if it does I'm happy to go with the "old branch" which in the end should throw an error. 👍🏻

nimau (approved):

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
jonnyandrew: Sorry - I meant to add a link to the issue in the code comment itself - as it might help whoever runs into this workaround later. But it doesn't matter too much, given they can probably find this PR.

alfogrillo:
Not needed anymore.
The issue has been fixed by https://github.com/matrix-org/matrix-rich-text-editor/pull/820

PR #822, opened 2023-09-22
[Android] Add support for subcomposition and sharing state

jonnyandrew:
## Changes

- Add support for subcomposition. By adding a new a mode that doesn't register any state updates, the editor can be composed purely for measurement purposes and not trigger a state update.
- Add support for sharing state between editors at different call-sites; allow the editor to be constructed at a new call site without side effects. The AndroidView `factory` and `update` blocks can be called again by the replacement editor instance safely.
- Preserve trailing new lines on restoring state by saving and restoring the internal HTML representation of the editor rather than the message HTML.
- Make the state observable so that, when an editor leaves the composition, it doesn't need to clean up its reference from the state and cause unwanted state updates.
- Fix some issues with unstable classes in the state.
- Update some compose dependencies including compose compiler.

## Context

To enable the editor to be embedded in complex layouts that require pre-measurement and/or multiple layout trees.

- Part of https://github.com/vector-im/element-meta/issues/1973


platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -14,25 +14,31 @@ import androidx.compose.ui.viewinterop.AndroidView
...
+ *
  * @param state The state holder for this composable. See [rememberRichTextEditorState].
  * @param modifier The modifier for the layout
+ * @param registerStateUpdates If true, register the state for updates.
  * @param style The styles to use for any customisable elements
  * @param onError Called when an internal error occurs
  */
 @Composable
 fun RichTextEditor(
     state: RichTextEditorState,
```
julioromano, line 39: would it make sense to have a default value here?
```suggestion
    state: RichTextEditorState = rememberRichTextEditorState(),
```

julioromano: Actually I see that some default params are in the `private fun RealEditor` composable. They should probably be hoisted to the public one.

julioromano (approved):
Just a few nits hoping they area useful. Sorry for not being more of help here.

Although I don't understand what's the reason for switching off `registerStateUpdates` in a subcomposition. Without state updates the RTE is pretty much useless isn't it?

jonnyandrew:
> Although I don't understand what's the reason for switching off registerStateUpdates in a subcomposition. Without state updates the RTE is pretty much useless isn't it?

In a subcomposition the editor is composed twice, once for measurement and once again to be placed in the layout given the first measurement. During the first measurement I need the editor to just be a dumb view and read the state but not try to update it in any way that might impact the second step. Hope that makes sense?

julioromano:
> > Although I don't understand what's the reason for switching off registerStateUpdates in a subcomposition. Without state updates the RTE is pretty much useless isn't it?
> 
> In a subcomposition the editor is composed twice, once for measurement and once again to be placed in the layout given the first measurement. During the first measurement I need the editor to just be a dumb view and read the state but not try to update it in any way that might impact the second step. Hope that makes sense?
> 
> An alternative could be to create a new instance of the state for every subcomposition but we'd also have the issue of the first composition requesting focus each frame, which I'm not sure how to avoid.

I still did not understand this completely. Can you point me to a piece of code that shows this? (using the RTE in a subcomp).

PR #823, opened 2023-09-26
[iOS] Fix pending format inconsistencies

aringenbach:
Fixes https://github.com/vector-im/element-x-ios/issues/1672

Also fixes a related issue where de-selecting and selecting a list item back would not apply the formats either. 
Added ViewModel unit tests for both cases.


alfogrillo (approved):
LGTM

PR #825, opened 2023-09-26
[Android] Fix selection state restoration

jonnyandrew:
## Problem

The selection position is not preserved when sharing state between two composable editors (which represent the same editor but exist at different call sites).

## Solution

- When creating a new editor, apply the selection position found in the state
- Also, save the selection when saving state

### DIFF
--- a/platforms/android/library-compose/build.gradle
+++ b/platforms/android/library-compose/build.gradle
@@ -4,6 +4,7 @@ import com.vanniktech.maven.publish.SonatypeHost
 plugins {
     id 'com.android.library'
     id 'org.jetbrains.kotlin.android'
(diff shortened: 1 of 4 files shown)


julioromano (approved):
LGTM, though I thought it wasn't supported to share the state between two editors.

From the docstring:
> Ensure that [RichTextEditorState] is not shared between multiple [RichTextEditor] composables that are displayed at the same time.

jonnyandrew:
Sorry, that was a bit unclear. This is not meant to support two different editors being displayed at the same time and sharing the same state; it's to support the same editor being displayed in different layouts (i.e. at different call sties in the compose tree).

PR #829, opened 2023-09-29
Fix background drawing when clearing quote / code block

aringenbach:
Fixes issue with background not always disappearing when removing a quote / code block in the Rich Text Editor. 

The issue occurred when clearing an empty quote or code block, if we backspaced the last character we update the `UITextView` attributedContent to a version without the ZWSP + attributes that should hold the information about having a quote / code block there.

This fixes this specific issue by ensuring we always trigger the redraw of the `UITextView` background layers when the model is completely empty, avoiding any remnants from previous content.

One issue remained, described in https://github.com/matrix-org/matrix-rich-text-editor/issues/830

Partial fix for https://github.com/vector-im/element-x-ios/issues/1747


alfogrillo (approved):
LGTM

PR #831, opened 2023-09-29
Fix reconciliate breaking with leading whitespaces

aringenbach:
Fix reconciliate triggering with wrong indexes when there are user created leading whitespaces

Fixes https://github.com/vector-im/element-x-ios/issues/1803


alfogrillo (approved):
LGTM

PR #833, opened 2023-10-02
Fix delete word not replacing text

### LINKED ISSUE #808: Web editor not updating content after deleting a word (ctrl+backspace)
## Summary
When deleting a word with `ctrl+backspace` in the web composer nothing happens.

## Expected
Word is removed from the editor after deleting it with `ctrl+backspace`

## Actual
After deleting a word with `ctrl+backspace`, the content of the editor does not get updated. Nothing happens visually.

The issue also occures in the [editor example](https://matrix-org.github.io/matrix-rich-text-editor/) and current development release of develop.element.io

ezgif com-video-to-gif

## Problem
Internally the model of the editor seems to update properly, but the visible content does not. After pressing anything but `ctrl+backspace` the editor content gets updated and the deleted word is also gone (expected behaviour in the first place).

I tracked a part of the problem down to `platforms/web/lib/useListeners/event.ts:205` where `replace_all` returns `undefined` after `ctrl+backspace` and therefore the editor content does not get replaced. I was not able to dig further into the issue due to my lacking experience in rust land.

```typescript
const repl = update.text_update().replace_all;
if (repl) {
    replaceEditor(
        editor,
        repl.replacement_html,
        repl.start_utf16_codeunit,
        repl.end_utf16_codeunit,
    );
    testUtilities.setEditorHtml(repl.replacement_html);
}
```

## Related issues
- Somehow part of [PSB-369](https://element-io.atlassian.net/browse/PSB-369)

[PSB-369]: https://element-io.atlassian.net/browse/PSB-369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

aringenbach:
Should fix #808 

I couldn't explore potential Web related corner cases so it might be better if a Web developer is available to double check this.

### DIFF
--- a/crates/wysiwyg/src/composer_model/delete_text.rs
+++ b/crates/wysiwyg/src/composer_model/delete_text.rs
@@ -267,17 +267,17 @@ where
                     node.offset_is_inside_node(current_offset, &direction);
 
                 // delete to the cursor
-                self.delete_to_cursor(current_position);
+                let update = self.delete_to_cursor(current_position);
 
                 // if we have stopped inside the node and we didn't start at whitespace, stop
                 if offset_is_inside_node && start_type != CharType::Whitespace {
-                    return ComposerUpdate::keep();
+                    return update;
                 }
 
                 // otherwise make a recursive call
                 let next_args = self.get_remove_word_arguments(&direction);
                 match next_args {
-                    None => ComposerUpdate::keep(),
+                    None => update,
                     Some(args) => {
                         let (location, next_type) = args;
                         let type_argument = if offset_is_inside_node {
--- a/crates/wysiwyg/src/tests/test_deleting.rs
+++ b/crates/wysiwyg/src/tests/test_deleting.rs
@@ -925,6 +925,13 @@ fn backspace_mention_from_end() {
 }
 
 #[test]
+fn backspace_word_returns_replace_all_update() {
+    let mut model = cm("Some text with multiple words|");
+    let update = model.backspace_word();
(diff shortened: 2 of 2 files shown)


jonnyandrew:
Thanks for the PR @aringenbach.

> I couldn't explore potential Web related corner cases so it might be better if a Web developer is available to double check this.

I wonder if @florianduros / @andybalaam could help with this?

florianduros:
> Thanks for the PR @aringenbach.
> 
> > I couldn't explore potential Web related corner cases so it might be better if a Web developer is available to double check this.
> 
> I wonder if @florianduros / @andybalaam could help with this?

I tested on Macos (`Option + Delete` shortcut) and it doesn't work on multiple lines:

https://github.com/matrix-org/matrix-rich-text-editor/assets/2621378/b0f8d651-af99-467e-93a8-1a84ebb262a6

aringenbach:
> 
> I tested on Macos (`Option + Delete` shortcut) and it doesn't work on multiple lines:
> 

Yeah I actually tested a bit in the meantime (on Chrome & Safari for MacOS) and it looks like a few whacky things are happening with the delete word feature both with or without this change. 
I think the Rust change here makes sense from the model perspective, but it seems like there might be some tidying up to do around the entire feature that might be out of my current level of expertise 🧐

Velin92 (approved):

PR #835, opened 2023-10-02
Exit build xcframework if swiftformat is unavailable

aringenbach:
On setup on a machine with no version of `swiftformat` available, the `sed` workaround to catch Rust panics will not be executed properly due to the Swift file not being formatted as expected by the script, leaving the resulting library in a state where Rust panics would crash the hosting application instead of recovering.
This change ensures that `make ios` or `./build_xcframework.sh` fails explicitly if the tool is not installed.


alfogrillo (approved):
LGTM

PR #836, opened 2023-10-04
Version 2.14.1

alfogrillo:


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+# [2.14.1] - 2023-10-04
+
+### Fixed
+
+- iOS: Fix background drawing when clearing quote / code block (#829)
+- iOS: Fix reconciliate breaking with leading whitespaces (#831)
+- iOS: Exit build xcframework if swiftformat is unavailable (#835)
```
jonnyandrew, line 9: I don't think this needs to go into the changelog as it only fixes something in the build process

jonnyandrew (approved):

CHANGELOG.md
alfogrillo: Right, I don't think none cares.
Removing it...

PR #840, opened 2023-10-09
MentionDisplayHelper to allow to inject context to properly display mentions in RTE

Velin92:
Since the NSTextAttachmentProvider can only see the text view, and the pill might require a context that holds various data and helpful functions to make the pill render correctly, this is just a markdown protocol that can be injected as a weak var to get such context.

For example on Element X this context could be the room context, useful to get all the informations of the already fetched members, and the image provider.

The using application will already know what type the context is and just type unwrap it.

Also added a flushPills function on the textview


nimau (approved):
SGTM

Velin92:
> SGTM

I might improve it tomorrow, since I was able to run it locally in El-X at the end

Issue #618, opened 2023-03-14
Edit link - allow editing both URL and displayed text - Web


aringenbach:
Alongside this [change](https://github.com/matrix-org/matrix-rich-text-editor/issues/615), it would make sense for the edit link UI to provide a way to edit the text of a link in a easier way.

Proposal: 
Re-use the link creation with text UI, and pre-fill it with the current text content of the link. 
The only drawback would be that if some parts of the link are formatted, this formatting might be lost depending on context.

PR #841, opened 2023-10-10
Edit link with text

aringenbach:
Replacement for #622 

* Adds a Rust API to allow editing text and url of a link at the same time
* Mobile & WASM bindings
* Adds an implementation of that new API in the iOS example app
* **Slight API break** with the change to the `LinkAction` for Edit now having both parameters (but it doesn't seem to break Android example app at least, and setting only the URL the old way is still compatible)

What's missing from there: a bit of boilerplate code to use this in Android applications (and maybe Web ?).

https://github.com/matrix-org/matrix-rich-text-editor/assets/80891108/575cf1e6-8f58-4b5e-9d12-5b12ddfa6d82


jonnyandrew:
- Closes #616
- Closes #617
- Closes #618

aringenbach:
Actually updated with Android code. 
I'm seeing an issue with the compose variant of the example app, but it does seem to happen with the current API too:

1. Write text
2. Select it
3. Create link
4. => doesn't work (probably because opening the dialog changes the selection in the field to a cursor)

CC @jonnyandrew

jonnyandrew:
> Actually updated with Android code. I'm seeing an issue with the compose variant of the example app, but it does seem to happen with the current API too:
> 
> 1. Write text
> 2. Select it
> 3. Create link
> 4. => doesn't work (probably because opening the dialog changes the selection in the field into a cursor)
> 
> 

@aringenbach I pulled main into this work and tested the above and it seems to be fixed, or at least I couldn't reproduce it. I pushed the branch [here](https://github.com/matrix-org/matrix-rich-text-editor/tree/edit-link-with-text) if you wanted to take a look.

aringenbach:
@jonnyandrew Cool, yeah I'm pretty sure it was not directly related (and it is the demo app only anyway)

As far as this feature go, I think it still requires some action from a Web developer to check whatever should be done for the Web to handle that change properly. Probably not too much work, but it needs to be done before this can be released. 

Please feel free to create a new PR and close this one.

PR #842, opened 2023-10-11
Use message content apis

langleyd:
Use the correct apis for message content as outlined in https://github.com/matrix-org/matrix-rich-text-editor/issues/679


jonnyandrew (approved):
Thanks!

PR #843, opened 2023-10-11
[iOS] expose the set at room mention function

Velin92:
This was missing a function exposure from the view model, in this way now we have an easy and simple way to set at the current suggestion pattern the @room mention on iOS


jonnyandrew (commented):
Please could you add iOS tests for the new `setAtRoomMention` functions?

alfogrillo (approved):
LGTM!

Velin92:
> Please could you add iOS tests for the new `setAtRoomMention` functions?

added them!

jonnyandrew (commented):
Thanks for the tests!

PR #849, opened 2023-10-18
`get_mentions_state()` implemented on the Rust side (only for users and @room)

Velin92:
Implemented and exposed a function to get the current state of mentions in RTE.
This will allow us to efficiently get the current state of mentions, which might be very useful for intentional mentions.


crates/wysiwyg/src/composer_model/mentions.rs
```diff
@@ -13,15 +13,43 @@
...
+        for node in self.state.dom.iter_mentions() {
+            match node.kind() {
+                MentionNodeKind::AtRoom => {
+                    mentions_state.has_at_room_mention = true
+                }
+                MentionNodeKind::MatrixUri { mention } => {
+                    match mention.kind() {
+                        matrix_mentions::MentionKind::Room => {
+                            // Not required yet
+                        }
```
jonnyandrew, line 40: The omission of rooms in the result could be a bit confusing given the name of the function because it sounds like it will return all mentions. Can I suggest either:
- Update the name of the function and result to `get_intentional_mentions() -> IntentionalMentions` and add documentation that references [MSC3952](https://github.com/matrix-org/matrix-spec-proposals/pull/3952) so that it's implied that only spec'ed mentions will be returned. Or,
- Include room IDs in the result and leave it up to the client to ignore them.

crates/wysiwyg/src/mentions_state.rs
```diff
@@ -0,0 +1,21 @@
+// Copyright 2022 The Matrix.org Foundation C.I.C.
```
jonnyandrew, line 1:
```suggestion
// Copyright 2023 The Matrix.org Foundation C.I.C.
```

crates/wysiwyg/src/tests/test_mentions.rs
```diff
@@ -594,6 +594,73 @@ fn can_insert_at_room_mention() {
...
+fn get_mentions_state_for_user_mention() {
+    let model = cm("<p>hello <a data-mention-type=\"user\" href=\"https://matrix.to/#/@alice:matrix.org\" contenteditable=\"false\">Alice</a>!|</p>");
+    let mut state = MentionsState::default();
+    state.user_ids.insert("@alice:matrix.org".into());
+    assert_eq!(model.get_mentions_state(), state)
+}
+
+#[test]
+fn get_mentions_state_for_multiple_user_mentions() {
+    let model = cm("<p>hello <a data-mention-type=\"user\" href=\"https://matrix.to/#/@alice:matrix.org\" contenteditable=\"false\">Alice</a> and <a data-mention-type=\"user\" href=\"https://matrix.to/#/@bob:matrix.org\" contenteditable=\"false\">Bob</a>!|</p>");
```
jonnyandrew, line 613: We shouldn't need to include `data-mention-type` and `contenteditable` in the input because the model should parse mentions automatically. Please can you try removing?

crates/wysiwyg/src/tests/test_mentions.rs
```diff
@@ -594,6 +594,73 @@ fn can_insert_at_room_mention() {
...
+    let model = cm("<p>hello @room|");
+    let state = MentionsState {
+        user_ids: Default::default(),
+        has_at_room_mention: true,
+    };
+    assert_eq!(model.get_mentions_state(), state)
+}
+
+#[test]
+fn get_mentions_state_for_multiple_user_and_at_room_mentions() {
```
jonnyandrew, line 641: Should we add one test for content containing multiple `@room` mentions and duplicate user mentions?

jonnyandrew (approved):
Looks good to me! Just a few comments

crates/wysiwyg/src/composer_model/mentions.rs
Velin92: Adding a way to also include `room_ids` and `room_aliases`

PR #854, opened 2023-10-23
[Android] Add support for mentions to the Compose library

jmartinesp:
Expose `replaceSuggestion` and <del>`setLinkSuggestion`</del> `insertMentionAtSuggestion` in `RichTextEditorState`. Also add suggestions UI to the compose example.

Also added `MentionDetector` to replace the client-side parsing of mentions with something from the RTE library.


jonnyandrew (commented):
Looks good! Just one comment about a potential performance issue with the composable

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditorDefaults.kt
```diff
@@ -117,7 +117,7 @@ object RichTextEditorDefaults {
      * @param backgroundColor The background color to apply
      */
     fun pillStyle(
-        backgroundColor: Color = Color.Transparent,
+        backgroundColor: Color = Color.LightGray,
```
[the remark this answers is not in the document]
jmartinesp: Oh true, sorry. I thought this was only used in the examples.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -131,8 +139,11 @@ private fun RealEditor(
 
         view
     }, update = { view ->
-        view.setStyleConfig(style.toStyleConfig(view.context))
-        view.applyStyle(style)
+        if (style != previousStyle) {
+            view.setStyleConfig(style.toStyleConfig(view.context))
```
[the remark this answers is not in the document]
jmartinesp: I wrapped the `mentionDisplayHandler` parameter in a lambda to make it stable and prevent recompositions, but there's still some weird issue with `style` and `textStyleTypeface`, where they won't cause recompositions when the code is build from a clean state, but if *anything* in the code changes and we re-build, the `update` block will recompose with *every* change in the composer.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditorDefaults.kt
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/c45252f7328b8ec5455e3a25f114a23e1e398158

platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
```diff
@@ -96,6 +115,30 @@ class MainActivity : ComponentActivity() {
                                 .fillMaxWidth()
                                 .padding(16.dp),
                         )
+
+                        Spacer(modifier = Modifier.weight(1f))
```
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/c5ddd527b63521ed74433867716df88323801116

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jmartinesp: @jonnyandrew could you double check if you can also reproduce the issue above? If not, it might just be an issue on my side and we can assume this is fixed.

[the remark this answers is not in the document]
jonnyandrew: Very odd - I can reproduce this unfortunately.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -40,7 +42,7 @@ fun RichTextEditor(
     state: RichTextEditorState = rememberRichTextEditorState(),
     registerStateUpdates: Boolean = true,
     style: RichTextEditorStyle = RichTextEditorDefaults.style(),
-    mentionDisplayHandler: MentionDisplayHandler? = null,
+    mentionDisplayHandler: (() -> MentionDisplayHandler)? = null,
```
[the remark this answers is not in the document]
jmartinesp: I don't think there is. If you feel like it's easier to just use a lambda I can change it.

platforms/android/library/src/main/java/io/element/android/wysiwyg/display/MentionDisplayHandler.kt
```diff
@@ -14,4 +14,9 @@ interface MentionDisplayHandler {
      * Return the method with which to display an at-room mention
      */
     fun resolveAtRoomMentionDisplay(): TextDisplay
+
+    /**
+     * Return true if the given URL is a mention one
+     */
+    fun isMention(url: String): Boolean
```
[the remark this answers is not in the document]
jmartinesp: @ganfra asked if we could make the RTE take care of detecting mentions instead of relying in the client's implementation. But you're probably right, the RTE library itself should be agnostic about this and don't rely on the Matrix mentions at all. 

Should we publish a separate library to handle this?

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jonnyandrew: Ok, then let's go with `(Mention) -> TextDisplay` as the type for this as it's a bit more idiomatic

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -335,11 +335,11 @@ internal class HtmlToSpansParser(
...
         }
-        // TODO: use data-mention-type instead
-        //  https://github.com/matrix-org/matrix-rich-text-editor/issues/709
-        val isMention = !last.span.contentEditable
+
+        val handler = mentionDisplayHandler()
+        val isMention = handler?.isMention(url) ?: false
         val textDisplay = if (isMention) {
-            mentionDisplayHandler()?.resolveMentionDisplay(innerText, url) ?: TextDisplay.Plain
+            handler?.resolveMentionDisplay(innerText, url) ?: TextDisplay.Plain
```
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/ce455e76045cef08db2c13ccc4e3486b315af47e

crates/wysiwyg/src/lib.rs
```diff
@@ -62,3 +62,7 @@ pub use crate::suggestion_pattern::SuggestionPattern;
 pub use crate::text_update::ReplaceAll;
 pub use crate::text_update::Selection;
 pub use crate::text_update::TextUpdate;
+
+pub fn is_user_mention<S: UnicodeString>(url: &S) -> bool {
+    matrix_mentions::Mention::from_uri(&url.to_string()).is_some()
```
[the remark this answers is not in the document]
jmartinesp: Changed it to `is_mention`.

crates/matrix_mentions/src/mention.rs
```diff
@@ -175,7 +175,7 @@ impl Mention {
 /// 3 - As a custom uri
 ///  
 /// If any of the above succeed, return Some<MatrixIdI. Else return None.
-fn parse_matrix_id(uri: &str) -> Option<MatrixId> {
+pub fn parse_matrix_id(uri: &str) -> Option<MatrixId> {
```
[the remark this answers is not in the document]
jmartinesp: Fixed in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/ce455e76045cef08db2c13ccc4e3486b315af47e

bindings/wysiwyg-ffi/src/ffi_mention_detector.rs
```diff
@@ -0,0 +1,19 @@
...
+    pub fn new() -> Self {
+        Self {}
+    }
+}
+
+#[uniffi::export]
+impl MentionDetector {
+    pub fn is_user_mention(self: &Arc<Self>, url: String) -> bool {
+        let url = Utf16String::from_str(&url);
+        wysiwyg::is_user_mention(&url)
```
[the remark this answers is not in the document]
jmartinesp: Fixed in https://github.com/matrix-org/matrix-rich-text-editor/commit/ce455e76045cef08db2c13ccc4e3486b315af47e

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jmartinesp: Then I'm not sure what we should do with this. It looks like some kind of Compose Compiler bug? Should we re-use the workaround of `previousStyle` to prevent unnecessary re-renderings of the HTML?

platforms/android/library/src/main/java/io/element/android/wysiwyg/display/MentionDisplayHandler.kt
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/ce455e76045cef08db2c13ccc4e3486b315af47e

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jmartinesp: Done in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/ce455e76045cef08db2c13ccc4e3486b315af47e

jmartinesp:
The UI tests do work locally, btw. It seems like the project has had issues running these tests for a while, from what I've seen in the results of run actions for other PRs.

platforms/android/example-view/src/main/java/io/element/android/wysiwyg/poc/matrix/MatrixMentionMentionDisplayHandler.kt
```diff
@@ -2,15 +2,18 @@ package io.element.android.wysiwyg.poc.matrix
...
 import io.element.android.wysiwyg.display.MentionDisplayHandler
+import uniffi.wysiwyg_composer.MentionDetector
 
 /**
  * Convenience implementation of a [MentionDisplayHandler] that detects Matrix mentions and
  * displays them as default pills.
  */
-class MatrixMentionMentionDisplayHandler : MentionDisplayHandler {
+class MatrixMentionMentionDisplayHandler(
+    private val mentionDetector: MentionDetector?
```
jonnyandrew: Same as above, this can be removed now.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
```diff
@@ -24,21 +27,31 @@ import io.element.android.wysiwyg.display.MentionDisplayHandler
 fun EditorStyledText(
     text: CharSequence,
     modifier: Modifier = Modifier,
-    mentionDisplayHandler: MentionDisplayHandler? = null,
+    resolveMentionDisplay: ((text: String, url: String) -> TextDisplay)? = null,
+    resolveRoomMentionDisplay: (() -> TextDisplay)? = null,
```
jonnyandrew, line 36: Please could you add these params to the docstring?

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -40,15 +43,24 @@ fun RichTextEditor(
     state: RichTextEditorState = rememberRichTextEditorState(),
     registerStateUpdates: Boolean = true,
     style: RichTextEditorStyle = RichTextEditorDefaults.style(),
-    mentionDisplayHandler: MentionDisplayHandler? = null,
+    resolveMentionDisplay: ((text: String, url: String) -> TextDisplay)? = null,
+    resolveRoomMentionDisplay: (() -> TextDisplay)? = null,
```
jonnyandrew, line 50: Please could you add these to the docstring?

jonnyandrew, line 50: nit: The interface might be a bit clearer and more self documenting with a default implementation e.g.

```kotlin
(text: String, url: String) -> TextDisplay = RichTextEditorDefaults.MentionDisplay
```

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditorState.kt
```diff
@@ -203,6 +203,14 @@ class RichTextEditorState(
         this.hasFocus = hasFocus
     }
 
+    suspend fun replaceSuggestion(text: String) {
+        _viewActions.emit(ViewAction.ReplaceSuggestionText(text))
+    }
+
+    suspend fun insertMentionAtSuggestion(text: String, link: String) {
+        _viewActions.emit(ViewAction.InsertMentionAtSuggestion(text, link))
+    }
```
jonnyandrew, line 212: Please add documentation to these new functions.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -43,35 +43,66 @@ class EditorEditText : AppCompatEditText {
 
     private var inputConnection: InterceptInputConnection? = null
 
-    private lateinit var styleConfig: StyleConfig
+    var styleConfig: StyleConfig? = null
+        set(value) {
+            field = value
+
+            htmlConverter = value?.let { createHtmlConverter(it) }
```
jonnyandrew, line 50: It's a public API so it should behave properly whatever value is supplied but at the moment setting to `null` doesn't seem to be supported.

Can we avoid making `StyleConfig` nullable? If it's absolutely necessary to make it nullable, perhaps we can handle a `null` value by applying the default style?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -43,35 +43,66 @@ class EditorEditText : AppCompatEditText {
 
     private var inputConnection: InterceptInputConnection? = null
 
-    private lateinit var styleConfig: StyleConfig
+    var styleConfig: StyleConfig? = null
```
jonnyandrew, line 46: If this must become a public API, can we add documentation?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -13,23 +13,43 @@ import io.element.android.wysiwyg.view.StyleConfig
...
  */
 open class EditorStyledTextView : AppCompatTextView {
 
+    private var mentionDetector: MentionDetector? = null
+
     private lateinit var inlineCodeBgHelper: SpanBackgroundHelper
     private lateinit var codeBlockBgHelper: SpanBackgroundHelper
-    private lateinit var styleConfig: StyleConfig
+
+    var styleConfig: StyleConfig? = null
```
jonnyandrew, line 29: Same comment as above: [please can we either remove the nullability or handle it properly?

jonnyandrew, line 29: Now that this is a public API, please can you add docs?

platforms/android/library/src/main/java/io/element/android/wysiwyg/internal/viewmodel/EditorViewModel.kt
```diff
@@ -256,6 +256,6 @@ internal class EditorViewModel(
     }
 
     private fun stringToSpans(string: String): CharSequence =
-        htmlConverter.fromHtmlToSpans(string)
+        htmlConverter?.fromHtmlToSpans(string) ?: ""
```
jonnyandrew, line 259: nit: Should we log an error if htmlConverter is null here?

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -156,8 +157,14 @@ internal class HtmlToSpansParser(
...
-                val contentEditable = attrs?.getValue("contenteditable")?.toBoolean()
-                handleHyperlinkStart(url, contentEditable ?: true)
+                val contentEditable = attrs.getValue("contenteditable")?.toBoolean()
+                val data = buildMap<String, String> {
+                    for (i in 0..attrs.length) {
+                        val key = attrs.getLocalName(i) ?: continue
+                        set(key, attrs.getValue(i))
+                    }
+                }
+                handleHyperlinkStart(url, contentEditable ?: true, data)
```
jonnyandrew, line 167: `contentEditable` could be removed now

jonnyandrew:
> The UI tests do work locally, btw. It seems like the project has had issues running these tests for a while, from what I've seen in the results of run actions for other PRs.

Yeah indeed, I think there must be a caching issue because the tests seem to pass fairly reliably on the second CI run.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jonnyandrew: If I understood correctly, the issue shouldn't affect the production library because it's a clean build, so a workaround shouldn't be necessary. But DevX on the other hand, becomes a nightmare and risky if we cannot check the behaviour is correct without doing a clean build.

I can see two paths forward:
- isolate the the change that causes this then at least we'd be able to exclude it manually during development by commenting out code. Or,
- add a UI test that can ensure the `update` block is not called more often than it should (something we should probably do at some point regardless).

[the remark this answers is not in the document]
jmartinesp: Sorry, I forgot to mention this: https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/817e5bbf03d40da873794fd98c7b206ee68453dd actually solves this issue by using:

```kotlin
update = remember(capturedArgs...) { 
	{
		// Actual update function
	}
}
```

I added the `previousStyle != style` workaround for `EditorStyledText` to prevent the `HtmlConverter` from being re-created for each `text` change though.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -146,21 +161,14 @@ private fun RealEditor(
...
-        view.mentionDisplayHandler = object : MentionDisplayHandler {
-            override fun resolveMentionDisplay(text: String, url: String): TextDisplay {
-                return resolveMentionDisplay?.invoke(text, url) ?: TextDisplay.Plain
-            }
-
-            override fun resolveAtRoomMentionDisplay(): TextDisplay {
-                return resolveRoomMentionDisplay?.invoke() ?: TextDisplay.Plain
-            }
-
+    }, update = remember(style, typeface, mentionDisplayHandler) {
```
jonnyandrew, line 164: This workaround seems to work, thank you! Please could you add some comments to explain the problem and why this is needed? I've created an issue to link to as well https://github.com/matrix-org/matrix-rich-text-editor/issues/856. Please feel free to add more detail that could help us in future.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
```diff
@@ -18,27 +24,47 @@ import io.element.android.wysiwyg.display.MentionDisplayHandler
...
-            view.text = text
-        } else {
-            view.setHtml(text.toString())
+    }, update = remember(style != previousStyle, typeface, mentionDisplayHandler, text) {
+        { view ->
+            if (style != previousStyle) {
+                view.styleConfig = style.toStyleConfig(view.context)
+                view.applyStyleInCompose(style)
+                previousStyle = style
+            }
```
jonnyandrew, line 60: Same as above, thanks for the workaround! Please could you add some comments explaining the problem and why this is needed, linking to the GH issue?

jonnyandrew, line 60: I'm not sure about the implications of setting a variable from within the `update` block. Is it strictly necessary to prevent the style getting updated more than necessary given the `text` is unlikely to be getting updated frequently like it is with the editor itself?

platforms/android/library/src/main/java/io/element/android/wysiwyg/internal/viewmodel/EditorViewModel.kt
jmartinesp: Log or actually throw? The only reason why `HtmlConverter` could be null here is if there's been some configuration issue, so it seems reasonable to me to crash, but maybe we don't want to be so strict?

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
jmartinesp: I'm removing this and trying to use a better API.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: I'm reverting it to a private API.

jmartinesp:
All mentioned issues should have been fixed.

platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/DefaultMentionDisplayHandler.kt
```diff
@@ -2,8 +2,10 @@ package io.element.wysiwyg.compose
 
 import io.element.android.wysiwyg.display.MentionDisplayHandler
 import io.element.android.wysiwyg.display.TextDisplay
+import uniffi.wysiwyg_composer.MentionDetector
```
jonnyandrew, line 5: This import is unused

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -59,84 +74,109 @@ private fun RealEditor(
...
-        view.typeface = textStyleTypeface
-        view.rustErrorCollector = RustErrorCollector(onError)
-        view.mentionDisplayHandler = mentionDisplayHandler
-    })
+            view
+        },
+        // The `update` lambda is called when the view is first created, and then again whenever the actual `update` lambda changes. That is, it's replaced with
+        // a new lambda capturing different variables from the surrounding scope. However, there seems to be an issue that causes the `update` lambda to change
+        // more than it's strictly necessary. To avoid this, we can use a `remember` block to cache the `update` lambda, and only update it when needed.
+        update = remember(style, typeface, mentionDisplayHandler) {
```
jonnyandrew, line 170: `onError` is missing here

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/FakeViewConnection.kt
```diff
@@ -50,6 +50,8 @@ internal class FakeViewActionCollector(
             ViewAction.ToggleQuote -> toggleQuote()
             ViewAction.Undo -> undo()
             ViewAction.Unindent -> unindent()
+            is ViewAction.ReplaceSuggestionText -> Unit
+            is ViewAction.InsertMentionAtSuggestion -> Unit
```
jonnyandrew, line 54: Just FYI, leaving these no-ops will make it impossible to test this behaviour in host apps.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -232,6 +245,21 @@ class EditorEditText : AppCompatEditText {
         }
     }
 
+    /**
+     * Sets up the [HtmlConverter] used to translate HTML to Spanned text.
+     * @param styleConfig The styles to use for the generated spans.
+     * @param mentionDisplayHandler Used to decide how to display any mentions found in the HTML text.
+     */
+    fun setupHtmlConverter(styleConfig: StyleConfig, mentionDisplayHandler: MentionDisplayHandler?) {
```
jonnyandrew, line 253: Users of the View library may have set the style config through XML so it should be possible to set the mention display handler while respecting this. Can we make the `styleConfig` optional so that this use case is not broken?

jonnyandrew, line 253: The `HtmlConverter` is an implementation detail that could confuse users who are simply trying to change the appearance of the editor. Can we rename the public API to something a bit more friendly like `setStyle`?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -55,22 +65,29 @@ open class EditorStyledTextView : AppCompatTextView {
         codeBlockBgHelper.clearCachedPositions()
     }
 
+    /**
+     * Sets up the [HtmlConverter] used to translate HTML to Spanned text.
+     * @param styleConfig The styles to use for the generated spans.
+     * @param mentionDisplayHandler Used to decide how to display any mentions found in the HTML text.
+     */
+    fun setupHtmlConverter(styleConfig: StyleConfig, mentionDisplayHandler: MentionDisplayHandler?) {
```
jonnyandrew, line 73: Same comment as above, please can we rename the public API to something more friendly?

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -329,17 +335,16 @@ internal class HtmlToSpansParser(
         // If the link is not editable, tag all but the first character of the anchor text with
         // ExtraCharacterSpans. These characters will then be taken into account when translating
         // between editor and composer model indices (see [EditorIndexMapper]).
-        val isContentEditable = !last.span.contentEditable
+        val isContentEditable = last.span.data.containsKey("contenteditable")
```
jonnyandrew, line 338: The presence of `contenteditable` doesn't imply the value is true so I think we'd need to check the value of the attribute here. It may actually be possible to remove this and replace it with an `isMention` check, but I'm not 100% sure.

jonnyandrew (commented):
Thanks for the changes! I have a few more comments as there are quite a few more changes. I've made a few suggestions - Ii'd be great to address the public API issues to minimise breaking changes that we'd need to release in order to fix them.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jonnyandrew: (friendly reminder to please link to the issue in the code comments)

platforms/android/library/src/main/java/io/element/android/wysiwyg/internal/viewmodel/EditorViewModel.kt
jonnyandrew: Personally I'd prefer to er on the side of caution and not crash, in case we accidentally introduce a bug that triggers it in future.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/FakeViewConnection.kt
jmartinesp: I'm wondering what I should do here, I can replace the `state.menuAction` with `MenuAction.None`, thus 'removing' the suggestion? It's not like we can or should modify the underlying text here, right?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
jmartinesp: If you use the current implementation, it *should* just re-set the same `styleConfig` it previously had unless I messed it up. This was the only way I could think of for reducing the number of times we re-create the `HtmlConverter` in each `update` recomposition.

jmartinesp: Sounds good, `setStyle/updateStyle` works for me.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/FakeViewConnection.kt
jonnyandrew: The idea isn't that it works correctly, only that something in the state changes which is visible to the test. If you wanted to implement it, you could do something simple like search and replace the final occurrence of `@.*`.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
jonnyandrew: I'm not sure I understand - I was talking about apps that aren't using compose e.g. Element Android. The [removal](https://github.com/matrix-org/matrix-rich-text-editor/pull/854/files/18ed04108c0f31da539e07a686f0bf5de67eaf72#diff-0759d3c3c35a9122b7b81526317c7a65ba00cbc0dca69c3c0302c4e8c27d649dL99-L106) of the `var mentionDisplayHandler` means that they'll be unable to set this variable independently of the style (for which they may use XML).

> reducing the number of times we re-create the HtmlConverter in each `update` recomposition

Given that `update` only fires when one of the config variables (`style`, `typeface`, `onError` etc changes) it's unlikely to be called really frequently because of their nature. So IMO we don't need to over optimise it.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/FakeViewConnection.kt
jmartinesp: Ok, I'll do just that 👍 .

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
jmartinesp: > I'm not sure I understand - I was talking about apps that aren't using compose e.g. Element Android. The [removal](https://github.com/matrix-org/matrix-rich-text-editor/pull/854/files/18ed04108c0f31da539e07a686f0bf5de67eaf72#diff-0759d3c3c35a9122b7b81526317c7a65ba00cbc0dca69c3c0302c4e8c27d649dL99-L106) of the var mentionDisplayHandler means that they'll be unable to set this variable independently of the style (for which they may use XML).

Sorry, I think I wasn't clear enough. The `styleConfig` property's getter is now public, so they can just do:

```kotlin
editText.updateStyle(styleConfig = this.styleConfig, mentionDisplayHandler = mentionDisplayHandler)
```

It does feel a bit weird, but the alternative is having custom setters for the 2 properties, making `StyleConfig` nullable or checking with `this::styleConfig.isInitialized` and then creating a new `HtmlConverter` with the `styleConfig` already set and whatever `mentionDisplayHandler` is passed or was already there, making the logic harder to follow IMO.

> Given that update only fires when one of the config variables (style, typeface, onError etc) changes it's unlikely to be called really frequently because of their nature. So IMO we don't need to over optimise it.

It's true it won't be called that often, but it'll still create 2 `HtmlConverter` instances for each update, always discarding the first one, which is pointless.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jonnyandrew: Now that we have a bit more time to work with, is there any way we can eliminate this workaround?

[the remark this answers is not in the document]
jmartinesp: It depends on whether we want to avoid creating a new `HtmlConverter` every time anything changes in this component or not. The `remember` makes the update more predictable, but it will still be triggered when any of its params change.

The current `AndroidHtmlConverter` is quite light-weight since it only creates a lambda, so we could allow this, or even create the converter on demand whenever it's supposed to be used, but this might not be the case for future or alternative implementations.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
jonnyandrew: Ok, thanks for explaining, let's keep it this way.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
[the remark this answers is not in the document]
jmartinesp: Ok, I'll remove the `previousStyle` check, but the `remember` will have to stay for now, I don't really see what's causing the lambda to be re-created...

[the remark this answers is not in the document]
jmartinesp: Actually, `previousStyle` has already been removed 😅 . I completely forgot about that.

jonnyandrew (commented):
There are some mentions related tests failing when I run them locally - please can you fix those?

After that I think we are OK to merge this in its current state and unblock the work on mentions.

jmartinesp:
I think the last issues should have been fixed in https://github.com/matrix-org/matrix-rich-text-editor/pull/854/commits/ac1c05f5bd98b3c02ffce0449e0051fb66bdace1 (it seems like when I paused this work I forgot to migrate the mention detection logic). The tests work locally for me with those changes, but I can't make them work in the CI.

jmartinesp:
Thanks again for the latest changes, I was over complicating things quite a bit 😅 . Should this be ready to merge now?

jonnyandrew (approved):
Nice work, let's keep an eye out for any issues around https://github.com/matrix-org/matrix-rich-text-editor/issues/856

PR #857, opened 2023-10-27
[iOS] Add "shouldReplaceText" API 

alfogrillo:
Sometimes it's useful for the client to ignore the input from the keyboard while keeping the focus on the text view.


nimau (approved):
LGTM 👍

PR #858, opened 2023-10-30
Version 2.15.0

alfogrillo:


nimau (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+# [2.15.0] - 2023-10-30
+
+### Added
+
+- iOS: Add "shouldReplaceText" API to ignore keyboard input (#857)
+
```
jonnyandrew, line 8: There are quite a few more changes between 2.14.1 and 2.15.0 which have not been documented in the changelog. Please can you add these so that they are not lost?

alfogrillo: I see tags for `2.14.2`, `2.14.3`, `2.14.4` but apparently no related PRs nor change logs.
How is that possible?
This is supposed to be the only change after `2.14.4`

jonnyandrew: Some releases must have been triggered by manually pushing tags unfortunately. It seems the releases have only worked for iOS and not Android though.

Perhaps we there is a way to prevent the Github workflows running when this happens?

alfogrillo: > Some releases must have been triggered by manually pushing tags unfortunately

Yep, we should prevent this shortcut imho.

>  It seems the releases have only worked for iOS and not Android though.

Is it related with the Android tests failing?

alfogrillo: Added all the changes, [here](https://github.com/matrix-org/matrix-rich-text-editor/pull/858/commits/859830e4de9497bf4e04f2c70dbc267239a26ab6) under 2.15.0.
At this point it would be better to delete the tags without a change log later on.

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
...
+### Fixed
+- iOS: Use message content APIs (#842)
+- Android: Fix crash due to out of bounds error (#845)
+
+### Added
+- iOS: Add "shouldReplaceText" API to ignore keyboard input (#857)
+- iOS: Add getMentionsState API (#850)
+- Common: get_mentions_state() implemented on the Rust side (#849)
+- Android: Handle and track more internal exceptions (#846)
```
jonnyandrew, line 13: This is a bug fix

jonnyandrew (commented):
Thanks for adding these

CHANGELOG.md
jonnyandrew: Sounds good :+1:

jonnyandrew (approved):

PR #860, opened 2023-11-06
[Android] Make some spans clickable (pills, links) in TextView

jmartinesp:
- Add an associated `url` to `PillSpan` and `CustomReplacementSpan`.
- Rename `CustomReplacementSpan` to `CustomMentionSpan`.
- Override `EditorStyledTextView.onTouchEvent` to detect clicks in these clickable spans.
- Add `onLinkClickedListener` to get notified when they are clicked.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.graphics.Canvas
 import android.text.Spanned
 import android.util.AttributeSet
+import android.view.MotionEvent
 import androidx.appcompat.widget.AppCompatTextView
 import androidx.core.graphics.withTranslation
 import io.element.android.wysiwyg.display.MentionDisplayHandler
@@ -12,6 +13,9 @@ import io.element.android.wysiwyg.utils.HtmlConverter
 import io.element.android.wysiwyg.view.StyleConfig
 import io.element.android.wysiwyg.view.inlinebg.SpanBackgroundHelper
 import io.element.android.wysiwyg.view.inlinebg.SpanBackgroundHelperFactory
+import io.element.android.wysiwyg.view.spans.CustomMentionSpan
+import io.element.android.wysiwyg.view.spans.LinkSpan
+import io.element.android.wysiwyg.view.spans.PillSpan
 import io.element.android.wysiwyg.view.spans.ReuseSourceSpannableFactory
 import uniffi.wysiwyg_composer.MentionDetector
 import uniffi.wysiwyg_composer.newMentionDetector
@@ -38,6 +42,8 @@ open class EditorStyledTextView : AppCompatTextView {
 
     private var mentionDisplayHandler: MentionDisplayHandler? = null
     private var htmlConverter: HtmlConverter? = null
+
+    var onLinkClickedListener: ((String) -> Unit)? = null
 
     init {
         setSpannableFactory(spannableFactory)
@@ -127,5 +133,39 @@ open class EditorStyledTextView : AppCompatTextView {
                 }
             }
         )
+    }
+
+    override fun onTouchEvent(event: MotionEvent?): Boolean {
+        return when (event?.action) {
+            MotionEvent.ACTION_DOWN -> {
+                // Allow ACTION_UP to be called
+                super.onTouchEvent(event)
+                true
+            }
+            MotionEvent.ACTION_UP -> {
+                // Find selection matching the pointer coordinates
+                val offset = getOffsetForPosition(event.x, event.y)
+                // Look for clickable spans in that position
+                val spans = (text as? Spanned)?.getSpans(offset, offset, Any::class.java) ?: return false
+                val linkSpans = spans.filterIsInstance<LinkSpan>()
+                val pillSpans = spans.filterIsInstance<PillSpan>()
+                val customMentionSpans = spans.filterIsInstance<CustomMentionSpan>()
+                // Notify the link has been clicked
+                for (span in linkSpans) {
+                    if (span.url == null) continue
+                    onLinkClickedListener?.invoke(span.url)
+                }
+                for (span in pillSpans) {
+                    if (span.url == null) continue
+                    onLinkClickedListener?.invoke(span.url)
+                }
+                for (span in customMentionSpans) {
+                    if (span.url == null) continue
+                    onLinkClickedListener?.invoke(span.url)
+                }
+                true
+            }
+            else -> super.onTouchEvent(event)
+        }
     }
 }
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
@@ -4,10 +4,12 @@ import android.content.Context
 import android.graphics.Canvas
 import android.text.Spanned
 import android.util.AttributeSet
+import android.view.GestureDetector
 import android.view.MotionEvent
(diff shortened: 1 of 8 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -128,4 +134,38 @@ open class EditorStyledTextView : AppCompatTextView {
...
+            MotionEvent.ACTION_DOWN -> {
+                // Allow ACTION_UP to be called
+                super.onTouchEvent(event)
+                true
+            }
+            MotionEvent.ACTION_UP -> {
+                // Find selection matching the pointer coordinates
+                val offset = getOffsetForPosition(event.x, event.y)
+                // Look for clickable spans in that position
+                val spans = (text as? Spanned)?.getSpans(offset, offset, Any::class.java) ?: return false
```
jonnyandrew, line 149: Nice idea! A few thoughts which may or may not not be issues:
- Have you tested if there are any conflicts using `setMovementMethod(LinkMovementMethod)` after this change?
- What happens if an app returns a custom clickable span in the mention display handler?
- Could this trigger if a user scrolls by dragging a link/pill/mention?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -128,4 +134,38 @@ open class EditorStyledTextView : AppCompatTextView {
             }
         )
     }
+
+    override fun onTouchEvent(event: MotionEvent?): Boolean {
```
jonnyandrew, line 138: Please could you also add a UI test for the new behaviour?

jonnyandrew (commented):
Looks good! Just a question on the behaviour.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: > * Have you tested if there are any conflicts using `setMovementMethod(LinkMovementMethod)` after this change?

I think `LinkMovementMethod` needs the `TextView` to be selectable, which we don't allow at the moment, but I'd have to double check. Maybe instead of returning `true` in this case I should call the super function. Nice catch.

> * What happens if an app returns a custom clickable span in the mention display handler?

I think internally it would be represented by a `CustomMentionSpan`, which is a `ReplacementSpan` and can't contain a `ClickableSpan`. With `TextDisplay.Custom` you can just pass a `ReplacementSpan` if I'm not mistaken.

> * Could this trigger if a user scrolls by dragging a link/pill/mention?

Oh, nice catch. Maybe I should perform this same search in `ACTION_DOWN`, keep a reference to the selected spans and then check if I get the same ones in `ACTION_UP`.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: I'll try to, but it's not that simple to select a pill in the TextView. Maybe if I make the *only* content of the TextView it'll make things easier.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: I couldn't add any automated tests for the 1st question, but I can confirm it works. I also added some new tests and made sure scrolling won't trigger a click (which apparently does happen with the default `ClickableSpan`).

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -135,37 +137,64 @@ open class EditorStyledTextView : AppCompatTextView {
...
-                val spans = (text as? Spanned)?.getSpans(offset, offset, Any::class.java) ?: return false
-                val linkSpans = spans.filterIsInstance<LinkSpan>()
-                val pillSpans = spans.filterIsInstance<PillSpan>()
-                val customMentionSpans = spans.filterIsInstance<CustomMentionSpan>()
-                // Notify the link has been clicked
-                for (span in linkSpans) {
-                    if (span.url == null) continue
-                    onLinkClickedListener?.invoke(span.url)
+            MotionEvent.ACTION_MOVE -> {
+                val distance = initialTouchCoordinates?.let { hypot(it.first - event.x, it.second - event.y) } ?: 0f
```
jonnyandrew, line 158: We could skip a this calculation too by doing `if (!maybeClickEvent) return` here.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -135,37 +137,64 @@ open class EditorStyledTextView : AppCompatTextView {
...
+                // If the distance between the initial coordinates and the current ones is too big (4dp), we can assume it's not a click
```
jonnyandrew, line 159: I am worried that manual detection of scroll / click gestures could be a bit error prone and lead to something which is either too sensitive or not sensitive enough. I wonder if there is a way to avoid implementing this ourselves manually?

I saw [`GestureDetector`](https://developer.android.com/reference/android/view/GestureDetector) which looks like it does something similar - is it worth looking into?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jonnyandrew: No problem - thanks for adding these tests!

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: I guess we could, but I think `ACTION_MOVE` can only happen after `ACTION_DOWN`, so `maybeClickEvent` should always be true at that point. I can add it, just in case.

jonnyandrew: I thought `ACTION_MOVE` could be called many times during scrolling before `ACTION_UP` so thought of skipping any after `maybeClickEvent` gets reset [here](https://github.com/matrix-org/matrix-rich-text-editor/pull/860/files/78f8d42c9aedd1de5975306afee64016cc95ebab..931486ff3433dc40c423d4534d96dcf6cb29fcf9#diff-06656006549b36b19cc7b9afbd701cd4e71f2832c84971b16316f0641bbb486eR161). Maybe I missed something though?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: `SimpleGestureDetector` does just this and simplifies the code **a lot**. Thanks for the suggestion!

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
jmartinesp: No, you're right, I completely forgot it could be called several times. In any case, using the `GestureDetector` makes this code no longer needed, so I'll remove it.

jonnyandrew (approved):
Nice improvement, thank you!

jmartinesp:
Oops, I merged this before the CI passed for a couple of jobs. I guess I'm too used to the 'auto merge when all required jobs pass' feature on the EXA repo 😅 .

PR #863, opened 2023-11-14
[Android] Add special treatment for indexes of replacement spans

jmartinesp:
This workarounds an issue with indexes around mentions, which placed the cursor inside the `ReplacementSpan` when backspacing a mention. I tried modifying several other index mapping functions, but whatever made mentions work well broke indexes in lists or multiple paragraphs.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/EditorIndexMapper.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/EditorIndexMapper.kt
@@ -2,6 +2,7 @@ package io.element.android.wysiwyg.utils
 
 import android.text.Editable
 import android.text.Spanned
+import android.text.style.ReplacementSpan
 import androidx.core.text.getSpans
 import io.element.android.wysiwyg.view.spans.ExtraCharacterSpan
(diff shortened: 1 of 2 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/EditorIndexMapper.kt
```diff
@@ -66,12 +67,17 @@ object EditorIndexMapper {
...
+                val replacementSpans = editableText.getSpans<ReplacementSpan>(start = i, end = i + 1)
+                // We need to skip 1 character for every replacement span found
+                if (replacementSpans.isEmpty()) {
+                    consumed++
+                }
                 i++
             } else {
-                i += spans.count()
+                i += extraCharSpans.count()
             }
```
jonnyandrew, line 81: nit: The logic is getting a bit hard to follow. Would it help to refactor the nested `if` out i.e.

```kotlin
while(...) {
  if(skipCondition) {
    consumed++
  }
  i += something
}

jonnyandrew (approved):

jonnyandrew (commented):
@jmartinesp it'd be great to add a corresponding test for this change also

PR #864, opened 2023-11-14
[Android] Add workarounds for ignoring state restoration and invalid line count

jmartinesp:
I found 2 more bugs in the interop between Compose and the AndroidView:

1. ~~When restoring state in `RealEditor`'s `AndroidView`, for some reason if there is a single NSBP char at the end of the HTML it'll just remove it, and it can cause crashes because the selection may become invalid.~~ - this was [fixed in Rust](https://github.com/matrix-org/matrix-rich-text-editor/pull/865). In any case, I believe this state restoration is not needed because the state is kept at the `ComposerModel` and the text and selection are restored in the `EditorEditText.onRestoreInstanceState` method. This should fix the crashes we're seeing in EXA when a mention is added.
2. Related to the previous issue, the `factory` contents run before this `onRestoreInstanceState` runs, so there's no way to know if we're dealing with a restoration, AFAICT, so by the time we're adding the text changed listener on the factory the state hasn't been restored and the `lineCount` will always be 0, which in turn will trigger a recomposition of the whole component, and then another one once we have the right `lineCount = 1` value. I changed the code to only take into account `lineCount` if it's > 0, which should fit our usage since we don't care about value 0 (the edittext being empty).

### DIFF
--- a/platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
+++ b/platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
@@ -15,6 +15,7 @@ import androidx.test.espresso.matcher.ViewMatchers.withText
 import io.element.android.wysiwyg.utils.NBSP
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.test.runTest
+import org.junit.Assert
 import org.junit.Rule
 import org.junit.Test
 
@@ -52,6 +53,37 @@ class RichTextEditorStateTest {
 
         composeTestRule.onNodeWithText("Alternative editor").assertIsDisplayed()
         onView(withText("Hello, world")).check(matches(isDisplayed()))
+    }
+
+    @Test
+    fun testStateRestoration() = runTest {
+        val state = RichTextEditorState()
+        val hideEditor = MutableStateFlow(false)
+        composeTestRule.setContent {
+            MaterialTheme {
+                val hide by hideEditor.collectAsState()
+                Column {
+                    if (!hide) {
+                        Text("Editor")
+                        RichTextEditor(state = state)
+                    }
+                }
+            }
+        }
+
+        state.setHtml("Hello, world")
+        // Ensure line count is set
+        Assert.assertEquals(1, state.lineCount)
+
+        // Hide and show the editor to simulate a configuration change
+        hideEditor.emit(true)
+        hideEditor.emit(false)
+        composeTestRule.awaitIdle()
+
+        // If the text is found, the state was restored
+        onView(withText("Hello, world")).check(matches(isDisplayed()))
+        // Line count is kept
+        Assert.assertEquals(1, state.lineCount)
     }
 
     @Test
--- a/platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
+++ b/platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
@@ -35,10 +35,10 @@ class RichTextEditorStateTest {
                 Column {
                     if (!showAlt) {
                         Text("Main editor")
-                        RichTextEditor(state = state)
+                        RichTextEditor(state = state, modifier = Modifier.fillMaxWidth())
                     } else {
                         Text("Alternative editor")
-                        RichTextEditor(state = state)
+                        RichTextEditor(state = state, modifier = Modifier.fillMaxWidth())
                     }
                 }
             }
@@ -64,7 +64,7 @@ class RichTextEditorStateTest {
         composeTestRule.setContent {
             MaterialTheme {
                 val hide by hideEditor.collectAsState()
-                Column() {
+                Column {
                     if (!hide) {
                         Text("Editor")
                         RichTextEditor(modifier = Modifier.fillMaxWidth(), state = state)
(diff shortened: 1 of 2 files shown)


jmartinesp:
Maestro passed [here](https://github.com/vector-im/element-x-android/actions/runs/6875187159/job/18698285975), so I'm merging this.

jmartinesp:
Can I get another review for this PR? I added tests for checking the text + selection restoration, as well as keeping the right line count. I also tested it on EXA and it seems to work fine.

platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
```diff
@@ -54,6 +57,37 @@ class RichTextEditorStateTest {
...
+        }
+
+        state.setHtml("Hello, world")
+        // Ensure line count is set
+        Assert.assertEquals(1, state.lineCount)
+
+        // Hide and show the editor to simulate a configuration change
+        hideEditor.emit(true)
+        hideEditor.emit(false)
+        composeTestRule.awaitIdle()
```
jonnyandrew, line 83: Does this not need an extra `composeTestRule.awaitIdle()` between hiding and showing?

platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
```diff
@@ -54,6 +57,37 @@ class RichTextEditorStateTest {
...
+
+        // If the text is found, the state was restored
+        onView(withText("Hello, world")).check(matches(isDisplayed()))
+        // Line count is kept
+        Assert.assertEquals(1, state.lineCount)
```
jonnyandrew, line 88: There is something a bit odd going on with the line count now; adding this assertion to the test above (`testSharingState`) causes it to fail with `state.lineCount == 12`. Shall we add add the line count check to that test and if possible get it returning the correct result?

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -135,9 +138,11 @@ private fun RealEditor(
...
                 applyDefaultStyle()
 
-                // Restore the state of the view with the saved state
-                setHtml(state.internalHtml)
-                setSelection(state.selection.first, state.selection.second)
+                // Set initial HTML and selection, state will be restored automatically in the view
+                if (editableText.isEmpty()) {
+                    setHtml(state.internalHtml)
+                    setSelection(state.selection.first, state.selection.second)
+                }
```
jonnyandrew, line 145: In EXA, the selection state is no longer restored when toggling the text formatting options:

[Screen_recording_20231115_162355.webm](https://github.com/matrix-org/matrix-rich-text-editor/assets/4940864/4a8a8528-7292-4773-be8b-dbd56caedb87)

jonnyandrew, line 145: Is it still necessary to make this change after https://github.com/matrix-org/matrix-rich-text-editor/pull/865 ?

platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
jmartinesp: This is because the `EditText` view is way too small and it ends up wrapping the text, I added a `Modifier.fillMaxWidth()` to the other test to fix this same issue, I can add it there too.

platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
jmartinesp: I just added it and it makes no difference so I can leave it there if it helps ensure we're actually changing the state.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jmartinesp: You're right. I'm reverting this code! It's weird, because it contradicts some of my previous tests while debugging, but maybe fixing the previous bug also changed that behaviour.

jonnyandrew: I've added an extra test for this case to help catch this in future: https://github.com/matrix-org/matrix-rich-text-editor/pull/868

platforms/android/library-compose/src/androidTest/java/io/element/android/wysiwyg/compose/RichTextEditorStateTest.kt
jonnyandrew: Oh of course, that makes sense. Thank you!

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jmartinesp: Oh cool, I was adding something similar but your tests are more thorough so let's use them instead 👍 .

jmartinesp: Nope, removing this now and uploading some new changes.

jmartinesp: Actually, I accidentally uploaded some changes, but those also verify line count is not 1 by default, so maybe we can change the #868 PR to be based on these?

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -135,7 +138,7 @@ private fun RealEditor(
 
                 applyDefaultStyle()
 
-                // Restore the state of the view with the saved state
+                // Set initial HTML and selection based on the provided state
```
jonnyandrew, line 141: This is less confusing, thanks!

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jonnyandrew: Sure that makes sense - I can update #868 to use two lines of text.

jonnyandrew (approved):

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
jonnyandrew: Sure that makes sense - I can update https://github.com/matrix-org/matrix-rich-text-editor/pull/868 to use two lines of text.

PR #865, opened 2023-11-14
Don't trim NBSP chars at the end of a container

jmartinesp:
This was causing crashes with state restoration on Android, since something like this:

``` html
<b>test</b>&nbsp;|
```

When restored, would look like:
``` html
<b>test</b>
``` 

But the selection would be kept at the same index, so when trying to apply it to the underlying UI the apps would crash.

### DIFF
--- a/crates/wysiwyg/src/dom/parser/parse.rs
+++ b/crates/wysiwyg/src/dom/parser/parse.rs
@@ -125,7 +125,8 @@ mod sys {
                         // Special case for code block, translate '\n' into <br /> nodes
                         let is_inside_code_block =
                             self.current_path.contains(&CodeBlock);
-                        convert_text(&text.content, node, is_inside_code_block);
+                        let is_only_child_in_parent = node.children().len() == 0;
+                        convert_text(&text.content, node, is_inside_code_block, is_only_child_in_parent);
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


crates/wysiwyg/src/dom/parser/parse.rs
```diff
@@ -125,7 +125,14 @@ mod sys {
                         // Special case for code block, translate '\n' into <br /> nodes
                         let is_inside_code_block =
                             self.current_path.contains(&CodeBlock);
-                        convert_text(&text.content, node, is_inside_code_block);
+                        let is_only_child_in_parent =
+                            node.children().is_empty();
```
jonnyandrew, line 129: I'm confused at how `is_only_child_in_parent` is being calculated from `node.children().is_empty()`. Should we not look at the size of `panode.children`?

jonnyandrew (commented):
Looks good, and thanks for the test!

jonnyandrew (approved):

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditorState.kt
```diff
@@ -142,6 +142,15 @@ class RichTextEditorState(
...
     }
 
+    /**
+     * Sets the text selection to the provided indexes.
+     * @param start The start index of the selection
+     * @param end The end index of the selection. If not provided, the selection will be a cursor.
+     */
+    suspend fun setSelection(start: Int, end: Int = start) {
+        _viewActions.emit(ViewAction.SetSelection(start, end))
+    }
```
jonnyandrew, line 152: Please could you add a test for this?

Issue #856, opened 2023-10-25
Composable AndroidView `update` block is run on every key press


jonnyandrew:
## Problem

The `update` block of the composable `AndroidView` is run on every key press, whereas it should only run when a variable used inside the block is changed.

This block is supposed to run any time the style or callbacks are changed and isn't optimised to be run very frequently.

- Version 2.14.1 is not affected.
- Version 2.15.0 is affected.
- The main branch works around the issue to reduce performance impact (see https://github.com/matrix-org/matrix-rich-text-editor/pull/854#discussion_r1368618231) but the `AndroidView` still recomposed on every key press.

## Proposed solution

Workaround implemented in https://github.com/matrix-org/matrix-rich-text-editor/pull/854

No solution is known.

jmartinesp:
I think this is actually working as expected (although we didn't expect it to work like that 😅 ): looking at the [source code](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/viewinterop/AndroidView.android.kt;l=374?q=ViewFactoryHolder), we can see the `update` lambda will only be called when the lambda instance changes, and this can happen in 2 scenarios:
1. Any variable captured in the lambda changes.
2. A recomposition of the parent composable.

By using a `remember` block we can cache the lambda using some keys (in this case, the captured arguments) and reuse it even if the parent composable is recomposed, instead of getting a new lamda.

As for why there were different behaviours after a clean build and an incremental one, I have no idea. It might be a compiler bug.

jonnyandrew:
> the update lambda will be called on a recomposition of the parent composable.

I thought the `update` block should only run again when one of the variables inside it changes, not whenever the parent recomposes. I created [a small sample](https://gist.github.com/jonnyandrew/dfe54c48dac6fa067809937e8fbe6b47) to test it and it seems to be working in this way. Can you let me know if I misunderstood something here?

jmartinesp:
My theory was the `AndroidView` would be recomposed when anything captured in the `onFactory` lambda changed too, and this would in turn recompose the `update` lambda, but that wasn't the case after some quick testing, so it seems like I was wrong.

jonnyandrew:
With Android Studio Hedgehog we're able to debug the reasons for recomposition and it looks to be caused by the `onError` passed from the sample app:

[...]

PR #866, opened 2023-11-14
[Android] Remove workaround for excessive calls to `AndroidView` `update`

jonnyandrew:
## Problem

The `update` block of the composable `AndroidView` is run on every key press, whereas it should only run when a variable used inside the block is changed.

See https://github.com/matrix-org/matrix-rich-text-editor/issues/856 for more details.

## Solution

There is currently a workaround in place for this bug but after some investigation using the Compose inspection tools available in AS Hedgehog, I found that the recomposition is triggered by the value of `onError` being detected as changed despite it being a static function reference.

I don't know why this causes the recomposition but replacing the function reference with a lambda function fixes the problem.

- Fixes https://github.com/matrix-org/matrix-rich-text-editor/issues/856


jmartinesp (approved):
LGTM. I also changed some code so the `StyledHtmlConverter` is only configured when needed in the compose example, which should clear the logs a bit.

jonnyandrew:
Thanks for the improvement @jmartinesp!

PR #867, opened 2023-11-15
Handle external IDs for mentions that use client-style paths rather than matrix.to-style

andybalaam:


crates/matrix_mentions/src/mention.rs
```diff
@@ -215,6 +217,19 @@ fn parse_external_id(uri: &str) -> Result<MatrixToUri, IdParseError> {
     MatrixToUri::parse(&uri_for_ruma)
 }
 
+/// Attempt to find `/#/user/` or `/#/room/` in the supplied URI, and split it on one of those if
+/// found, meaning it is a URI into a client like Element Web. Otherwise split it on `/#/`,
+/// treating it as if it were a matrix.to URI.
+fn split_uri_on_prefix(uri: &str) -> Vec<&str> {
```
jonnyandrew, line 223: Since this function is only needed for custom Matrix URLs should it be annotated with the `feature` attribute?
```
#[cfg(any(test, feature = "custom-matrix-urls"))]
```

jonnyandrew (approved):
Looks good to me! Just a small question.

crates/matrix_mentions/src/mention.rs
andybalaam: Good spot, thank you. Fixed in 69823a65

PR #869, opened 2023-11-16
Version 2.17.0

jmartinesp:


jmartinesp:
The Android tests will most likely fail, since the emulators are becoming more and more flaky. The separate branches were tested and worked fine, so we can skip that check.

andybalaam (approved):
Looks good!

Issue #870, opened 2023-11-16
Android UI tests are flaky


jonnyandrew:
The Android UI tests fail randomly due to the following error:
```
androidx.test.espresso.base.RootViewPicker$RootViewWithoutFocusException: Waited for the root of the view hierarchy to have window focus and not request layout for 10 seconds. If you specified a non default root matcher, it may be picking a root that never takes focus
```

A screenshot taken from one such test run reveals the cause as an ANR dialog from the Quickstep app:

![20231116_152527](https://github.com/matrix-org/matrix-rich-text-editor/assets/4940864/4f509236-60a8-4449-aad6-7dd93b48b7ad)

PR #871, opened 2023-11-16
[Android] Take a screenshot when instrumentation tests fail

jonnyandrew:
## Changes

Take a screenshot when the core library UI tests fail.

An example of the output can be found in the [CI artifacts](https://github.com/matrix-org/matrix-rich-text-editor/actions/runs/6893260877) for https://github.com/matrix-org/matrix-rich-text-editor/pull/871/commits/04b84f5c990f41969fd7ee07b15c11d92e345c00.

## Context

To aid debugging of test failures.

- #870

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -92,10 +92,8 @@ jobs:
           disable-animations: true
           enable-hw-keyboard: true
           script: |
-            ./gradlew unitTestsWithCoverage ${{ env.CI_GRADLE_ARG_PROPERTIES }}
-            ./gradlew generateUnitTestCoverageReport ${{ env.CI_GRADLE_ARG_PROPERTIES }}
-            ./gradlew instrumentationTestsWithCoverage ${{ env.CI_GRADLE_ARG_PROPERTIES }}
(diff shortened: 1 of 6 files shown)


jmartinesp:
Nice, it seems like we can get the cause of the issue now! Maybe we can automate closing that dialog? I'm pretty sure I saw something similar in some issue a long time ago.

jmartinesp:
Maybe [this](https://github.com/ReactiveCircus/android-emulator-runner/issues/129#issuecomment-1024989188)? They also recommend using non-google-api versions if possible.

jonnyandrew:
> Maybe [this](https://github.com/ReactiveCircus/android-emulator-runner/issues/129#issuecomment-1024989188)? They also recommend using non-google-api versions if possible.

Yeah this would be handy but the API is deprecated from API 31 so I don't think we can use that.

I was going to try to see if we can use UiAutomator to dismiss the dialog because I see a few other projects doing this.

jonnyandrew:
@jmartinesp am I OK to merge this change so that it's not blocked by the eventual fix?

jmartinesp:
Oh, sure. I thought you wanted to apply the fix in this PR, that's why I hadn't approved yet.

jmartinesp (approved):

PR #875, opened 2023-11-17
[iOS] Fix for dictation not working properly

Velin92:
We found a way to intercept dictation, which would not work if the content is reconciliated while the dictation is still in progress, so we delay the `reconciliateIfNeeded()` function until the dictation state is false.

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
@@ -72,8 +72,12 @@ public class WysiwygTextView: UITextView {
     }
     
     @objc private func textInputCurrentInputModeDidChange(notification: Notification) {
-        guard isFirstResponder,
-              let inputMode = textInputMode?.primaryLanguage,
+        // We don't care about the input mode if this is not the first responder
+        guard isFirstResponder else {
+            return
+        }
+        
+        guard let inputMode = textInputMode?.primaryLanguage,
               inputMode == "dictation" else {
             isDictationRunning = false
             return
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
@@ -44,10 +44,7 @@ protocol WysiwygTextViewDelegate: AnyObject {
 public protocol MentionDisplayHelper { }
 
 public class WysiwygTextView: UITextView {
-    private var inputMode: String?
-    var isDictationRunning: Bool {
-        inputMode == "dictation"
-    }
+    private(set) var isDictationRunning = false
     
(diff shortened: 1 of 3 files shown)


Velin92:
@stefanceriu <3

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
```diff
@@ -172,6 +202,10 @@ public class WysiwygTextView: UITextView {
 
         wysiwygDelegate.textView(self, didReceivePasteWith: provider)
     }
+    
+    deinit {
+        NotificationCenter.default.removeObserver(self)
```
stefanceriu, line 207: I don't think this is needed

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
```diff
@@ -44,6 +44,9 @@ protocol WysiwygTextViewDelegate: AnyObject {
 public protocol MentionDisplayHelper { }
 
 public class WysiwygTextView: UITextView {
+    private var lastInputMode: String?
+    private(set) var isDictationRunning = false
```
stefanceriu, line 48: Maybe it's better if this is property computed based on the `lastInputMode`?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
```diff
@@ -53,12 +56,39 @@ public class WysiwygTextView: UITextView {
...
     }
     
     required init?(coder: NSCoder) {
         super.init(coder: coder)
+        commonInit()
+    }
+    
+    private func commonInit() {
         contentMode = .redraw
+        NotificationCenter.default.addObserver(self,
```
stefanceriu, line 69: I wonder how this behaves when multiple textviews are present on screen. Should we maybe check for isFirstResponder?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
Velin92: it needs to be changed also according to two override methods for ending the recording and when the recording fails

stefanceriu: Maybe those should clean up the `lastInputMode` instead?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
Velin92: yeah the stack overflow solution used vert old code, now these are deallocated automatically

stefanceriu (approved):
lgtm 👍

PR #878, opened 2023-11-20
Fix disappearing numbers with some keyboards

jmartinesp:
Don't set composition if the replaced text is all digits. This seems to be the default behaviour in all keyboards I've tested (Gboard, AnySoftKeyboard, Simple Keyboard...), and prevents the composition (the last digit) from being replaced with a whitespace.

Fixes https://github.com/vector-im/element-x-android/issues/1415.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
@@ -9,6 +9,7 @@ import android.view.inputmethod.*
 import android.widget.TextView
 import androidx.annotation.RequiresApi
 import androidx.annotation.VisibleForTesting
+import androidx.core.text.isDigitsOnly
 import io.element.android.wysiwyg.internal.utils.TextRangeHelper
 import io.element.android.wysiwyg.internal.viewmodel.EditorInputAction
 import io.element.android.wysiwyg.internal.viewmodel.ReplaceTextResult
@@ -322,7 +323,11 @@ internal class InterceptInputConnection(
         beginBatchEdit()
         editable.removeFormattingSpans()
         editable.replace(0, editable.length, charSequence)
-        setComposingRegion(compositionStart, compositionEnd)
+        val start = compositionStart.coerceIn(0, editable.length)
+        val end = compositionEnd.coerceIn(0, editable.length)
+        if (!editable.substring(start, end).isDigitsOnly()) {
+            setComposingRegion(start, end)
+        }
         endBatchEdit()
     }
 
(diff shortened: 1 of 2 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -322,7 +323,12 @@ internal class InterceptInputConnection(
         beginBatchEdit()
         editable.removeFormattingSpans()
         editable.replace(0, editable.length, charSequence)
-        setComposingRegion(compositionStart, compositionEnd)
+        val start = compositionStart.coerceIn(0, editable.length)
+        val end = compositionEnd.coerceIn(0, editable.length)
+        val newComposition = editable.substring(start, end)
+        if (newComposition.isEmpty() || !newComposition.isDigitsOnly()) {
```
jonnyandrew, line 329: The case where `isEmpty() == true` is already covered by `!isDigitsOnly()` so it could be simplified. Did you mean `isBlank() || !isDigitsOnly()`?

jonnyandrew, line 329: It seems a bit odd that `replaceAll` will not honour the given `compositionStart` and `compositionEnd`. Maybe the caller could be responsible for passing the right arguments? Or we can move the responsibility of setting the new composition out of `replaceAll`?

jonnyandrew (approved):
Thanks for finding the cause! It was a bit unexpected that `replaceText` would also update the composing region. I wonder if there's a way to make this a bit less opaque. But feel free to ignore these suggestions.

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
jmartinesp: Actually, it's not. The tests were failing before I added this condition. `isDigitsOnly` will go through the different codepoints in the string and when one is not a digit return false, otherwise return true after checking all. Since there are no codepoints, it will return that all characters (0 in this case) are digits.

jonnyandrew: Ah OK, my mistake!

jmartinesp: No worries, I thought the same at first and it took me a while to figure out what was wrong 😅 . It's super counter-intuitive.

jmartinesp: I changed this on https://github.com/matrix-org/matrix-rich-text-editor/pull/878/commits/88c81f918230f1c18295795ba23b23bd3e56c898

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -280,7 +281,7 @@ internal class InterceptInputConnection(
                 processInput(EditorInputAction.BackPress)
             }
             if (result != null) {
-                replaceAll(result.text, 0, editable.length)
+                replaceAll(result.text, result.selection.first, result.selection.last)
```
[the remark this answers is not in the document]
jmartinesp: Oh, maybe they're not. I forgot these were *composer* indexes that need to be translated to editor ones. Damn, I'll create a new PR with a fix for that.

PR #881, opened 2023-11-20
[Android] Apply a better solution for #878

jmartinesp:
- Make sure we translate any index that comes from the `ComposerModel` into an editor one.
- Move setting the composing region out of `replaceAll`.
- Apply the `isDigitOnly` check at the `onHardwareCharacterKey` method, since it's where the logic differs from the IME and setting it in `setComposingText` instead causes issues with Gboard: if you type a word, add a space and start typing a number, both `setComposignText` and `commitText` will be called for it and it'll be added twice.


platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -235,7 +243,13 @@ internal class InterceptInputConnection(
         // append to the current composition with the hardware keyboard.
         finishComposingText()
 
-        return setComposingText(newChars, 1)
+        return if (newChars.isDigitsOnly()) {
```
jonnyandrew, line 246: This check has moved to `onHardwareCharacterKey()` - don't we need to handle the same case for software keyboards?

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -125,8 +129,12 @@ internal class InterceptInputConnection(
...
 
         return if (result != null) {
-            replaceAll(result.text, compositionStart = end, compositionEnd = end)
-            setSelectionOnEditable(editable, result.selection.last, result.selection.last)
+            beginBatchEdit()
+            replaceAll(result.text)
+            val editorSelectionIndex = editorIndex(result.selection.last, editable)
+            setSelection(editorSelectionIndex, editorSelectionIndex)
+            setComposingRegion(editorSelectionIndex, editorSelectionIndex)
+            endBatchEdit()
```
jonnyandrew, line 137: There are now a few blocks similar to this. Would it be possible/worthwhile to create a function so that we can write:

```kotlin
val result = processTextEntry(text, start, end)

handleEditorResult(result)
```

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -260,9 +274,10 @@ internal class InterceptInputConnection(
                 processInput(action)
             }
             if (result != null) {
-                replaceAll(result.text, result.selection.first, result.selection.last)
-                setSelectionOnEditable(editable, result.selection.first, result.selection.last)
-                setComposingRegion(result.selection.first, result.selection.last)
+                replaceAll(result.text)
+                val editorSelectionIndex = editorIndex(result.selection.first, editable)
+                setSelection(editorSelectionIndex, editorSelectionIndex)
+                setComposingRegion(editorSelectionIndex, editorSelectionIndex)
```
jonnyandrew, line 280: nit: I think it's reasonable to assume that `selection.first == selection.last` but it feels a bit strange not to trust the Rust model. Anyway, it's not a big deal but if we did manage to refactor the result handling into a function (as mentioned in my other comment), that might fix this.

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
jmartinesp: I don't think so: after taking another look at the debugger, it turns out the problematic keyboards (Simple Keyboard, AnySoftKeyboard) just report the input events for these digits as hardware key presses, while Gboard uses `commitText`.

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
jmartinesp: I'm not sure. The strategies on those different places are different: 

- `setComposingText` applies the selection with the `end` part of the resulting composer update selection, but sets the composition in a custom way.
- `commitText` also applies the selection with the `end` update selection, and the composition will match it, to allow starting a new one at that point.
- `deleteSurroundingText` applies the `start` update selection instead, using it also for the composition.

I don't see a good way to extract a reusable function here, since each action needs different strategies for updating the different text, selection and composition values.

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
jmartinesp: I mostly did it following what was already there and trying to avoid calculating indexes several times, I see no need to convert twice index `x` from the composer to get the same `y` editor index in both cases, doing extra work. Also, the composer doesn't have the concept of 'composition', so we some times may have to get a bit creative with selections.

In the future we could add this concept to the composer, since it seems to be used in both Android an iOS (on iOS I believe it's called 'marked text'), but right now that would be way too much work for the time we have to work on the library.

jonnyandrew (approved):

PR #882, opened 2023-11-20
[iOS] Fix the position of the caret

nimau:
This PR fixes https://github.com/vector-im/element-x-ios/issues/1674

The caret position was calculated using the glyph bounding rectangle, but this rectangle sometimes (almost always when the caret is set at the end of a wrapped line) coincided with the start of the next line. So the final caret position was at the correct x position, but at the wrong y position.

The fix is to let the UITextView calculate the correct caret rect and just update it's y and height.

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
@@ -150,16 +150,12 @@ public class WysiwygTextView: UITextView {
     override public func caretRect(for position: UITextPosition) -> CGRect {
         // Compute system expected caret rect.
         let rect = super.caretRect(for: position)
-        // Determine rect for glyph at expected position.
-        let index = offset(from: beginningOfDocument, to: position)
-        let glyphRange = layoutManager.glyphRange(forCharacterRange: .init(location: index, length: 1), actualCharacterRange: nil)
-        let glyphRect = layoutManager.boundingRect(forGlyphRange: glyphRange, in: textContainer)
         // Use the system caret rect for `x` position and width and correct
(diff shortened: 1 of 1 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygTextView.swift
```diff
@@ -150,16 +150,12 @@ public class WysiwygTextView: UITextView {
     override public func caretRect(for position: UITextPosition) -> CGRect {
         // Compute system expected caret rect.
         let rect = super.caretRect(for: position)
-        // Determine rect for glyph at expected position.
-        let index = offset(from: beginningOfDocument, to: position)
-        let glyphRange = layoutManager.glyphRange(forCharacterRange: .init(location: index, length: 1), actualCharacterRange: nil)
-        let glyphRect = layoutManager.boundingRect(forGlyphRange: glyphRange, in: textContainer)
         // Use the system caret rect for `x` position and width and correct
-        // the `y` position and the height using the text glyphs.
+        // the `y` position and the height.
```
alfogrillo, line 154: Question: what if delete the full override `caretRect(for:)` at this point?
Made a quick try and LGTM.
wdyt?

nimau: I think it's safe to keep it to increase the height of the caret.

alfogrillo (approved):
LGTM

ara4n:
when will this turn up in EXI? am still getting weird caret positioning in today's nightly:

image

nimau:
> when will this turn up in EXI? am still getting weird caret positioning in today's nightly:
> 
> image

I'll make a version 2.19.0 to include this fix in EX.

ara4n:
thanks!

PR #883, opened 2023-11-21
[Android] Fix using emojis inside composing regions

jmartinesp:
I found a crash when testing the RTE that happened when I added an emoji in the middle of a word, the composer suddenly got an invalid index in the middle of the emoji and it crashes.

It turns out it comes from `replaceAll` doing `editable.replace(0, editable.length, someOtherValue)`. To test this, I:

1. Wrote `Test` in the composer.
2. Set the cursor in the middle of the word (`Te|st`).
3. Added an emoji.
4. Realised that, before calling `editable.replace(0, editable.length, ...)` inside `replaceAll`, the selection is at 2, but after that it's automatically moved to 3 by no code of ours, something inside Android must be doing it.

The behavior of `replaceAll` was done like that to prevent the suggestion from flashing when typing, but I just now tried to do:

```kotlin
editable.clear()
editable.append(newValue)
```

And the issue is fixed and the suggestions still work as expected. Moreover, we're not sending invalid indexes anymore to the composer, which seems like an unexpected win.

I had to change the order of the expected results of `dumpSpans()` since the new spans are not added at the end now, but at position `1 `.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
@@ -333,7 +333,8 @@ internal class InterceptInputConnection(
 
     private fun replaceAll(charSequence: CharSequence) {
         editable.removeFormattingSpans()
-        editable.replace(0, editable.length, charSequence)
+        editable.clear()
+        editable.append(charSequence)
     }
 
     private fun editorIndex(composerIndex: Int, editable: Editable): Int {
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnectionIntegrationTest.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnectionIntegrationTest.kt
@@ -53,13 +53,13 @@ class InterceptInputConnectionIntegrationTest {
             textView.text.dumpSpans(), equalTo(
                 listOf(
                     "hello: android.widget.TextView.ChangeWatcher (0-5) fl=#6553618",
+                    "hello: android.text.style.StyleSpan (0-5) fl=#33",
                     "hello: android.text.method.TextKeyListener (0-5) fl=#18",
+                    "hello: android.text.style.UnderlineSpan (0-5) fl=#289",
+                    "hello: android.view.inputmethod.ComposingText (0-5) fl=#289",
                     "hello: android.widget.Editor.SpanController (0-5) fl=#18",
                     ": android.text.Selection.START (5-5) fl=#546",
                     ": android.text.Selection.END (5-5) fl=#34",
-                    "hello: android.text.style.StyleSpan (0-5) fl=#33",
-                    "hello: android.text.style.UnderlineSpan (0-5) fl=#289",
-                    "hello: android.view.inputmethod.ComposingText (0-5) fl=#289",
                 )
             )
         )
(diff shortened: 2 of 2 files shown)


jmartinesp:
FWIW, I've also tested this with CJK languages and it works as expected, so I'm not sure why the previous behaviour was decided long ago.

jonnyandrew:
The exception is from Rust:
```
Caused by: uniffi.wysiwyg_composer.InternalException: assertion failed: self.is_char_boundary(idx)
```

Do you think it could be an issue with indexing? Perhaps something similar to https://github.com/matrix-org/matrix-rich-text-editor/pull/824 i.e. not taking account of code points vs code units?

jmartinesp:
We could try to solve this in Rust instead, or add some extra protection against it. But assuming the same example as above with `Te😃st`, having indexes 0-6 and valid indexes `[0, 1, 2, 4, 5, 6]`, if Android sent a `3`, should we just ignore it, or move it to `2` or `4`?

jonnyandrew:
I see what you mean... it looks like the EditText _is_ working in code units as we see by manually moving the cursor from `"Te|😃st"` (position 2) to `"Te😃|st"` (position 4), the correct indices are sent to `onSelectionChanged` .

But if we insert `"😃"` into `"Te|st"` (position 3) then at some point during `commitText()`, `onSelectionChanged` is called with index 3.

I'm wondering if there is a missing a piece inside our input connection which is causing `onSelectionChanged` to be called with an out-of-date selection. Looking at the `BaseInputConnection` I can see some [cursor positioning logic right around the point of replacing text](https://android.googlesource.com/platform/frameworks/base/+/47cf4f1/core/java/android/view/inputmethod/BaseInputConnection.java#1014). Perhaps we need to do something similar?

jmartinesp:
I was talking about `Editable.replaceText`, since that's what we are using at the moment, but I see the `BaseInputConnection` also has this same method.

The new position seems to be calculated there as:

```kotlin
newCursorPosition = newCursorPosition + start
```

In our case, we always pass ` newCursorPosition = 1` && `start = 0` at that point, so technically we could use `newCursorPosition` to set the right index, although that means setting the selection twice for some reason according to that code, and only if the underlying `baseInputConnection` is actually of type `BaseInputConnection` (right now we assume it's `InputConnection`, which has a different implementation).

However, I don't think doing this is needed at all. We're actually replacing the whole contents of the Editable with each change in the RTE, so doing `clear` and then repopulating the contents makes more sense to me than calling that `replaceText` function, which will try to preserve selection and spans.

jmartinesp:
I think it's this code that's causing the selection change, the 2nd if, probably: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/text/SpannableStringBuilder.java;l=552?q=SpannableStringBuilder

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -333,7 +333,8 @@ internal class InterceptInputConnection(
 
     private fun replaceAll(charSequence: CharSequence) {
         editable.removeFormattingSpans()
```
jonnyandrew, line 335: Is this still necessary now we are clearing the editable?

jonnyandrew (approved):
I tested and there is no issue with the flickering suggestion indicator (when using a soft keyboard). And given the text watcher issue fixed in https://github.com/matrix-org/matrix-rich-text-editor/pull/248 has not resurfaced, I don't see any issues with this change.

Thanks for fixing the issue!

platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
jmartinesp: Probably not, I'll double check just in case.

jmartinesp: It is needed, it seems like `clear()` only removes the text and selection:

image

Here I started a list and just wrote "Let". The list item span was added 3 extra times, one for each new character.

PR #889, opened 2023-11-23
Remove uniffi fork from dependencies (#783)

### LINKED ISSUE #783: Replace Uniffi fork when a new version lands
Since moving to a dynamic Swift library (https://github.com/matrix-org/matrix-rich-text-editor/pull/782), symbols won't clash anymore. So the only reason to keep our fork of `uniffi-rs` is a library mode build fix around crate types that is unreleased yet, see https://github.com/matrix-org/matrix-rich-text-editor/pull/768.

The required fix should be available in uniffi version `0.24.4` / `0.25.0`. We can then replace the configuration for the fork at https://github.com/matrix-org/matrix-rich-text-editor/blob/d76a9f861e489191cace786ea7a0536a938688d3/Cargo.toml#L19 with the standard new version of uniffi.

nimau:
This PR closes #783. 

The Uniffi fork has been replaced by the latest version of [Uniffi](https://github.com/mozilla/uniffi-rs/tags) (v0.25.2)


jonnyandrew (approved):
Looks good, glad we can do this! I've just started the Android/iOS CI jobs so let's wait for those before merging.

bindings/wysiwyg-ffi/src/ffi_dom_creation_error.rs
```diff
@@ -51,3 +51,5 @@ impl From<DomCreationError> for wysiwyg::DomCreationError {
         unimplemented!("Error is not needed as input")
     }
 }
+
+impl Error for DomCreationError {}
```
jonnyandrew, line 55: Out of curiosity, why is this needed?

nimau: It won't compile without it. I suspect it's because of the uniffi update.

jonnyandrew:
The checks have passed. I'm not sure why they are not showing on the PR itself but here they are:
- [Android](https://github.com/matrix-org/matrix-rich-text-editor/actions/runs/6972505116)
- [iOS](https://github.com/matrix-org/matrix-rich-text-editor/actions/runs/6972505105)

PR #890, opened 2023-11-24
Fix inline predictive text and keyboard suggestion toolbar use cases.

langleyd:
This PR aims to fix a number of issues with predictive suggestions(both inline and in the toolbar) causing in consistent state between RTE model and text view content. Uses cases tested include:

- tapping on the suggestion to accept it
- using spacebar to accept it
- using backspace when on the boundary of a suggestion
- Replacing a selected word with one from the keyboard suggestion toolbar

It should help us address the problems in:
https://github.com/element-hq/element-x-ios/issues/2345

This issues arise due to the old reconciliate solution.

## The Problem:
In the simpler editor use cases one can assume keystokes and selection events are reflected in both the text view an the internal RTE representation and match. 
This is not always the case though, fancy features like inline predictive suggestions, text dictation and keyboards for non-latin based languages often dump "pending" text updates into the text views text storage(an attributed string). This mismatch in content makes reconciling the rte and text view content and the selection indices difficult.

## The old solution (reconciliate with string differ):
- The text view is the source of truth 
- When it is different from our RTE content, we calculate the difference between them and try and insert it.
- This only works if there was a simple difference located in one place(which is generally the case for some of the use cases mentioned above). Any more complex differences meant it it errors out and tries to restore to a previous state. 
- This solution works for some of he main autocomplete use cases but was a bit prone to bugs and didn't cover all cases.

## The new solution:
- One fortunate behaviour of the iOS text view apis is that the `shouldChangeTextIn` delegate function is called when text is _committed_ to the text view. So for example when the predictive text is still grey, it will be reflected in `textVIew.attributedText`, but crucially `shouldChangeTextIn` will not be called until the suggestion is actually accepted(say by tapping on it).
- The new solution creates a new text storage in the form of `committedAttributedText` which represents the text that has been committed. So in addition to applying mutations to text view content to be reflected in the UI, we also apply them to `committedAttributedText`. 
- In `committedAttributedText` we have a way to determine when there is text in the editor that has not yet been committed by the user(e.g in line predictive text). 
- This combination of data allows us to better address some of the edge cases on the uses cases mentioned at the top.

Edit: We determined using this `committedAttributedText` works well for latin character languages which make sensible use of the `shouldChangeTextIn` api for committed text. Some other languages like `Japanese Kana` however do not, and will substitute characters without sensibly notifying via the apis. We therefore we have kept the reconciliate function and fall back to it if the content of the editor changes beyond simple latin character mutations.


nimau:
@langleyd, the "." shortcut seems problematic as it doesn't call `func textView(_ textView: UITextView, shouldChangeTextIn range: NSRange, replacementText text: String) -> Bool`. So the '.' is present in the textview but disappears once you delete something.

aringenbach:
> @langleyd, the "." shortcut seems problematic as it doesn't call `func textView(_ textView: UITextView, shouldChangeTextIn range: NSRange, replacementText text: String) -> Bool`. So the '.' is present in the textview but disappears once you delete something.

AFAIK any language that magically replace latin characters with symbols, have the exact same behaviour as the "." shortcut.

langleyd:
I did test a number of CJK keyboards and from my initial tests I couldn't break it.

On testing again though, I can see that although `shouldChangeTextIn` does seem to be called (unlike the "." shortcut), the index in some cases is not correct. For example of you use the Koran 10 key keyboard and cycle through the characters by pressing 1 key multiple times, it replaces the character in the editor but adds each one in the RTE state as the index returned does not really seem correct. :/

aringenbach:
@langleyd I'm not sure anymore about Korean keyboards behaviour, they might have less issues than some of the others. 
I think you can try with Japanese Romaji Qwerty for example, this one had lots of issues, and most system events will not get triggered, just magically replaces the text.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -135,7 +135,18 @@ public class WysiwygComposerViewModel: WysiwygComposerViewModelProtocol, Observa
...
-
+    
+    /// This is used to track the text commited to the editor by the user, as opposed to text
+    /// that could be in the editor that is not yet committed (e.g. from inline predictive text or dictation ).
+    private lazy var committedAttributedText = NSAttributedString(string: "", attributes: defaultTextAttributes)
+    
+    private var lastReplaceTextUpdate: ReplaceTextUpdate?
+    
+    /// Wether the view contains uncommitted text(e.g. a predictive suggestion is shown in grey).
+    private var viewHasUncommitedText: Bool {
```
Velin92, line 146: would rename this `hasUncommitedText`

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -378,13 +419,50 @@ public extension WysiwygComposerViewModel {
             }
             plainTextContent = textView.attributedText
         } else {
-            reconciliateIfNeeded()
+            let dotInserted = checkForDoubleSpaceToDotConversion()
+            if !dotInserted {
```
Velin92, line 423: if `!isDoubleSpaceToDotConversion`

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -378,13 +419,50 @@ public extension WysiwygComposerViewModel {
...
+        }
+        let dotRange = dotStartIndex..<dotEndIndex
+        let textPotentialDot = String(text[dotRange])
+        let contentPotentialDot = String(content[dotRange])
+        if textPotentialDot == ".", contentPotentialDot != "." {
+            let replaceUpdate = model.replaceTextIn(newText: ".",
+                                                    start: UInt32(dotStart),
+                                                    end: UInt32(dotEnd))
+            applyUpdate(replaceUpdate, skipTextViewUpdate: true)
+            return true
```
Velin92, line 461: I have now seen that tis function actually also modifies the content... okay... at this point instead of having it has a property... I would keep it as a function.... and have it return the `replaceUpdate` as an optional.
Would avid having a function like this have a collateral effect.

so it would be something like 
```
if let update = doubleSpaceToDotConversionUpdate() {
	applyUpdate(replaceUpdate, skipTextViewUpdate: true)
	reconciliateIfNeeded()
} 
```

when used inside the replace text function

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -378,13 +419,50 @@ public extension WysiwygComposerViewModel {
             }
             plainTextContent = textView.attributedText
         } else {
-            reconciliateIfNeeded()
+            let dotInserted = checkForDoubleSpaceToDotConversion()
+            if !dotInserted {
+                reconciliateIfNeeded()
+            }
```
Velin92, line 425: check my comment about this function

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -560,15 +646,23 @@ private extension WysiwygComposerViewModel {
...
-
+    
     /// Reconciliate the content of the model with the content of the text view.
     func reconciliateIfNeeded() {
         do {
             guard !textView.isDictationRunning,
                   let replacement = try StringDiffer.replacement(from: attributedContent.text.htmlChars,
-                                                                 to: textView.attributedText.htmlChars) else {
+                                                                 to: textView.attributedText.htmlChars)
+            else {
```
Velin92, line 656: `swiftformat` doesn't complain about this?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -656,3 +750,18 @@ extension WysiwygComposerViewModel: ComposerModelWrapperDelegate {
...
+    var date: Date
+    var range: NSRange
+    var text: String
+}
+
+private extension ReplaceTextUpdate {
+    func isDuplicate(of other: ReplaceTextUpdate) -> Bool {
+        range == other.range
+            && text == other.text
+            && fabs(date.timeIntervalSince(other.date)) < Self.debounceThreshold
```
Velin92, line 765: why is this necessary?

platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Extensions/String+LatinLangugesTests.swift
```diff
@@ -0,0 +1,36 @@
...
+import XCTest
+
+final class StringLatinLangugesTests: XCTestCase {
+    func testLatinLangugeCharacters() throws {
+        XCTAssertTrue("hello".containsLatinAndCommonCharactersOnly)
+        XCTAssertTrue("helló".containsLatinAndCommonCharactersOnly)
+        XCTAssertTrue("helló, ".containsLatinAndCommonCharactersOnly)
+        XCTAssertTrue("helló, ".containsLatinAndCommonCharactersOnly)
+        XCTAssertTrue("😄🛴🤯❤️".containsLatinAndCommonCharactersOnly)
+        XCTAssertTrue(String.object.containsLatinAndCommonCharactersOnly)
```
Velin92, line 27: what's this test case about?

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -309,28 +321,57 @@ public extension WysiwygComposerViewModel {
...
             textView.typingAttributes = defaultTextAttributes
         }
 
         let update: ComposerUpdate
-        let shouldAcceptChange: Bool
+        let skipTextViewUpdate: Bool
 
+        // The system handles certain auto-compelete use-cases with somewhat unusual replacementText/range
+        // combinations, some of those edge cases are handled below.
+
```
pixlwave, line 344: The `let isReplacingWordWithSuggestion` and `if isReplacingWordWithSuggestion` could go up here before the other 2 properties to help follow the logic when reading through.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -340,18 +381,18 @@ public extension WysiwygComposerViewModel {
...
             }
         default: break
         }
         
-        applyUpdate(update, skipTextViewUpdate: shouldAcceptChange)
-
-        return shouldAcceptChange
+        applyUpdate(update, skipTextViewUpdate: skipTextViewUpdate)
+        lastReplaceTextUpdate = nextTextUpdate
+        return skipTextViewUpdate
```
pixlwave, line 393: From the outside, this change in return is hard to understand given that the nature of this property is inverted, but there's no doc comments to explain what it is meant to signal to the caller.

platforms/ios/example/WysiwygUITests/WysiwygUITests+InlinePredictiveText.swift
```diff
@@ -0,0 +1,56 @@
...
+    }
+    
+    func testDoubleSpaceIntoDot() {
+        sleep(5)
+        app.typeTextCharByCharUsingKeyboard("Hello")
+        app.keys["space"].tap()
+        app.keys["space"].tap()
+        assertTextViewContent("Hello. ")
+    }
+}
```
Velin92, line 44: @langleyd added these UI tests to check if our most important cases are covered and they work very well.... however the predictive text test may be a bit flaky if the simulator is not recently initialised, which luckily should not be an issue on the CI

Velin92: Are there any other test cases we may want to cover that this PR addresses? 
(I am also looking into finding a way to test CJK, but seems there is no reliable way to simulate a keyboard suggestion tap, will look more into it)

platforms/ios/example/Wysiwyg.xcodeproj/project.pbxproj
```diff
@@ -392,6 +396,24 @@
...
+			inputPaths = (
+			);
+			name = "Force software keyboard on simulator";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "external_kb_connected=false\n\nosascript -e 'quit app \"Simulator\"'\n\nSIMUS_KEYBOARD=$(/usr/libexec/PlistBuddy -c \"Print :DevicePreferences\" ~/Library/Preferences/com.apple.iphonesimulator.plist | perl -lne 'print $1 if /^    (\\S*) =/')\n\necho \"$SIMUS_KEYBOARD\" | while read -r a; do /usr/libexec/PlistBuddy -c \"Set :DevicePreferences:$a:ConnectHardwareKeyboard $external_kb_connected\" ~/Library/Preferences/com.apple.iphonesimulator.plist || /usr/libexec/PlistBuddy -c  \"Add :DevicePreferences:$a:ConnectHardwareKeyboard bool $external_kb_connected\" ~/Library/Preferences/com.apple.iphonesimulator.plist; done\n";
```
Velin92, line 415: this script is needed to force software keyboard on the sim for UI tests, since predictive text will only work if the software keyboard is used directly, on the simulator

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
```diff
@@ -300,6 +311,7 @@ public extension WysiwygComposerViewModel {
         compressedHeight = min(maxCompressedHeight, max(minHeight, idealTextHeight))
     }
 
+    // swiftlint:disable cyclomatic_complexity
```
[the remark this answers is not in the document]
langleyd: I think it is unless we have a sensible way to reduce the complexity of the function, the warning still exists when I remove this.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
langleyd: I'm not sure I follow where the above code is supposed to go?  The func `replaceText(range: NSRange, replacementText: String)` fuction? 

I believe`reconciliateIfNeeded` and ` doubleSpaceToDotConversion` need to be called from `didUpdateText`. `didUpdateText` is called after the text view is updated. After it's updated we are checking if the textView was changed from under us (via 2 spaces => dot or some other change to the text view that requires the string differ/reconciliate)

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
langleyd: Doesn't appear to, want the else moved up to the previous line? happy to do so.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
langleyd: There is a descriptive comment where this function is called. We get 2 text updates with the same values when we tap on predictive inline text. This function is used to determine if we should ignore the second of these updates. `isDuplicate` is a bit misleading. `shouldDebounce` might be clearer.

platforms/ios/lib/WysiwygComposer/Tests/WysiwygComposerTests/Extensions/String+LatinLangugesTests.swift
langleyd: This is the obeject replacement unicode character as defined [here](https://github.com/matrix-org/matrix-rich-text-editor/blob/main/platforms/ios/lib/WysiwygComposer/Sources/HTMLParser/Extensions/String%2BCharacter.swift#L31). It get's inserted during some of the`shouldChangeTextIn` updates, I think I noticed it during voice dictation so is one we should accept.

Velin92:
I also found another bug with CJK keyboards.
If you type a word for example "kudesai" with a romanji keyboard, before accepting any suggestion, if you tap backspace to delete the last character we get, happens that we either get a html range error, o we duplicate the text and append it at the end.
@langleyd 

I am trying to write a test for it to reproduce it through it

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
langleyd: `shouldAcceptChange ` seemed a bit unclear to me given the language on line 391 was 

```
applyUpdate(update, skipTextViewUpdate: shouldAcceptChange)
```

So I was trying to make it consistent.

I've documented the return value in the function signature. Does that clear things up?

platforms/ios/example/WysiwygUITests/WysiwygUITests+InlinePredictiveText.swift
langleyd: No these cover the ones I can think of without the software keyboard.

langleyd:
Interesting thanks @Velin92 will test it tomorrow. Sounds a bit like the `containsLatinAndCommonCharactersOnly` might be handling some cases incorrectly :/

Velin92:
> Interesting thanks @Velin92 will test it tomorrow. Sounds a bit like the `containsLatinAndCommonCharactersOnly` might be handling some cases incorrectly :/

can actually be reproduced even more easily by simply writing with the romanji keyboard any japanese character for example when writing "ku" which is converted into "く" , and then tapping backspace before accepting the suggestion makes throws a range error and prevents deletion

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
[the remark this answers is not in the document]
Velin92: okay I see then I would suggest replacing it with `// swiftlint:disable:next cyclomatic_complexity` which will only suppress the next warning, this line may disable the warning everywhere in the codebase

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
Velin92: yeah they should go in the `didUpdateText` (my bad mistyped the name) the idea is just to turn the function into a simple function that returns an optional, and the apply the  update in the `didUpdateText`.
This is because this function returns a boolean but is also applying changes in the textView, which is not clear from the name and intent of the function

langleyd:
Ive added a fix @Velin92 

```
  if range != attributedContent.selection {
            select(range: range)
        }
```

needed to moved above the logic that used `attributedContent.selection` otherwise the selection of the textview does not match that of `attributedContent.selection` and you get the index issue.

I have found another though, I'll try create a test case for it.

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
pixlwave: Yep yep 👍

langleyd:
When you accept an inline predictive text suggestion with spacebar, then tap `.`. It inserts the dot before the space, next to the last work, but the `shouldChangeTextIn` gives you the index as if it's just inserted it after the space. 🫠

It would have to be another special case for inline predictive text. (unless we want to go back to blanket reconciliate magic, or build the proper tree based reconcile).

platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerViewModel.swift
langleyd: I've just handled the new case of the dot after the inline text prediction, which is similar but different to the two space to dot conversion.

I've kept the `applyUpdate` inside the functions as it is intrinsically linked to the model update so I think should be kept together to handle each special case. I've updated function names and also added docs, so hopefully its clear to the calling site.

langleyd:
The moving back of the dot after accepting inline text prediction was an optimisation brought in in 17.5. I was able to verify outside of out app(e.g. in the reminders app). This is why it was failing in the CI(17.2) but not for me locally(17.5).

I've added the version check, so should be all good now.

Velin92 (approved):
the only thing left is to change the cyclomatic complexity part @langleyd

PR #895, opened 2023-11-29
Android: handle all `ClickableSpans` in the gesture detector

jmartinesp:
While testing the latest version on EX Android I realised auto-linkified links (using `URLSpan`) weren't clickable since the gesture detector ignored them. This fixes the issue.

### DIFF
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/EditorStyledTextViewTest.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/EditorStyledTextViewTest.kt
@@ -3,6 +3,7 @@ package io.element.android.wysiwyg
 import android.graphics.Canvas
 import android.graphics.Paint
 import android.text.style.ReplacementSpan
+import android.text.style.URLSpan
 import android.widget.TextView
 import androidx.core.text.buildSpannedString
 import androidx.core.text.inSpans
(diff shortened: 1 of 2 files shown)


platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/EditorStyledTextViewTest.kt
```diff
@@ -34,6 +35,29 @@ internal class EditorStyledTextViewTest {
             .check(matches(withText("Hello, world")))
     }
 
+    @Test
+    fun testUrlClicks() {
```
jonnyandrew, line 39: Should we also add a test for setting a mention using `setHtml()` and then clicking it? It shouldn't block this PR but noticed it is missing in the [coverage report](https://app.codecov.io/gh/matrix-org/matrix-rich-text-editor/pull/895/blob/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt).

jonnyandrew (approved):
Thanks for the test!

platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/EditorStyledTextViewTest.kt
jmartinesp: Added it in https://github.com/matrix-org/matrix-rich-text-editor/pull/895/commits/ed4b1ed53d78e4ccc549379283474e285d6e4f98. Let's see if it gets correctly picked by the coverage collector.

PR #896, opened 2023-11-29
[Android] Improve text view test

jonnyandrew:
- Use Espresso `ViewAction`s instead of `onActivity` + `findViewById`.
- Add `FakeClickListener` test util.
- Add assertions for the correct URL being clicked.
- Add test for parsing a mention and clicking it.


jmartinesp (approved):
LGTM, thanks!

PR #897, opened 2023-11-29
[Android] Allow specifying line height for all API versions

jmartinesp:
The previous version only allows specifying this value for API >= 28, and does not protect the user against setting a negative or unspecified value.

### DIFF
--- a/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/RichTextEditorStyleExt.kt
+++ b/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/RichTextEditorStyleExt.kt
@@ -16,6 +16,8 @@ import androidx.compose.ui.text.font.FontStyle
 import androidx.compose.ui.text.font.FontSynthesis
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.Density
+import androidx.compose.ui.unit.TextUnitType
+import androidx.compose.ui.unit.isSpecified
 import androidx.core.content.ContextCompat
 import io.element.android.wysiwyg.compose.BulletListStyle
 import io.element.android.wysiwyg.compose.CodeBlockStyle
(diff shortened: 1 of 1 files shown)


platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/RichTextEditorStyleExt.kt
```diff
@@ -106,6 +118,12 @@ internal fun TextView.applyStyleInCompose(style: RichTextEditorStyle) {
     }
 }
 
+private fun TextUnitType.toTypeUnit() = when (this) {
+    TextUnitType.Sp -> TypedValue.COMPLEX_UNIT_SP
+    TextUnitType.Em -> TypedValue.COMPLEX_UNIT_FRACTION_PARENT
+    else -> TypedValue.COMPLEX_UNIT_PX
```
jonnyandrew, line 124:
```suggestion
    TextUnitType.Unspecified -> TypedValue.COMPLEX_UNIT_PX
```

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/RichTextEditorStyleExt.kt
```diff
@@ -89,13 +91,23 @@ internal fun TextStyle.rememberTypeface(): State<Typeface> {
...
 internal fun TextView.applyStyleInCompose(style: RichTextEditorStyle) {
     setTextColor(style.text.color.toArgb())
-    val lineHeightInPx = TypedValue.applyDimension(
-        TypedValue.COMPLEX_UNIT_DIP,
-        style.text.lineHeight.value,
-        context.resources.displayMetrics
-    )
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
-        lineHeight = lineHeightInPx.roundToInt()
+    if (style.text.lineHeight.isSpecified && style.text.lineHeight.value > 0f) {
```
jonnyandrew, line 94: Could the style class enforce this requirement somehow? And provide a default line height also?

jonnyandrew, line 94: If we can't enforce that `style.text.lineHeight` is valid, can we add an `else` which applies a default line height?

jonnyandrew (approved):

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/RichTextEditorStyleExt.kt
jmartinesp: I don't think this is enough, since `TextUnitType` is a value class, not an enum. I tried using that before and the IDE forced me to add an else branch.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/internal/RichTextEditorStyleExt.kt
jmartinesp: The default line hight is already provided in `textStyle()`. But I don't think `TextUnit` actually enforces a positive value at all, and I don't think we can enforce them with annotations either. This is just here to protect agains crashes because some invalid value was passed. We could maybe log an error instead?

Also, passing `TextUnit.Unspecified` can be useful to default to the line height of the component instead of the default one contained in the `TextStyle` data class.

PR #898, opened 2023-11-30
[Android] Fix line height changes in `EditorStyledText`

jmartinesp:
When using `EditorStyledText` for the timeline items in EXA I sometimes saw the line height changing in real time. 

This was probably caused by setting this line height before the typeface, since this latter change recalculates the line height too.

While at it, I thought it would be nice to be able to customise `includeFontPadding` so the text in these components behave the same as the one in Compose.

Also, I think it's probably better to default to the local text style and font color than just setting a default one ourselves? If this is not right, I can revert these changes.


bmarty (approved):
LGTM, thanks!

PR #899, opened 2023-11-30
[Android] Fixes needed to get `EditorSyledTextView` to work with screenshot tests

jmartinesp:
## Changes

- Workaround for isInEditMode() being always false in Paparazzi tests. `isInEditMode` is used in Android code to decide whether to load the native libs or not.
- Use `Context` instead of `Application` in `AndroidResourcesHelper` (LayoutLib in Paparazzi can't cast `applicationContext` to `Application`).


bmarty (approved):
LGTM, thanks!

PR #900, opened 2023-12-01
[Android] Add typing callback

jmartinesp:
This adds the missing typing callback for having typing notifications in the Android app.

### DIFF
--- a/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
+++ b/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
@@ -39,7 +39,7 @@ import timber.log.Timber
  * @param inputType The input type for the editor. Defaults to [RichTextEditorDefaults.inputType].
  * @param resolveMentionDisplay A function to resolve the [TextDisplay] of a mention.
  * @param resolveRoomMentionDisplay A function to resolve the [TextDisplay] of an `@room` mention.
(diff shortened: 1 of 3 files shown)


platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/RichTextEditor.kt
```diff
@@ -120,9 +124,17 @@ private fun RealEditor(
...
+                        val prevInternalHtml = state.internalHtml
+                        val prevMarkdown = state.messageMarkdown
+
                         state.internalHtml = getInternalHtml()
                         state.messageHtml = getContentAsMessageHtml()
                         state.messageMarkdown = getMarkdown()
+
+                        // This is needed so we don't send a typing status when the editor is first displayed
+                        if (prevInternalHtml != state.internalHtml || prevMarkdown != state.messageMarkdown) {
+                            onTyping()
```
bmarty, line 136: Do we want to provide a boolean to indicate if the content is empty (so the app should consider that the user stops typing) of not (so the user starts/continues typing)? Or can this will be handled (and debounce/filtered) app side?

jmartinesp: Oh, I guess we could. The Rust code also has some automatic 4s timeout for sending a 'not typing anymore' event if I'm not mistaken, so these could complement each other.

PR #902, opened 2023-12-12
[Android] Implement `replaceText` 

jmartinesp:
This is used in Android 13+ instead of `commitText` for replacing text with suggestions.

It should fix the issue of the RTE input sometimes not being up to date after selecting a suggestions in newer Android APIs.

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
@@ -141,6 +141,29 @@ internal class InterceptInputConnection(
         }
     }
 
+    // In Android 13+, this is called instead of [commitText] when selecting suggestions from IME
+    override fun replaceText(
+        start: Int,
+        end: Int,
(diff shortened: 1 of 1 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnection.kt
```diff
@@ -141,6 +141,29 @@ internal class InterceptInputConnection(
...
+    ): Boolean {
+        val result = processTextEntry(text, start, end)
+
+        return if (result != null) {
+            beginBatchEdit()
+            replaceAll(result.text)
+            val editorSelectionIndex = editorIndex(result.selection.last, editable)
+            setSelection(editorSelectionIndex, editorSelectionIndex)
+            setComposingRegion(editorSelectionIndex, editorSelectionIndex)
+            endBatchEdit()
```
bmarty, line 160: Maybe extract to a sub function to avoid duplicated code but not a big deal.

bmarty (approved):
LGTM, thanks for fixing this!

bmarty (approved):

PR #904, opened 2023-12-13
Fix enter right after mention crashing the editor

### LINKED ISSUE #903: Crash when adding a new line right after a mention
## Problem

When adding a line break right after a mention, the composer crashes.

## Steps to reproduce
- Add a mention to the composer. This will also add a whitespace after it.
- Remove the whitespace and place the cursor right after the mention.
- Hit enter/add a line break.

The composer crashes. This happens because when adding a new line we're supposed to split the DOM tree in 2 and use paragraphs/line break nodes in between, but the code doesn't make an exception for mentions, which have only 1 char length.

## Proposed solution

Fix the DOM tree splitting algorithm.

jmartinesp:
Should fix #903 .

### DIFF
--- a/crates/wysiwyg/src/composer_model/new_lines.rs
+++ b/crates/wysiwyg/src/composer_model/new_lines.rs
@@ -480,4 +480,28 @@ mod test {
         </ul>"
         );
     }
+
+    #[test]
+    fn test_enter_before_mention() {
+        let mut model = cm(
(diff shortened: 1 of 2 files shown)


bmarty (approved):
Thanks. I do not understand everything, but I guess you fixed the issue.
I am wondering if we would need some Rust dev review for such PR (and not a review from https://github.com/orgs/matrix-org/teams/element-android).

Velin92 (approved):

PR #907, opened 2023-12-13
[Rust] fix multiple lines in markdown mode

Velin92:
Markdown mode was not able to parse newlines, and when the number of consecutive newlines was more than 1 the model generation panicked, resulting in an empty DOM when sending the message.
This makes markdown mode more stable, by allowing newlines to be parsed correctly.

This PR also removes the shiftEnter KeyCommand from iOS since we never actually use it/override in any project, and it just creates the issue of bypassing the correct shift+enter behaviour in markdown mode, which is the behaviour for the software return keyboard on iOS.
Is probably best to completely remove it until we find a better architecture to handle the key commands.


jmartinesp (approved):
The Rust changes LGTM, but maybe we could add some comments to explain why they're needed?

PR #908, opened 2023-12-13
[iOS] Key Commands Refactor

Velin92:
The issue with the current KeyCommandsHandler, is that even if the key command is set to be false, the command is not really ignored and was stll being handled by the default function, which at the same time was not working properly with plain text mode since it forced an update in the model.

Also the usage of the enter() function was redundant also for the RTE mode, since we already handle the enter directly in the replaceText properly, and we are constraining the usage of the key commands only to the ones we have defined in the library not giving to the users of the library a proper control on which and when they should be and should enabled.

The alternative approach was to disabled the default handling in plain text mode for such commands, but this created two issues:
- The command was still handled by the TextView, which was then preventing the key to use the replace text function in the delegate properly, which means that shift+enter and enter were not doing anything.
- The ignored command was not really ignored at all, but just silently handled in the background, which kinda created a side effect that was not clear to the user of the library

The current solution allows full control and customisation for the key commands while also providing the possibility of creating some default rapid key commands (like .enter) for the general use cases, while also fixing the shift+enter bug in plain text mode.

### DIFF
--- a/platforms/ios/example/Wysiwyg/Views/Composer.swift
+++ b/platforms/ios/example/Wysiwyg/Views/Composer.swift
@@ -22,7 +22,7 @@ import WysiwygComposer
 struct Composer: View {
     @ObservedObject var viewModel: WysiwygComposerViewModel
     let itemProviderHelper: WysiwygItemProviderHelper?
-    let keyCommandHandler: KeyCommandHandler?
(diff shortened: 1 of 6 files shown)


pixlwave (approved):
Whilst I'm not familiar with the insides of RTE, this looks like a totally reasonable change to me and makes the API much nicer to use to my eyes.

PR #912, opened 2023-12-26
Android: add `onTextLayout` listener on the TextView component

jmartinesp:
This should allow us to use this result for text-based layouts.

### DIFF
--- a/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
+++ b/platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
@@ -1,5 +1,6 @@
 package io.element.android.wysiwyg.compose
 
+import android.text.Layout
 import android.text.Spanned
 import android.widget.TextView
 import androidx.compose.runtime.Composable
@@ -34,6 +35,7 @@ fun EditorStyledText(
     resolveMentionDisplay: (text: String, url: String) -> TextDisplay = RichTextEditorDefaults.MentionDisplay,
     resolveRoomMentionDisplay: () -> TextDisplay = RichTextEditorDefaults.RoomMentionDisplay,
     onLinkClickedListener: ((String) -> Unit) = {},
+    onTextLayout: (Layout) -> Unit = {},
     style: RichTextEditorStyle = RichTextEditorDefaults.style(),
 ) {
     val typeface by style.text.rememberTypeface()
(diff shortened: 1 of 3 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorStyledTextView.kt
```diff
@@ -158,6 +159,12 @@ open class EditorStyledTextView : AppCompatTextView {
         htmlConverter?.fromHtmlToSpans(htmlText)?.let { setText(it, BufferType.SPANNABLE) }
     }
 
+    override fun onMeasure(widthMeasureSpec: Int, heightMeasureSpec: Int) {
+        super.onMeasure(widthMeasureSpec, heightMeasureSpec)
+
+        layout?.let { onTextLayoutChanged?.invoke(it) }
```
jmartinesp, line 165: Here the text has already been laid out so we can return `this.layout` through the callback.

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/RustCleanerTask.kt
```diff
@@ -1,13 +1,16 @@
...
+import io.element.android.wysiwyg.BuildConfig
 import timber.log.Timber
 import uniffi.wysiwyg_composer.Disposable
 
 internal class RustCleanerTask(
     private val disposable: Disposable,
 ) : Runnable {
     override fun run() {
-        Timber.d("Cleaning up disposable: $disposable")
+        if (BuildConfig.DEBUG) {
```
jmartinesp, line 11: I added this because the debug logs added here were being printed by the EXA app in debug mode. I could also remove the log completely.

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
```diff
@@ -34,6 +35,7 @@ fun EditorStyledText(
     resolveMentionDisplay: (text: String, url: String) -> TextDisplay = RichTextEditorDefaults.MentionDisplay,
     resolveRoomMentionDisplay: () -> TextDisplay = RichTextEditorDefaults.RoomMentionDisplay,
     onLinkClickedListener: ((String) -> Unit) = {},
+    onTextLayout: (Layout) -> Unit = {},
```
bmarty, line 38: Is there a reason why the parameter is not called `onTextLayoutChanged`?

bmarty (approved):
LGTM, thanks!

platforms/android/library-compose/src/main/java/io/element/android/wysiwyg/compose/EditorStyledText.kt
jmartinesp: Not really, I think I just missed the mismatch here.

PR #916, opened 2024-01-03
[Android + Rust] Several editor fixes and tweaks

jmartinesp:
Changes in this PR:

## Android

- Fix incorrect transformation of NBSP chars to 'extra char' spans to indicate offsets between the RTE state and the EditText.
- Add undo/redo context menu actions, also add them to Ctrl+Z and Ctrl+Shift+Z keyboard shortcuts (which is what some software keyboards use).
- Make sure we don't try to apply spans outside the bounds of the parsed text.
- Make sure applying a selection only happens inside the actual bounds of the `Editable` object.
- Destroy the `ComposerModel` when it crashes and we instantiate a new one. It should remove a source of memory leaks.

## Rust

- Fix for replacing text/deleting several selected empty paragraphs. Previously, any text in them was removed but the actual paragraphs were kept.
- After replacing text, make sure the new selection is inside the bounds of the DOM tree.

# Tests

If you want to run manual tests, to double check the changes:

- Open either the Android app with these changes or build the `example-compose` project.
- Write a text with several empty paragraphs (consecutive line breaks). The select all and delete it. The editor should be blank.
- Paste some text with some NBSP characters, try editing text after these characters. If it doesn't crash or behaves strangely, it's working properly.

The rest of the changes are for issues that we don't really know how to reproduce, but we know are present.


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -259,6 +259,14 @@ class EditorEditText : AppCompatEditText {
...
                 return false
             }
+            android.R.id.undo -> {
+                undo()
+                return false
+            }
+            android.R.id.redo -> {
+                redo()
+                return false
+            }
```
bmarty, line 269: Maybe update the comment of this fun?

Also the doc of `TextView.onTextContextMenuItem` says:

> Returns: true if the context menu item action was performed.

So I do not see why all the handled cases here are returning `false` and not `true`. But it looks like parents are also returning `false` so 🤷‍♂️ .

bmarty (approved):
LGTM for the Kotlin part.

PR #918, opened 2024-01-04
Disable AVD cache for Android CI job

### LINKED ISSUE #917: Android CI gets stuck
The `Android` CI step gets stuck when trying to kill the emulator, it's probably what's seen [here](https://github.com/ReactiveCircus/android-emulator-runner/issues/362).

Disabling the AVD cache would probably workaround the issues, but it'll make the Android test step way longer.

jmartinesp:
Should fix #917 .


bmarty (approved):
OK

PR #919, opened 2024-01-05
Android: add IME rich content compatibility

jmartinesp:
This is enabled on API >= 25 (Android 7.1).

I had one device on Android 11 that was refusing to consume any rich content (images, stickers, etc.) from Gboard, while it worked on every other device and emulator I tried. With this change, it works.


bmarty (approved):
Thanks for investigating and fixing this issue!

PR #922, opened 2024-01-12
Fix for dictation not working on iOS 17

### LINKED ISSUE #921: [iOS] Dictiation is broken on iOS 17.0 and above
This is caused by the fact that the new AI based dictation system, retains the current language of the keyboard, and does not default back to a temporary "dictation" general language, making our check to determine if dictation has been activated useless.

Velin92:
This should fix the issue for all iOS versions, since we are relying on a specific keyboard notification to intercept the state

fixes #921


stefanceriu (approved):
Looks good to me, ship it! 🚀

PR #931, opened 2024-02-07
[iOS] Text View is now created by the Wrapper, and swapped in case it's re-rendered

### LINKED ISSUE #930: iOS: re-rendering RTE will make the composer not interactable
This issue is what causes:
https://github.com/element-hq/element-x-ios/issues/2364

The issue is caused by the fact that the `makeUIView` function is not instantiating the view, but is actually just taking the current textView reference in the WysiwygViewModel.

Since on El-X when rotating the screen and switching from a split view to a stack view and viceversa, we regenerate the view, but we keep the same object models to restore their states, what happens is that the new re-rendered SwiftUI view embedding the text view is reusing the instance from the previous view hierarchy, causing it to break its frame.

The `makeUIView` should actually be the one responsible of instantiating the UIKit view since its responsible of managing the view in the view hierarchy (even the documentation specifies that this function should be the one responsible of instantiating the view object in the first place).
We might need to re-architecture a bit the iOS side of RTE to allow remove the depency that has been created between the view model and the UIKit text view.

Velin92:
fixes #930

### DIFF
--- a/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
+++ b/platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
@@ -149,7 +149,7 @@ struct UITextViewWrapper: UIViewRepresentable {
         return textView
     }
 
-    func updateUIView(_ uiView: WysiwygTextView, context: Context) { 
+    func updateUIView(_ uiView: WysiwygTextView, context: Context) {
         if uiView !== viewModel.textView {
             viewModel.textView = uiView
         }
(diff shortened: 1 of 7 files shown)


platforms/ios/lib/WysiwygComposer/Sources/WysiwygComposer/Components/WysiwygComposerView/WysiwygComposerView.swift
```diff
@@ -121,8 +121,15 @@ struct UITextViewWrapper: UIViewRepresentable {
     }
     
     func makeUIView(context: Context) -> WysiwygTextView {
-        let textView = viewModel.textView
-        
+        // Default text container have a slightly different behaviour
+        // than what iOS would use if textContainer is nil, this
+        // fixes issues with background color not working on nealine characters.
```
stefanceriu, line 126:
```suggestion
        // fixes issues with background color not working on newline characters.
```

stefanceriu (approved):
Looks good to me! 👏  It definitely beats rewriting the whole thing.

PR #935, opened 2024-02-08
[iOS] Improved text view injection

Velin92:
- A default text view is always provided, to allow for easier testing and avoid using an optional, even if you should always inject it from the UI wrapper
- The initialisation is done directly by a convenience init from the text view itself
- the MentionsDisplayHelper is now managed by the viewModel which will refresh it on the textView


pixlwave (approved):
👍

PR #937, opened 2024-02-28
Android : fix touch handling on EditorStyledText

ganfra:
The touch events were not handled correctly as it could lead to touch events being processed twice.

PR #938, opened 2024-02-29
Android: try to get AVD action working again

jmartinesp:
After some changes in the available space in the GH runners, the Android emulator job was failing because there was not enough space.

This PR contains the next changes:
- We were downloading the Android SDK twice: once when setting up Gradle and another when launching the emulator action. This takes *lots* of disk space and was the main source of issues. Since we can't stop the AVD action from re-downloading the SDK, we'll avoid pre-building the tests so the first SDK download is skipped instead.
- Downsized the Android emulator's data partition to 3GB (in theory, I'm not confident this is actually changing anything).
- Fix the broken UI tests introduced in #937 (my bad!) since the tests were skipped there because of this issue.


ganfra (approved):
Thanks !

PR #939, opened 2024-02-29
Bump Rust version to 1.76 (stable) to fix WASM issues

jmartinesp:
Changes:
- Bump Rust to `v1.76` and fix the clippy issues found.
- Bump UniFFi to `v0.26.1`. To be compatible.
- Bump `speculoos` dependency since it became incompatible.
- Remove AVD cache from the Android CI, reduce the AVD `data` partition size to 3GB.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set Version
-      run: rustup default 1.74
+      run: rustup default 1.76
     - name: Set Component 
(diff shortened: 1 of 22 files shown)


.github/workflows/rust.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set Version
-      run: rustup default 1.71
+      run: rustup default 1.74
```
ganfra, line 24: Why not 1.76 ?

ganfra (approved):

PR #940, opened 2024-03-01
Avoid dispatching to text watcher an empty text when calling replaceAll

ganfra:
We have a custom `InterceptInputConnection` and to avoid messing up with selection/composition spans, we are calling sequentially : 
```
editable.clear
editable.append
```
Which causes registered text watchers to get an empty string which shouldn't be dispatched.

This PR introduce a `TextWatcher` which can be temporary disabled.

I don't want the burden of filtering this events on the client side.


jmartinesp (approved):
LGTM. The idea is quite clever, and it was something I thought about doing back in the day, but never found a good way to implement. I wish it wasn't needed though...

PR #942, opened 2024-03-08
[iOS] Disabling auto correction

Velin92:
Until we fix the predictive text issue is better to just disable it


pixlwave (commented):
LGTM 👍

stefanceriu (approved):

PR #944, opened 2024-03-11
Android: make `initialMarkdown` in `RichTextEditorState` customisable

jmartinesp:
This is needed to fix the send button state in Android for a composer with a blank (all whitespace) text.


bmarty (approved):
LGTM, thanks!
I am wondering if this addition will also be useful when we will work on drafts.

jmartinesp:
> I am wondering if this addition will also be useful when we will work on drafts.

I hope it'll be!

PR #948, opened 2024-03-18
Android: reconcile UI text with editor text

jmartinesp:
This should help in cases where the UI text is modified externally by the OS or some app.

Basically, what we do here is we intercept any change made outside of the editor by either the OS or the implementing app, pass them to the editor, apply them, then notify the text watchers of the result of these editor changes.

It seems to work well so far, the tests pass. We probably need to thoroughly test it with some day-to-day usage.

PS: this is heavily based on @ganfra 's work [here](https://github.com/matrix-org/matrix-rich-text-editor/pull/940). Thanks!

## Video example

https://github.com/matrix-org/matrix-rich-text-editor/assets/480955/33234304-038f-40c5-bbd3-b00bd3ffc7f0

### DIFF
--- a/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
+++ b/platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
@@ -8,8 +8,10 @@ import android.graphics.Canvas
 import android.net.Uri
 import android.os.Build
 import android.os.Parcelable
+import android.text.Editable
 import android.text.Selection
 import android.text.Spanned
+import android.text.TextWatcher
 import android.util.AttributeSet
 import android.view.KeyEvent
 import android.view.MotionEvent
@@ -37,13 +39,15 @@ import io.element.android.wysiwyg.view.inlinebg.SpanBackgroundHelperFactory
 import io.element.android.wysiwyg.view.models.InlineFormat
 import io.element.android.wysiwyg.view.models.LinkAction
 import io.element.android.wysiwyg.view.spans.ReuseSourceSpannableFactory
+import timber.log.Timber
 import uniffi.wysiwyg_composer.*
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * An [EditText] that handles rich text editing.
  */
 class EditorEditText : AppCompatEditText {
-
+    private lateinit var textWatcher: EditorTextWatcher
     private var inputConnection: InterceptInputConnection? = null
 
     /**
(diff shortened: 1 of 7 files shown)


platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -615,3 +637,61 @@ private fun KeyEvent.isMovementKey(): Boolean {
...
+        nestedWatchers.forEach { it.beforeTextChanged(s, start, count, after) }
+    }
+
+    fun notifyOnTextChanged(s: CharSequence?, start: Int, before: Int, count: Int) {
+        nestedWatchers.forEach { it.onTextChanged(s, start, before, count) }
+    }
+
+    fun notifyAfterTextChanged(s: Editable?) {
+        nestedWatchers.forEach { it.afterTextChanged(s) }
+    }
```
jmartinesp, line 696: I don't really like this solution, to be honest. I'd rather use the original overridden methods, but I couldn't find a good way to do it.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -578,6 +582,18 @@ class EditorEditText : AppCompatEditText {
         )
     }
 
+    override fun addTextChangedListener(watcher: TextWatcher) {
+        if (!this::textWatcher.isInitialized) {
```
jmartinesp, line 586: Thanks @ganfra for this, I didn't really remember how to handle code that was called when a view is inflated and was stuck with a null pointer exception.

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
```diff
@@ -578,6 +581,18 @@ class EditorEditText : AppCompatEditText {
...
+            this.textWatcher = EditorTextWatcher()
+            super.addTextChangedListener(this.textWatcher)
+        }
+        textWatcher.addChild(watcher)
+    }
+
+    override fun removeTextChangedListener(watcher: TextWatcher) {
+        textWatcher.removeChild(watcher)
```
ganfra, line 593: If there is no child, remove the textWatcher from parent too?

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorTextWatcher.kt
```diff
@@ -0,0 +1,110 @@
...
+
+import android.text.Editable
+import android.text.TextWatcher
+import timber.log.Timber
+import java.util.concurrent.atomic.AtomicBoolean
+
+/**
+ * A [TextWatcher] that intercepts changes in the underlying text and can have child watchers.
+ */
+class EditorTextWatcher: TextWatcher {
```
ganfra, line 11: internal?

ganfra (approved):
Small questions, otherwise LGTM, happy to see that my previous work is useful :D

platforms/android/library/src/main/java/io/element/android/wysiwyg/EditorEditText.kt
jmartinesp: I don't think it's worth it, to be honest. Keeping it there has no cost as far as I can tell, and will also be needed to notify the composer of any changes made by any external actors as the spell checker or some suggestion replacements.

PR #949, opened 2024-03-19
Android: fix English characters in some Chinese keyboards

jmartinesp:
Some Chinese keyboards send an empty composing text event before each key event for English characters. Those should be ignored and instead the current composition should be finished.

Fixes https://github.com/element-hq/element-x-android/issues/2439.


ganfra (approved):
Makes sense!

PR #952, opened 2024-03-27
Add an Info.plist to both iOS frameworks to fix Xcode 15.3 deployment.

pixlwave:
This fixes the issue with deploying an app that contains the RTE that was build with Xcode 15.3. It was missing an `Info.plist` inside each of the frameworks embedded in the xcframework.

I've added the most minimal set of properties Xcode would allow me to so it can be shared between platforms:

```xml
<key>CFBundleExecutable</key>
<string>WysiwygComposerFFI</string>
<key>CFBundleIdentifier</key>
<string>org.matrix.WysiwygComposerFFI</string>
<key>CFBundleInfoDictionaryVersion</key>
<string>6.0</string>
<key>CFBundlePackageType</key>
<string>FMWK</string>
```


stefanceriu (approved):
👏 👏 👏

PR #954, opened 2024-03-28
Attach Swift binaries to GitHub releases.

pixlwave:
This PR adds a new release script for Swift that attaches the XCFramework to GitHub releases instead of committing to keep the dependency size down. There's a new repo [here](https://github.com/matrix-org/matrix-rich-text-editor-swift) for us to start fresh.

Still a WIP, need to figure out the following
- [ ] Push rights from the workflow on this repo to the package repo.
- [ ] API rights from the workflow on this repo to create the release on the package repo.
- [ ] Update the release workflow to `cd && swift run release --version x.y.z`.
- [ ] Find out if running `release_ios.sh` without a `TAG` value is needed anywhere?

### DIFF
--- a/platforms/ios/tools/release/Sources/Release.swift
+++ b/platforms/ios/tools/release/Sources/Release.swift
@@ -1,0 +1,238 @@
+import ArgumentParser
+import CryptoKit
+import Foundation
+
+@main
+struct Release: AsyncParsableCommand {
+    
+    @Option(help: "The version of the library that is being released.")
+    var version: String
+    
+    var apiToken = ProcessInfo.processInfo.environment["GITHUB_TOKEN"]!
+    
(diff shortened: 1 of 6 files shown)


platforms/ios/tools/release/Sources/Release.swift
```diff
@@ -0,0 +1,238 @@
...
+
+@main
+struct Release: AsyncParsableCommand {
+    
+    @Option(help: "The version of the library that is being released.")
+    var version: String
+    
+    var apiToken = ProcessInfo.processInfo.environment["GITHUB_TOKEN"]!
+    
+    var packageRepo = "pixlwave/rich-text-editor-test"
```
pixlwave, line 13: I'm using this repo for testing purposes.

PR #956, opened 2024-04-01
Simplify input connection code by using `InputConnectionWrapper`.

jmartinesp:
Also reduce the number of times the selection changed listener is called.


ganfra (approved):
Makes sense (not tested)

PR #967, opened 2024-04-11
iOS tests: Use xcresultparser instead of slather

manuroe:
As for Element-X-iOS


stefanceriu (approved):
Very nice! Compared it to what we in ElX, looks good to me 👍

PR #970, opened 2024-04-24
Android: make `RichTextEditorState.messageMarkdown` store pure markdown

jmartinesp:
At the moment it contains some pseudo MD-HTML hybrid we could use to display mention pills, but we're not actually using right now AFAICT.

It's causing https://github.com/element-hq/element-x-android/issues/2214.


bmarty (approved):
Looks decent, thanks!

PR #971, opened 2024-05-02
Android: use the right selection indexes when deleting text.

jmartinesp:
- Backspacing was not updating the selection and it could become out of sync causing https://github.com/element-hq/element-x-android/issues/2784.
- Forward deletion wasn't sending composer indexes, it was using Android editor's ones.


bmarty (approved):
SGTM, but since there is no Unit test (or at least no change in any unit test), it's hard to see the effects of such change.

jmartinesp:
Added a test in https://github.com/matrix-org/matrix-rich-text-editor/pull/971/commits/ef52f45fa9b93c4d6cc89f9dec4952fb8106e2aa so we won't have that regression in the future.

PR #972, opened 2024-05-03
Version 2.37.1

jmartinesp:
Changes:

- [Android] Return pure Markdown in the `RichTextEditorState.messageMarkdown` property instead of MD + HTML for mention pills.
- [Android] Fix selection indexes not being up to date when adding a punctuation mark using GBoard keyboard in Android < 13.


bmarty (approved):
Thanks!

PR #973, opened 2024-05-06
Android: fix selection mismatch on Android 14 after autocompletion

jmartinesp:
This was caused by `EditorInputAction.UpdateSelection` not returning a `ReplaceTextResult`.

`ReplaceTextResult` was wrapped inside a `ComposerResult` sealed interface and the missing `SelectionUpdated` was added to it.


bmarty (approved):
Trying to find a bug reading the code, but I can't, nice work!
Thanks for working on this!

PR #978, opened 2024-05-14
Android: add `RichTextEditorState.setMarkdown(text)` function

jmartinesp:
Also:
- Make `TextView.applyStyleInCompose(RichTextEditorStyle)` public.
- Handle focus request in compose in a better way.
- When sending some actions through `RichTextEditorState`, wait until the UI is ready to process them.

This code is needed to add a plain text editor on Android for Compose.


bmarty (approved):
LGTM, thanks!

PR #979, opened 2024-05-15
Use the shared tools in the Swift release script.

pixlwave:
Sources most of the logic from https://github.com/element-hq/swift-command-line-tools instead of keeping it locally (this is shared with the Rust SDK release script).


stefanceriu (approved):
Such a nice cleanup! 👏

PR #980, opened 2024-05-15
Use the Swift Git tool in the release script.

pixlwave:
Similar to the SDK release scripts, plus this one adds clone, fetch and checkout steps.


stefanceriu (approved):
🥇

Issue #986, opened 2024-06-14
HTML escape mention display names


stefanceriu:
The clients can use the `insertMentionAtSuggestion` method to pass in free text which can contain HTML that will try to be interpreted, breaking the composer and generating wrong export strings.

PR #988, opened 2024-06-19
Escape text used for mentions and links

jmartinesp:
Fixes https://github.com/matrix-org/matrix-rich-text-editor/issues/986


jmartinesp:
I actually wanted to do this at a more internal level, but at that point the `String` passed to the RTE is now a `Utf16String` and `html_escape` won't work on that.

stefanceriu (approved):
lgtm, 🐑 it!

PR #1004, opened 2024-07-03
Remove Ruby version: it was used for Slather, but we don't use it anymore

jmartinesp:
See the changelog:

```
# [2.37.0] - 2024-04-12
- [Android] Discard mentions with no associated text.
- [iOS] iOS tests: Use xcresultparser instead of slather.
```


pixlwave (approved):
Nice! Always happy to see less Ruby 😅

We can probably remove the group from the renovate.json now too right?

PR #1009, opened 2024-07-05
Auto-replace plain text with emoji

langleyd:
## What's included this PR?
Custom Suggestion Pattern API and Auto-replace plain text with emoji feature(Web only so far). E.g.`:) => 🙂`

## What does it look like?
https://github.com/matrix-org/matrix-rich-text-editor/assets/6135282/0a7e6ccf-d303-464f-bc79-6c01d921dd90

## Implementation detail

I considered 2 approaches for this:
1. Initially I considered passing all of the text-to-emoji mappings and have all the logic on the rust side to do the replacement when the text changes. The advantage to this is that it maximises code re-use on the rust side.
2. Then I considered extending the Suggestions api(that slash commands and mentions are built upon) to support "custom" suggestion patterns(in addition to the existing static ones `/`,`@`, `#`). The advantage of this is that it builds upon an existing pattern rather than creating a new way of doing things. It also gives flexibility to the client as to how the feature works. The version implemented works as the web-client does, by automatically replacing the suggestions, but there has been a suggestion previously that it should work as it does on Telegram where suggestions are rendered first from which the user can choose to accept:

image


jmartinesp (approved):
The Rust part LGTM, good job!

florianduros (approved):

PR #1013, opened 2024-07-10
Fix lib type export

langleyd:
`TraceAction` wasn't exported so the last release broke the renovate PR on react-sdk.

### DIFF
--- a/platforms/web/vite.config.ts
+++ b/platforms/web/vite.config.ts
@@ -34,6 +34,7 @@ export default defineConfig({
                 'lib/types.ts',
                 'lib/constants.ts',
                 'lib/useListeners/types.ts',
+                'lib/useTestCases/types.ts',
             ],
             rollupTypes: true,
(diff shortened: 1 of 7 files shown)


platforms/web/vite.config.ts
```diff
@@ -34,6 +34,7 @@ export default defineConfig({
                 'lib/types.ts',
                 'lib/constants.ts',
                 'lib/useListeners/types.ts',
+                'lib/useTestCases/types.ts',
```
florianduros, line 37: Hmmm, why do we need to export a type used in test cases?

langleyd: Yea, its a bit weird that the function in question I think just returns this type so that we can unit test it.

At a high level you are right though, the app code shouldn't depend on the test code. I'll move the type to the app code to switch this around.

florianduros (approved):

PR #1019, opened 2024-07-16
Use Jsoup instead of TagSoup for HTML parsing.

jmartinesp:
Jsoup, being way more flexible, allows us to simplify the parsing *a lot* and do checks we couldn't do before like one which lets us know when extra line breaks between inline nodes and block ones are needed.

### DIFF
--- a/platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
+++ b/platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
@@ -81,6 +81,9 @@ class MainActivity : ComponentActivity() {
                 var linkDialogAction by remember { mutableStateOf<LinkAction?>(null) }
                 val coroutineScope = rememberCoroutineScope()
 
+                LaunchedEffect(state.messageHtml) {
+                    Timber.d("Message HTML: '${state.messageHtml}'")
+                }
                 val htmlText = htmlConverter.fromHtmlToSpans(state.messageHtml)
 
                 linkDialogAction?.let { linkAction ->
--- a/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnectionIntegrationTest.kt
+++ b/platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnectionIntegrationTest.kt
@@ -26,7 +26,8 @@ class InterceptInputConnectionIntegrationTest {
     private val viewModel = EditorViewModel(
         provideComposer = { newComposerModel() },
     ).also {
-        it.htmlConverter = HtmlConverter.Factory.create(context = app,
+        it.htmlConverter = HtmlConverter.Factory.create(
+            context = app,
             styleConfig = styleConfig,
             mentionDisplayHandler = null,
         )
@@ -57,7 +58,7 @@ class InterceptInputConnectionIntegrationTest {
             textView.text.dumpSpans(), equalTo(
                 listOf(
                     "hello: android.widget.TextView.ChangeWatcher (0-5) fl=#6553618",
-                    "hello: android.text.style.StyleSpan (0-5) fl=#33",
+                    "hello: android.text.style.StyleSpan (0-5) fl=#17",
                     "hello: android.text.method.TextKeyListener (0-5) fl=#18",
                     "hello: android.text.style.UnderlineSpan (0-5) fl=#289",
                     "hello: android.view.inputmethod.ComposingText (0-5) fl=#289",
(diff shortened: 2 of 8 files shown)


platforms/android/library/src/androidTest/java/io/element/android/wysiwyg/inputhandlers/InterceptInputConnectionIntegrationTest.kt
```diff
@@ -57,7 +58,7 @@ class InterceptInputConnectionIntegrationTest {
             textView.text.dumpSpans(), equalTo(
                 listOf(
                     "hello: android.widget.TextView.ChangeWatcher (0-5) fl=#6553618",
-                    "hello: android.text.style.StyleSpan (0-5) fl=#33",
+                    "hello: android.text.style.StyleSpan (0-5) fl=#17",
```
jmartinesp, line 61: These changes are caused by using `inSpans`, but they shouldn't affect the functionality of the editor.

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -46,464 +42,227 @@ internal class HtmlToSpansParser(
...
+            "i", "em" -> parseInlineFormatting(element, InlineFormat.Italic)
+            "u" -> parseInlineFormatting(element, InlineFormat.Underline)
+            "del" -> parseInlineFormatting(element, InlineFormat.StrikeThrough)
+            "code" -> parseInlineCode(element)
+            "ul", "ol" -> parseChildren(element)
+            "li" -> parseListItem(element)
+            "pre" -> parseCodeBlock(element)
+            "blockquote" -> parseQuote(element)
+            "p" -> parseParagraph(element)
+            "br" -> parseLineBreak(element)
```
bmarty, line 82: Maybe add a log for un-handled tag names?

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
```diff
@@ -46,464 +42,227 @@ internal class HtmlToSpansParser(
...
 
-    override fun endElement(uri: String?, localName: String, qName: String?) {
-        handleEndTag(localName)
+    private fun SpannableStringBuilder.parseInlineFormatting(element: Element, inlineFormat: InlineFormat) {
+        val span = when (inlineFormat) {
+            InlineFormat.Bold -> StyleSpan(Typeface.BOLD)
+            InlineFormat.Italic -> StyleSpan(Typeface.ITALIC)
+            InlineFormat.Underline -> UnderlineSpan()
+            InlineFormat.StrikeThrough -> StrikethroughSpan()
+            InlineFormat.InlineCode -> return
```
bmarty, line 105: This `return` statement is strange to me.

bmarty (commented):
Nice work, thanks!

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
jmartinesp: I added some comments to clarify this.

platforms/android/library/src/main/java/io/element/android/wysiwyg/utils/HtmlToSpansParser.kt
jmartinesp: Good call! In fact, I added a clean step where we remove any tags and attributes we don't recognise.

platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
```diff
@@ -81,6 +81,9 @@ class MainActivity : ComponentActivity() {
                 var linkDialogAction by remember { mutableStateOf<LinkAction?>(null) }
                 val coroutineScope = rememberCoroutineScope()
 
+                LaunchedEffect(state.messageHtml) {
+                    Timber.d("Message HTML: '${state.messageHtml}'")
+                }
```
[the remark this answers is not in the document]
jmartinesp: Not really, this is in the compose integration example only 😅 .

jmartinesp:
I had to add some extra processing for whitespaces because Jsoup apparently doesn't directly strip it as Tagsoup did, and this could potentially cause index issues in the future.

platforms/android/example-compose/src/main/java/io/element/wysiwyg/compose/MainActivity.kt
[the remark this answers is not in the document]
bmarty: Ah OK, I read it a bit fast. I am always paranoid regarding what app and libraries are logging out.

bmarty (approved):
Thanks for the updates!

PR #1032, opened 2024-08-28
Escape text in links and mentions for wasm bindings

langleyd:
Escape text in links and mentions for wasm bindings

Same as https://github.com/matrix-org/matrix-rich-text-editor/pull/988 but for wasm


langleyd:
`js_sys::Map` couldn't be instantiated within the `lib.rs` tests as it's outside of a wasm context. Not sure how to work around this, so wrote the test in typescript to cover it.

florianduros (approved):

